PR #1, opened 2016-03-21
Support job arrays in nodewatcher plugin for Torque.

mkosmo:


dougalb:
Thanks! Will be included in cfncluster-1.2.0.

mkosmo:
Awesome, thanks!  I'll have a patch
for cfncluster-cookbook:templates/default/publish_pending.torque.erb
submitted soon for the same issue.

PR #3, opened 2016-07-11
there is no point in restarting the torque server each time

guicho271828:


dougalb:
This was added due to issues with Torque not correctly scheduling parallel jobs after new instances are added. Unless that behaviour of Torque has changed, this is a required restart due to the elastic nature of CfnCluster.

guicho271828:
scheduling --- I guess the problem is in pbs_sched, not pbs_server. Restarting the scheduler, not server, is less problematic. I do this, in fact, because sometimes pbs_sched dies on the restart of pbs_server: 

* restart pbs_sched
* hold all jobs temporarily
* release all jobs

pbs_sched is not actively maintained. Rather, the recommended way is to install Maui scheduler by Adaptive computing (licensed) which is more scalable and stable. I understand that due to the open source nature cfncluster should stick to pbs_sched. You should just restart pbs_sched.

dougalb:
Are you experiencing an issue with Torque running in CfnCluster?

guicho271828:
I'm writing scripts to handle the unexpected restart of pbs_server.

dougalb:
I'm not sure I understand, the restart of pbs_server is expected as part of
CfnCluster. Is the restart impacting the cluster?

Do you have a hard requirement for Torque. SGE performs the best in
CfnCluster.

dougalb:
I've carried out testing with torque-6.0.1 and without a restart, there is still issues with jobs being scheduled to new nodes.

guicho271828:
Yes, just stopping restarting the server is not a good idea. Instead, did you try restarting the scheduler?

dougalb:
Yes, that has been tested as well. Happy to look at pull requests that
removes the need for the restart.

The basic test, is to lauch a cluster with 1 or more instances in the
ComputeFleet. Then reduce the size to zero instances, followed by scaling
back up to 1 or more instances. After that simple batch and multi-node
parallel jobs should be submitted, and then scheduled to the new instances.

PR #4, opened 2016-08-18
Bugfix. 'stdout' referenced before assignment in sqswatcher slurm plugin

f403:


dougalb:
Thanks for catching this.

PR #5, opened 2016-09-08
Adding support for multiple queues

f403:
I'm working on the implementation of a Heterogeneous CFN Cluster (like mentioned in awslabs/cfncluster#120) and would like to have this included to cfncluster-node package. Otherwise custom Chef-cookbook and custom cfncluster-node pypi package have to be used, which makes my template less compatible with original CFN Cluster implementation.

Main idea is:
- AutoScalingGroup in cfn-template sets a "queue" tag for each Compute Node
- sqswatcher.py parses this tag (if present), saves it to dynamodb and passes to the scheduler plugin
- Scheduler plugin creates a queue with the same name and adds/removes this host to this queue; if "queue" tag is not present host is added to common queue.

Last step is currently implemented for slurm only.

I expect these changes to be compatible with normal CFN Cluster template.

### DIFF
--- a/sqswatcher/plugins/openlava.py
+++ b/sqswatcher/plugins/openlava.py
@@ -29,7 +29,7 @@ def __runOpenlavaCommand(command):
         log.error("Failed to run %s\n" % _command)
 
 
-def addHost(hostname, cluster_user, slots):
+def addHost(hostname, cluster_user, slots, queue=None):
     log.info('Adding %s with %s slots' % (hostname,slots))
 
     command = ('/opt/openlava/bin/lsaddhost -t linux -m IntelXeon -M "%s" %s' % (slots, hostname))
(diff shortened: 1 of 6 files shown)


bwbarrett:
@f403, is there another part of this patch?  I think I'm ok with the changes, but am likely to merge them after the next release (because we desperately need to get a release cadence going again, and this has the potential to be impactful).

mohanasudhan:
Changes looks fine. can you remove openlava from you changes? I am ok to push this change after that.

jerowe:
Any word on this? We would very much like multiple queues and instance types! ;-)

f403:
I'll take a look next weeks, and let you know.
Sorry for it taking long to response

enrico-usai:
See https://github.com/aws/aws-parallelcluster/issues/307

PR #6, opened 2017-05-12
Leave SGE compute nodes as admin hosts

xzy3:
When an cfnCluster using SGE scales down lockHost is used to avoid having jobs scheduled on the compute node. However the `qmod -d all.q@hostname` must be run on an administration host. As part of the creation new nodes are removed from the administration host list. This breaks lockHost and occasionally will allow a job to schedule on the node just as the node shuts down. This leaves the job orphaned and the cluster in a state that is very difficult to recover from without recreating the cluster.

Fixes https://github.com/awslabs/cfncluster/issues/148

### DIFF
--- a/sqswatcher/plugins/sge.py
+++ b/sqswatcher/plugins/sge.py
@@ -95,10 +95,6 @@ report_variables      NONE
         time.sleep(1)
     ssh.close()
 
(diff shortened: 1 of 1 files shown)


bwbarrett:
Thanks for the PR!

PR #8, opened 2017-06-01
Fix license string in setup.py

bwbarrett:
When the license for cfncluster-node was changed to Apache, the
classifier string was updated but not to the exact format that
PyPi wants, so PyPi rejected new versions.  Clean up the string
so that we can release 1.3.4.

Signed-off-by: Brian Barrett <bbarrett@amazon.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,6 +51,6 @@ setup(
         "Environment :: Console",
         "Programming Language :: Python",
         "Topic :: Scientific/Engineering",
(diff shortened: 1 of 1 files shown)


bwbarrett:
Here's the list of accepted classifiers: https://pypi.python.org/pypi?%3Aaction=list_classifiers

dougalb:
Ship it!

PR #11, opened 2017-09-21
Reschedule any jobs on the host before attempting to remove it

ajpasadyn:
Fixes issue https://github.com/awslabs/cfncluster-node/issues/10

### DIFF
--- a/sqswatcher/plugins/sge.py
+++ b/sqswatcher/plugins/sge.py
@@ -103,6 +103,42 @@ report_variables      NONE
     command = ('/opt/sge/bin/lx-amd64/qconf -aattr queue slots ["%s=%s"] all.q' % (hostname,slots))
     __runSgeCommand(command)
 
+def __getJobs(hostname):
+    # Checking for running jobs on the node
+    log.info("Checking for jobs running on %s", hostname)
+    command = "/opt/sge/bin/lx-amd64/qstat -u '*' -q all.q@%s" % (hostname)
+    log.info("%s", command)
+    _command = shlex.split(command)
+    log.debug(_command)
+    try:
+        _output = sub.check_output(_command, env=dict(os.environ, SGE_ROOT='/opt/sge'), universal_newlines=True)
+    except sub.CalledProcessError:
+        # this call will return error status if no jobs are in the queue
+        _output = ""
+
+    # sample array job output:
+    # job-ID  prior   name       user         state submit/start at     queue                          slots ja-task-ID
+    # -----------------------------------------------------------------------------------------------------------------
+    #       3 0.55500 test_job.s ajp          t     06/16/2017 14:16:54 all.q@ip-10-0-79-245.ec2.inter     1 39
+    #       3 0.55500 test_job.s ajp          t     06/16/2017 14:16:54 all.q@ip-10-0-79-245.ec2.inter     1 40
+    #       3 0.00000 test_job.s ajp          qw    06/16/2017 14:16:47                                    1 41-100:1
+
+    _jobs = list()
+    for line in _output.split('\n')[2:]:
+        if len(line.strip()) == 0: continue
+        log.info("%s", line)
+        parts = line.split()
+        state = parts[4]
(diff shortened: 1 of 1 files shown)


sqswatcher/plugins/sge.py
```diff
@@ -118,6 +154,12 @@ def removeHost(hostname,cluster_user):
     command = ("/opt/sge/bin/lx-amd64/qconf -dattr hostgroup hostlist %s @allhosts" % hostname)
     __runSgeCommand(command)
 
+    # Reschedule any jobs that are on the host
```
xzy3: You probably want to move this before line 145. The qmod manpage says `qmod -rj` requires admin-host privileges. I'm trying to find the relevant portion of the gridengine source code to verify that this is truly the case.

ajpasadyn: The code I've been running in production for the last couple months actually had the entire "Removing host as administrative host" stanza removed, not just relocated into removeHost() as you had done.

I don't trust moving the rescheduling before the all.q and allhosts changes, though, since that could allow jobs to sneak on to the node. So if you think the admin host removal still needs to happen I'd recommend moving that piece to right after the job rescheduling.

Aside from that, though, this `qmod -rj` is running on the master node as root, so I'm not convinced it matters whether the hostname being removed is in the admin list or not. In fact, at the time this runs for many of my test cases, the node in question has already stopped running.

xzy3:
I'd be OK with moving the admin host removal stanza below this. When a node terminates on auto scale down the nodewatcher tries to lock the node here https://github.com/awslabs/cfncluster-node/blob/4bf8ed9bf3e73f882136f6c48397c9ea86d58a53/nodewatcher/nodewatcher.py#L114 and the node must be an admin host for that to run successfully. Removal of the admin host is probably a good idea, GE does not clean up after itself well when a node is removed with running jobs as we've both noticed.

ajpasadyn:
I updated the pull request and did some basic testing with that version. Everything appears to be working when I terminate instances manually.

```
2017-09-21 14:35:42,390 INFO [sqswatcher:pollQueue] eventType=autoscaling:EC2_INSTANCE_TERMINATE
2017-09-21 14:35:42,391 INFO [sqswatcher:pollQueue] instanceId=i-06c95e47afa7eab79
2017-09-21 14:35:42,443 INFO [sge:removeHost] Removing ip-10-0-10-161
root@ip-10-0-14-146.ec2.internal modified "all.q" in cluster queue list
root@ip-10-0-14-146.ec2.internal modified "@allhosts" in host group list
2017-09-21 14:35:42,466 INFO [sge:__getJobs] Checking for jobs running on ip-10-0-10-161
2017-09-21 14:35:42,467 INFO [sge:__getJobs] /opt/sge/bin/lx-amd64/qstat -u '*' -q all.q@ip-10-0-10-161
2017-09-21 14:35:42,479 INFO [sge:__getJobs]       1 0.55500 test_job.s ajp          t     09/21/2017 14:33:18 all.q@ip-10-0-10-161.ec2.inter     1 112
2017-09-21 14:35:42,479 INFO [sge:__getJobs]       1 0.00000 test_job.s ajp          qw    09/21/2017 14:23:29                                    1 132-1000:1
2017-09-21 14:35:42,479 INFO [sge:__getJobs] Found 1 jobs on ip-10-0-10-161
Forced rescheduling of job-array task 1.112 on host ip-10-0-10-161.ec2.internal
root@ip-10-0-14-146.ec2.internal removed "ip-10-0-10-161.ec2.internal" from administrative host list
root@ip-10-0-14-146.ec2.internal removed "ip-10-0-10-161.ec2.internal" from execution host list
root@ip-10-0-14-146.ec2.internal removed "ip-10-0-10-161.ec2.internal" from submit host list
```

enrico-usai:
I'm going to close this PR because the #10 has been resolved in the upcoming 1.6.0 release, by an hard scaling up/down refactoring:
https://github.com/awslabs/cfncluster/blob/develop/CHANGELOG.rst#160

PR #14, opened 2017-10-31
Default to terminating 5min after last job completes.

adamrb:
Now that instances are billed by the second, there's no need to wait until the instance hour before terminating.  This will change the default to 5 minutes after the last job has been seen and allow it to be configurable via config the file (with updates to cfncluster and cfncluster-chef packages).

### DIFF
--- a/nodewatcher/nodewatcher.py
+++ b/nodewatcher/nodewatcher.py
@@ -14,6 +14,7 @@
 __author__ = 'dougalb'
 
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


jmenbo:
Hey @bwbarrett, I hear your objection to the change, but from a customer perspective, it would be great if the change could be implemented now to reduce operational costs, even if the file location is not ideal. Moving the change to a different location down the road is still possible. Thanks

PR #15, opened 2017-11-03
Dropping Python 3.3 builds and adding 3.6.

rajachan:
[Paramiko](http://www.paramiko.org/), a SSH library that sqswatcher uses, supports just Python 2.7/3.4+.
The Travis builds are failing because of this incompatibilty when installing
paramiko and cryptography from PyPi. Also adding 3.6 builds, which is the
current stable version of Python 3 which users are most likely to use.

Signed-off-by: Raghu Raja <craghun@amazon.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,9 @@ language: python
 python:
   - "2.6"
   - "2.7"
-  - "3.3"
   - "3.4"
   - "3.5"
+  - "3.6"
 
 install:
   - pip install -r requirements.txt


mohanasudhan:
Awesome !!! thanks for the PR

bwbarrett:
With the updated AMIs, what versions of Python are going to exist?

rajachan:
I think all distros still do 2.7.x by default (will verify), but this is for when users pull in Python 3 by choice.

PR #17, opened 2017-11-09
Bumping version to 1.4.1 as py_pi doesn't allow override

mohanasudhan:
Overriding existing version is not possible in py_pi
hence had to bump version.

Signed-off-by: Mohan Gandhi <mohgan@amazon.com>


rajachan:
Ship it!

PR #18, opened 2017-11-22
Correcting slurm restart logic

mohanasudhan:
Fixing Slurm restart in Master and compute nodes. With
respect to compute node restart, we are checking path existence 
in master node which is wrong, now moved this logic to command line 
so it executes in compute node. And in Master slurm restart we need to pass 
restart command as list.

Signed-off-by: Mohan Gandhi <mohgan@amazon.com>

### DIFF
--- a/sqswatcher/plugins/slurm.py
+++ b/sqswatcher/plugins/slurm.py
@@ -13,6 +13,7 @@ import subprocess as sub
 from tempfile import mkstemp
 from shutil import move
 import os
+import os.path
 import paramiko
 import socket
 import time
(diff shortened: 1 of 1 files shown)


sqswatcher/plugins/slurm.py
```diff
@@ -158,9 +155,10 @@ def removeHost(hostname, cluster_user):
 
 
 def restartMasterNodeSlurm():
-    slurmRestart = Path('/etc/systemd/system/slurmctld.service')
-    if slurmRestart.is_file():
-        command = 'sudo systemctl restart slurmctld.service'
+    if os.path.isfile('/etc/systemd/system/slurmctld.service'):
+        command = ['sudo', 'systemctl', 'restart', 'slurmctld.service']
+    elif os.path.isfile('/etc/systemd/system/slurmd.service'):
```
bwbarrett: This isn't needed.  This function is restartMasterNodeSlurm; the slurmd process never runs on the master node (ie, if this file exists, other things are broken).

mohanasudhan: Ya, correct now.

PR #27, opened 2018-06-06
Wake-up Torque scheduler when attaching new nodes

mauri-melato:
This fix wants to address a problem with the Torque scheduler in the
scenario when the cluster has no compute nodes available.
Jobs get scheduled and go in Q state, but even when new compute nodes
become available to the cluster, the jobs remain in Q state.
This has a major impact to customers that see the cluster scaling up to
the max size but without any job running.

From tested approaches and according to Adaptive, one way to recover
from this situation is to reset the Torque server flag 'scheduling=true'
This works when the node fully joined the cluster and it's 'free'.

This patch implements this approach, when a node is added to the
cluster and becomes free, the 'scheduling' flag is reset.

Torque server restart commands have been removed because according to
official documentation (https://tinyurl.com/y7vjrztw), "When you make
changes to nodes by directly editing the nodes file, you must restart
pbs_server for those changes to take effect. **Changes made using qmgr
do not require a restart.**"
Restart the server can impact the user's experience for a bunch of seconds, moreover frequent restarts when adding and removing nodes may undermine the stability of the pbs_server and pbs_sched processes.

Signed-off-by: Maurizio Melato <mmelato@amazon.com>

### DIFF
--- a/sqswatcher/plugins/torque.py
+++ b/sqswatcher/plugins/torque.py
@@ -16,17 +16,61 @@ import os
 import paramiko
 import logging
 import shlex
+import time
+import xml.etree.ElementTree as xmltree
 
 log = logging.getLogger(__name__)
 
 def __runCommand(command):
+    output = None
     log.debug(repr(command))
     _command = shlex.split(str(command))
     log.debug(_command)
     try:
-        sub.check_call(_command, env=dict(os.environ))
-    except sub.CalledProcessError:
-        log.error("Failed to run %s\n" % _command)
+        output = sub.check_output(_command, env=dict(os.environ), stderr=sub.STDOUT)
+    except sub.CalledProcessError as exc:
+        log.error("Failed to run %s:\n%s" % (_command, exc.output))
+    return output
+
+def isHostInitState(host_state):
+    # Node states http://docs.adaptivecomputing.com/torque/6-0-2/adminGuide/help.htm#topics/torque/8-resources/resources.htm#nodeStates
+    init_states = ("down", "offline", "unknown", str(None))
+    return str(host_state).startswith(init_states)
+
+def wakeupSchedOn(hostname):
+    log.info('Waking up scheduler on host %s', hostname)
+    command = ("/opt/torque/bin/pbsnodes -x %s" % (hostname))
+
+    sleep_time = 3
+    times = 20
+    host_state = None
+    while isHostInitState(host_state) and times > 0:
+        output = __runCommand(command)
+        try:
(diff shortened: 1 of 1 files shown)


sqswatcher/plugins/torque.py
```diff
@@ -74,8 +118,4 @@ def removeHost(hostname, cluster_user):
     __runCommand(command)
 
     command = ("/opt/torque/bin/qmgr -c 'delete node %s'" % hostname)
-    __runCommand(command)
-
-    command = ('/etc/init.d/pbs_server restart')
-    __runCommand(command)
-
+    __runCommand(command)
```
bwbarrett: you're missing the newline for the last line.

sqswatcher/plugins/torque.py
```diff
@@ -16,17 +16,62 @@
...
+    return str(host_state).startswith(init_states)
+
+def wakeupSchedOn(hostname):
+    log.info('Waking up scheduler on host %s', hostname)
+    command = ("/opt/torque/bin/pbsnodes -x %s" % (hostname))
+
+    sleep_time = 3
+    times = 20
+    host_state = None
+    #while host_state in init_states and times > 0:
```
bwbarrett: remove commented-out line

sqswatcher/plugins/torque.py
```diff
@@ -16,17 +16,62 @@
...
+            # Ex 2: <Data><Node><name>ip-10-0-76-39</name><state>free</state><power_state>Running</power_state><np>1</np><ntype>cluster</ntype>
+            #        <status>rectime=1527799181,macaddr=02:e4:00:b0:b1:72,cpuclock=Fixed,varattr=,jobs=,state=free,netload=210647044,gres=,loadave=0.00,
+            #        ncpus=1,physmem=1017208kb,availmem=753728kb,totmem=1017208kb,idletime=856,nusers=1,nsessions=1,sessions=19698,
+            #        uname=Linux ip-10-0-76-39 4.9.75-25.55.amzn1.x86_64 #1 SMP Fri Jan 5 23:50:27 UTC 2018 x86_64,opsys=linux</status>
+            #        <mom_service_port>15002</mom_service_port><mom_manager_port>15003</mom_manager_port></Node></Data>
+            xmlnode = xmltree.XML(output)
+            host_state = xmlnode.findtext("./Node/state")
+        except:
+            log.error("Error parsing XML from %s" % output)
+        # if (node_state in init_states):
```
lukeseawalker: remove commented-out line

mauri-melato:
Yes @bwbarrett, you're correct, we'll have to take this into account. In principle this wake-up is only needed when transitioning from 0 compute nodes to >0  and not for every new node being added. A quick mitigation would be to avoid to do two wake ups within few minutes.

PR #36, opened 2018-07-12
Scale down based on idle time instead of hour percentile

fnubalaj:
Get the idle time threshold parameter from nodewatcher.cfg
Every minute check the node for running jobs and if idle for more than the threshold, then terminate
Store the idle time as a json file in case process terminates


nodewatcher/nodewatcher.py
```diff
@@ -174,10 +185,15 @@ def main():
             jobs = getJobs(s, hostname)
             if jobs == True:
                 log.info('Instance actually has active jobs.')
+                data['idle_time'] = 0
                 lockHost(s, hostname, unlock=True)
                 continue
             else:
-                selfTerminate(region, asg, instance_id)
+                data['idle_time'] += 1
+                log.info('Instance has no job for the past %s minute(s)' % data['idle_time'])
```
sean-smith: Include instance id in log: "Instance i-1234567 has no..."

nodewatcher/nodewatcher.py
```diff
@@ -54,7 +57,7 @@ def getConfig(instance_id):
         os.rename(tup[1], 'nodewatcher.cfg')
 
     log.debug("region=%s asg=%s scheduler=%s" % (_region, _asg, _scheduler))
-    return _region, _asg, _scheduler
+    return _region, _asg, _scheduler, _idle_time
 
 def getHourPercentile(instance_id, conn):
```
sean-smith: Remove this method

nodewatcher/nodewatcher.py
```diff
@@ -37,6 +39,7 @@ def getConfig(instance_id):
         logging.getLogger().setLevel(lvl)
     _region = config.get('nodewatcher', 'region')
     _scheduler = config.get('nodewatcher', 'scheduler')
+    _idle_time = config.get('nodewatcher', 'idle_time')
```
sean-smith: I don't like the name `idle_time`, it should be something closer to `cooldown_period`

nodewatcher/nodewatcher.py
```diff
@@ -142,6 +145,12 @@ def maintainSize(region, asg):
         log.debug('capacity less then or equal to min size.')
         return True
 
+
+def saveIdleTime(persisted_data):
+    with open('/tmp/data.json', 'w') as outfile:
```
mauri-melato: Can we place this file under a `/opt/cfncluster/data` .  `data` dir is not there today but it'd be cleaner and safer to keep our stuff down there.

bwbarrett: /opt/cfncluster/data seems like the wrong place.  `/var/run/nodewatcher` or similar makes more sense to me...

dougalb: Agree with @bwbarrett, also I would make the filename more specific to it's purpose i.e. node_idletime.json

nodewatcher/nodewatcher.py
fnubalaj: There is already a cooldown parameter and adding something similar would causing confusion. Let's stick with idle time or rename the other one and change this to cooldown_period.

scaling_cooldown
Amount of time in seconds to wait before attempting further scaling actions.

Defaults to 300 for the default template.

scaling_cooldown = 300

nodewatcher/nodewatcher.py
```diff
@@ -142,6 +131,12 @@ def maintainSize(region, asg):
         log.debug('capacity less then or equal to min size.')
         return True
 
+
+def saveIdleTime(persisted_data):
+    with open('/var/run/nodewatcher/node_idletime.json', 'w') as outfile:
```
mauri-melato: At the start of the nodewatcher you should check the dir is there and if it's not then create it.

nodewatcher/nodewatcher.py
```diff
@@ -142,6 +132,22 @@ def maintainSize(region, asg):
         log.debug('capacity less then or equal to min size.')
         return True
 
+
+def saveIdleTime(persisted_data):
+    try:
+        if not os.path.exists('/var/run/nodewatcher/'):
```
mauri-melato: I'd suggest to define a couple of module _"constants"_ , something like
_NW_DATA_DIR = "/var/run/nodewatcher/"
_NW_IDLETIME_FILE  = _NW_DATA_DIR + "node_idletime.json"
and then refer them in the code.

nodewatcher/nodewatcher.py
```diff
@@ -142,6 +132,22 @@ def maintainSize(region, asg):
...
+            os.makedirs('/var/run/nodewatcher/')
+    except OSError as ex:
+        if ex.errno == errno.EEXIST and os.path.exists('/var/run/nodewatcher/'):
+            pass
+        else:
+            log.critical('Persisting idle time failed with exception: %s' % ex)
```
mauri-melato: I would add contextual info in the log file, something like
'Error persisting idle time data "%s" to file "%s": %s' % (persisted_data, _NW_IDLETIME_FILE, ex)

nodewatcher/nodewatcher.py
mauri-melato: I agree _cooldown_ would be confusing in this case and even if we are going to remove it from the new scale-up strategy we should preserve some consistency with the terminology used until now.

nodewatcher/nodewatcher.py
fnubalaj: Okay

nodewatcher/nodewatcher.py
fnubalaj: Okay

nodewatcher/nodewatcher.py
```diff
@@ -146,6 +144,23 @@ def maintainSize(asg_name, asg_conn):
         log.debug('capacity less then or equal to min size.')
         return True
 
+
+def saveIdleTime(persisted_data):
+    try:
+        if not os.path.exists(_NW_DATA_DIR):
+            os.makedirs(_NW_DATA_DIR)
+    except OSError as ex:
+        if ex.errno == errno.EEXIST and os.path.exists(_NW_DATA_DIR):
```
mauri-melato: I think we can safely get rid of this check and just log and raise the exception in case of errors

nodewatcher/nodewatcher.py
```diff
@@ -118,6 +111,11 @@ def getJobs(s,hostname):
 
     return _jobs
 
+def queueHasPendingJobs(s):
+    _has_pending_jobs, _error = s.queueHasPendingJobs()
+    log.debug("has_pending_jobs=%s, error=%s" %(_has_pending_jobs, _error))
```
mauri-melato: a space after %

nodewatcher/nodewatcher.py
```diff
@@ -154,35 +169,46 @@ def main():
...
     instance_id = getInstanceId()
     hostname = getHostname()
-    region, asg_name, scheduler, proxy_config = getConfig(instance_id)
+    region, asg_name, scheduler, proxy_config, idle_time = getConfig(instance_id)
+
 
     s = loadSchedulerModule(scheduler)
 
+    if os.path.isfile(_NW_IDLETIME_FILE):
+        data = json.loads(open(_NW_IDLETIME_FILE).read())
```
mauri-melato: use the `with ... as ...` construct

nodewatcher/nodewatcher.py
```diff
@@ -154,35 +169,46 @@ def main():
...
-        hour_percentile = getHourPercentile(instance_id, ec2_conn)
-        log.info('Percent of hour used: %d' % hour_percentile)
-
-        if hour_percentile < 95:
-            continue
 
         jobs = getJobs(s, hostname)
-        if jobs == True:
+        if jobs:
             log.info('Instance has active jobs.')
```
mauri-melato: Add %s for logging `jobs`

nodewatcher/nodewatcher.py
```diff
@@ -154,35 +169,46 @@ def main():
...
+            if jobs:
                 log.info('Instance actually has active jobs.')
+                data['idle_time'] = 0
                 lockHost(s, hostname, unlock=True)
                 continue
             else:
-                selfTerminate(asg_name, asg_conn, instance_id)
+                data['idle_time'] += 1
+                log.info('Instance %s has no job for the past %s minute(s)' % (instance_id, data['idle_time']))
+                if data['idle_time'] >= idle_time:
```
mauri-melato: it's confusing to use the same label/ var name for what is the *current_idletime* and the *scaledown_idletime/threshold/period*

nodewatcher/nodewatcher.py
```diff
@@ -154,35 +169,46 @@ def main():
...
         jobs = getJobs(s, hostname)
-        if jobs == True:
+        if jobs:
             log.info('Instance has active jobs.')
+            data['idle_time'] = 0
         else:
             if maintainSize(asg_name, asg_conn):
                 continue
             # avoid race condition by locking and verifying
             lockHost(s, hostname)
```
mauri-melato: I'd avoid the lock host operation when not strictly necessary.
The 'else' logic could be rephrased like:
```
  increment_idle_time
  if idle_time > scale-down-threshold
     lockHost()
     if not hasJobs() and not pendingJobs()
         selfTerminate()
     else
         unlockHost()
```

nodewatcher/plugins/sge.py
```diff
@@ -35,6 +36,37 @@ def getJobs(hostname):
 
     return _jobs
 
+def queueHasPendingJobs():
```
mauri-melato: _scheduler._`hasPendingJobs()` seems a better read

sean-smith: All methods should be lower_snake_case: https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables

nodewatcher/plugins/slurm.py
```diff
@@ -32,5 +34,32 @@ def getJobs(hostname):
 
     return _jobs
 
+def queueHasPendingJobs():
+    command = "/opt/slurm/bin/squeue -t PD"
```
mauri-melato: With `--noheader` you can avoid the first line
```
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
```
and then count check for `>0`

nodewatcher/plugins/torque.py
```diff
@@ -50,6 +51,43 @@ def getJobs(hostname):
...
+        return has_pending, error
+
+    queue_status = lines[2].split()
+
+    if len(qStatus) != 13:
+        log.error("Error in parsing the output %s\n" % lines[2])
+        error = True
+        return has_pending, error
+
+        has_pending = (int(queue_status[5]) != 0)
```
mauri-melato: This should not be indented but I'm also thinking that we are considering just the first queue.  There might be more configured and in this scale we should iterate that all the queue  are 0.

nodewatcher/plugins/torque.py
```diff
@@ -50,6 +51,43 @@ def getJobs(hostname):
...
+
+    output = process.communicate()[0]
+    exit_code = process.poll()
+
+    if exit_code != 0:
+        log.error("Failed to run %s\n" % command)
+        error = True
+        return has_pending, error
+
+    lines = output.split("\n")
```
mauri-melato: A comment with an example of the output would be helpful:
```
Queue              Max    Tot   Ena   Str   Que   Run   Hld   Wat   Trn   Ext T   Cpt
----------------   ---   ----    --    --   ---   ---   ---   ---   ---   --- -   ---
batch                0      7   yes   yes     5     2     0     0     0     0 E     0
```

nodewatcher/nodewatcher.py
fnubalaj: I left the existing design as was before my changes.

nodewatcher/plugins/torque.py
fnubalaj: Could you please elaborate on this one. Thanks

nodewatcher/plugins/torque.py
fnubalaj: Okay

nodewatcher/plugins/torque.py
mauri-melato: Yep I meant two things but I wasn't clear.
1. I think the `has_pending = ...` at line 87  should be aligned with the `if` at line 82.
2. We're parsing the output of `qstat -Q` to see if there are pending jobs. This commands returns the status of all the queues and parsing `lines[2]` we're just considering the first queue, that's the only one available by default but there might not be the case in general. We should better iterating on all the possibile queues to check if there are pending jobs.

nodewatcher/nodewatcher.py
mauri-melato: yep but in the previuos version if you were in this part of the code all the conditions to terminate the node were potentially satisfied (hourPercentile >= 95 && ! maintainSize) so we could lock the host, check for jobs and, if the case, terminate the instance.
In the new version, we've still to check if the passed idle_time "is enough" and only in this case it makes sense to disturb the scheduler to lock the host and check if new jobs were submitted in the meantime etc.

nodewatcher/nodewatcher.py
fnubalaj: Okay

nodewatcher/plugins/sge.py
fnubalaj: I am keeping the pattern in this file. If there needs to be a change, it should be done as a separate task.

nodewatcher/nodewatcher.py
fnubalaj: jobs is a boolean, why do we want to log it?

mauri-melato: No, the function should be better named as `hasJobs` but in another patch, ok.

fnubalaj: Okay, that I can do!

nodewatcher/nodewatcher.py
```diff
@@ -154,35 +167,44 @@ def main():
...
+                    has_jobs = hasJobs(s, hostname)
+                    if has_jobs:
+                        log.info('Instance has active jobs.')
+                        data[_NW_CURRENT_IDLETIME] = 0
+                    else:
+                        has_pending_jobs, error = hasPendingJobs(s)
+                        if not error and not has_pending_jobs:
+                            os.remove(_NW_IDLETIME_FILE)
+                            selfTerminate(asg_name, asg_conn, instance_id)
+                    lockHost(s, hostname, unlock=True)
```
mauri-melato: I think we should avoid to unlock the node in the case we're terminating it in order to avoid a time window where new jobs may get scheduled to this host.

nodewatcher/nodewatcher.py
```diff
@@ -23,8 +23,15 @@
 import boto3
 import ConfigParser
 from botocore.config import Config
+import json
+import atexit
+import errno
+
 
 log = logging.getLogger(__name__)
+_NW_DATA_DIR = "/var/run/nodewatcher/"
```
sean-smith: You can remove the `NW` prefix. Since this global var is only used in nodewatcher (as denoted by the `_` prefix) you don't need to indicate that it's for nodewatcher.

PR #44, opened 2018-07-20
Instance to slot mapping

fnubalaj:

### DIFF
--- a/sqswatcher/sqswatcher.py
+++ b/sqswatcher/sqswatcher.py
@@ -211,23 +211,22 @@ def pollQueue(scheduler, q, t, proxy_config):
 
 def getInstanceMappingFile(region, proxy_config):
     s3 = boto3.resource('s3', config=proxy_config)
+    cfncluster_dir = '/opt/cfncluster/'
+    cfncluster_file = cfncluster_dir + 'instances.json'
 
     try:
-        if not os.path.exists('/opt/cfncluster/'):
-            os.makedirs('/opt/cfncluster/')
+        if not os.path.exists(cfncluster_dir):
+            os.makedirs(cfncluster_dir)
     except OSError as ex:
-        if ex.errno == errno.EEXIST and os.path.exists('/opt/cfncluster/'):
-            pass
-        else:
-            log.critical('Could not save instance mapping file. Failed with exception: %s' % ex)
+            log.critical('Could not create directory %s. Failed with exception: %s' % (cfncluster_dir, ex))
             raise
 
     bucket_name = region + '-cfncluster'
     try:
         bucket = s3.Bucket(bucket_name)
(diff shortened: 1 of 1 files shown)


sqswatcher/sqswatcher.py
```diff
@@ -206,6 +208,29 @@ def pollQueue(scheduler, q, t, proxy_config):
...
 
+
+def getInstanceMappingFile(region, proxy_config):
+    s3 = boto3.resource('s3', config=proxy_config)
+
+    try:
+        if not os.path.exists('/opt/cfncluster/'):
+            os.makedirs('/opt/cfncluster/')
+    except OSError as ex:
+        if ex.errno == errno.EEXIST and os.path.exists('/opt/cfncluster/'):
```
mauri-melato: Is there the possibility of a race condition when checking/creating `/opt/cfncluster` dir? If it's not the case I would avoid this further check and raise the exception if creation fails

sqswatcher/sqswatcher.py
```diff
@@ -206,6 +208,29 @@ def pollQueue(scheduler, q, t, proxy_config):
...
+            pass
+        else:
+            log.critical('Could not save instance mapping file. Failed with exception: %s' % ex)
```
mauri-melato: "Could not create directory %s. ..."

sqswatcher/sqswatcher.py
```diff
@@ -206,6 +208,29 @@ def pollQueue(scheduler, q, t, proxy_config):
...
+            raise
+
+    bucket_name = region + '-cfncluster'
+    try:
+        bucket = s3.Bucket(bucket_name)
+        bucket.download_file('instances/instances.json', '/opt/cfncluster/instances.json')
+    except ClientError as e:
+        log.critical("Could not save instance mapping file. Failed with exception: %s" % e)
```
mauri-melato: Just more context, "Could not save instance mapping file %s from S3 bucket %s. ..."

sqswatcher/sqswatcher.py
```diff
@@ -206,6 +208,28 @@ def pollQueue(scheduler, q, t, proxy_config):
 
         time.sleep(30)
 
+
+def getInstanceMappingFile(region, proxy_config):
```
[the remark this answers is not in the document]
fnubalaj: We are doing it each time sqswatcher is starting up. Do you think we shouldn't be doing that? The race condition should not be a problem, because it the slot counts/ number of vcpus don't change that often and it is very rare that the file is being written by the jenkins task at the exact same time as it is being pulled from s3 bucket.

fnubalaj:
This pull request will be closed since this functionality will be moved to jobwatcher as a part of scale up code.

PR #48, opened 2018-08-16
Unlock host in case self termination fails

fnubalaj:
There's a race condition where two instances can both try and terminate at the same time. The ASG will throw an exception if this makes it go under minSize. This catches that exception and unlocks the host.

### DIFF
--- a/nodewatcher/nodewatcher.py
+++ b/nodewatcher/nodewatcher.py
@@ -22,7 +22,7 @@ import tempfile
 import logging
 import boto3
 import ConfigParser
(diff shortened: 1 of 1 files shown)


nodewatcher/nodewatcher.py
```diff
@@ -22,6 +22,7 @@
 import logging
 import boto3
 import ConfigParser
+import botocore.exceptions
```
sean-smith: `from botocore.exceptions import ClientError`

PR #59, opened 2018-09-05
Fix scale up when hyperthreading is disabled

fnubalaj:
Parse the cfnconfig file in /opt/cfncluster to get the cfn_scheduler_slots config parameter
Based on that parameter set the slot count
Exit gracefully with error messages in cases of invalid configuration


jobwatcher/jobwatcher.py
```diff
@@ -51,17 +52,56 @@ def get_asg_name(stack_name, region, proxy_config):
...
+                slots = vcpus
+
+        if not slots > 0:
+            log.critical("cfn_scheduler_slots config value '%s' was invalid" % cfn_scheduler_slots)
+            exit(1)
+
+        return {'slots': slots}
+
+    except KeyError:
+        log.error("Required config value 'cfn_scheduler_slots' not found in file %s." % cfnconfig_file)
```
enrico-usai: The message could be: "Required config **parameter** ...."

fnubalaj: Yes, that makes more sense.

jobwatcher/jobwatcher.py
```diff
@@ -51,17 +52,56 @@ def get_asg_name(stack_name, region, proxy_config):
...
+        cfn_scheduler_slots = cfnconfig_params["cfn_scheduler_slots"]
+        slots = 0
+        if cfn_scheduler_slots.isdigit():
+            slots = int(cfn_scheduler_slots)
+            log.info("Instance %s will use %s slots based on configuration." % (instance_type, slots))
+        else:
+            vcpus = get_vcpus_from_pricing_file(instance_type)
+            if cfn_scheduler_slots is "cores":
+                log.info("Instance %s will use number of cores as slots based on configuration." % instance_type)
+                slots = vcpus / 2
```
sean-smith: what if there's 1 vcpu?

jobwatcher/jobwatcher.py
```diff
@@ -51,17 +52,56 @@ def get_asg_name(stack_name, region, proxy_config):
...
             log.error("Instance %s not found in file %s." % (instance_type, pricing_file))
             exit(1)
 
+
+def get_instance_properties(instance_type):
+    cfnconfig_params = read_cfnconfig()
+    try:
+        cfn_scheduler_slots = cfnconfig_params["cfn_scheduler_slots"]
+        slots = 0
+        if cfn_scheduler_slots.isdigit():
```
sean-smith: I'd restructure this closer to a `switch` statement...
```
if cfn_scheduler_slots == ''vcpus":
...
elif cfn_scheduler_slots == ''vcpus":
...
else: # it's a digit or throw an error
```
Having nested if blocks makes it confusing to read.

jobwatcher/jobwatcher.py
fnubalaj: I could not find an instance with 1 vcpu? How does the cookbook handle that?

jobwatcher/jobwatcher.py
```diff
@@ -51,17 +52,56 @@ def get_asg_name(stack_name, region, proxy_config):
...
+                slots = vcpus / 2
+            elif cfn_scheduler_slots is "vcpus":
+                log.info("Instance %s will use number of vcpus as slots based on configuration." % instance_type)
+                slots = vcpus
+
+        if not slots > 0:
+            log.critical("cfn_scheduler_slots config value '%s' was invalid" % cfn_scheduler_slots)
+            exit(1)
+
+        return {'slots': slots}
```
sean-smith: Why return a dict versus an int?

jobwatcher/jobwatcher.py
fnubalaj: The problem with that is that we would either have 2 calls to get_vcpus_from_pricing_file or an unnecessary call in case it is a digit.

jobwatcher/jobwatcher.py
sean-smith: There's a couple single vcpu t2's. You should set the slots as `1` in that case.

https://aws.amazon.com/ec2/instance-types/

jobwatcher/jobwatcher.py
fnubalaj: Because the plugins use that. That was the original design.

jobwatcher/jobwatcher.py
fnubalaj: I am still interested in knowing what the cookbook does in that case. Maybe we are carrying a bug in the cookbook. But, I will make that change.

jobwatcher/jobwatcher.py
fnubalaj: On second thought, maybe code clarity is better than making an excess call to get_vcpus_from_pricing_file

fnubalaj: Done. This will be done in the next PR

jobwatcher/jobwatcher.py
fnubalaj: Done. This will be done in the next PR. Fixed the cookbook bug separately in https://github.com/awslabs/cfncluster-cookbook/pull/171

jobwatcher/jobwatcher.py
fnubalaj: Done. This will be done in the next commit

jobwatcher/jobwatcher.py
```diff
@@ -51,17 +52,56 @@ def get_asg_name(stack_name, region, proxy_config):
...
 
+
+def get_instance_properties(instance_type):
+    cfnconfig_params = read_cfnconfig()
+    try:
+        cfn_scheduler_slots = cfnconfig_params["cfn_scheduler_slots"]
+        slots = 0
+        vcpus = get_vcpus_from_pricing_file(instance_type)
+
+        if cfn_scheduler_slots is "cores":
```
fnubalaj: "is" is a test for identity not equality.  This is a bug and will be fixed in the next commit.

PR #64, opened 2018-09-30
Fix scaling up issues in torque and slurm

fnubalaj:
Scale up in 1.6.0rc1 through jobwatcher only took into account the number of nodes requested. This caused overscaling.
This fix takes into account both nodes and vcpus and packs the vcpus in a way that all the constraints during job submission are satisfied.


jobwatcher/plugins/torque.py
```diff
@@ -9,12 +9,16 @@ def get_required_nodes(instance_properties):
...
     output = _output.split("\n")[5:]
-    nodes = 0
+    slots_requested = 0
+    nodes_requested = []
     for line in output:
         line_arr = line.split()
         if len(line_arr) >= 10 and line_arr[9] in status:
-            nodes += int(line_arr[5])
-    return nodes
+            slots_requested += int(line_arr[5])
```
mauri-melato: It should be [6] here and [5] the line below.
Here the expected output of the `qstat -a` command:
```

ip-10-0-235-170.eu-west-1.compute.internal:
                                                                                  Req'd       Req'd       Elap
Job ID                  Username    Queue    Jobname          SessID  NDS   TSK   Memory      Time    S   Time
----------------------- ----------- -------- ---------------- ------ ----- ------ --------- --------- - ---------
8.ip-10-0-235-170.eu-w  centos      batch    job120.sh           --      1      1       --   01:00:00 Q       --
```

jobwatcher/plugins/utils.py
```diff
@@ -0,0 +1,39 @@
...
+    vcpus = instance_properties.get('slots')
+    nodes_required = 0
+    node_list_index = 0
+    while slots_requested > 0:
+        current_node_allocation = nodes_requested[node_list_index]
+        current_slot_allocation = vcpus * current_node_allocation
+        nodes_required += current_node_allocation
+        slots_requested -= current_slot_allocation
+        node_list_index += 1
+        if node_list_index >= len(slots_requested):
```
mauri-melato: `node_requested` ?

jobwatcher/plugins/utils.py
```diff
@@ -0,0 +1,39 @@
...
+        _output = process.communicate()[0]
+        return _output
+    except sub.CalledProcessError:
+        log.error("Failed to run %s\n" % _command)
+        exit(1)
+    finally:
+        DEV_NULL.close()
+
+
+def get_optimal_nodes(nodes_requested, slots_requested, instance_properties):
```
mauri-melato: I think this is going to under-scale in few scenarios that we may want to understand if we want to address, e.g.  in an env with 4 procs per node and 3 jobs submitted with the following requiremetns
job1: 4 nodes/2 proc per node
job2: 1 node / 4 proc
job3: 1 node/ 4 proc
Total # slots_required = 16 - so according to this approach the first 4 nodes would be enough when instead 6 nodes woudl be required.
The proposed 'filling jars' idea was coping with this, let's talk about this if it make sense

jobwatcher/plugins/slurm.py
```diff
@@ -1,19 +1,24 @@
...
+from utils import run_command, get_optimal_nodes
+
 
 log = logging.getLogger(__name__)
 
 # get nodes requested from pending jobs
 def get_required_nodes(instance_properties):
-    command = "/opt/slurm/bin/squeue -r -h -o '%t %D'"
+    command = "/opt/slurm/bin/squeue -r -h -o '%t %D %C'"
     _output = run_command(command, {})
```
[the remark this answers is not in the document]
fnubalaj: Done.

jobwatcher/plugins/torque.py
```diff
@@ -9,12 +9,16 @@ def get_required_nodes(instance_properties):
     status = ['Q']
     _output = run_command(command, {})
```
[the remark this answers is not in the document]
fnubalaj: Done.

jobwatcher/plugins/torque.py
fnubalaj: Done.

jobwatcher/plugins/utils.py
fnubalaj: Letting it be required for now.

jobwatcher/plugins/utils.py
fnubalaj: Changed the design

jobwatcher/plugins/torque.py
```diff
@@ -9,12 +9,16 @@ def get_required_nodes(instance_properties):
     status = ['Q']
```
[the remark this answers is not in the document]
fnubalaj: Done.

jobwatcher/plugins/utils.py
```diff
@@ -0,0 +1,56 @@
...
+    vcpus = instance_properties.get('slots')
+    slots_remaining_per_node = []
+
+    for node_idx, node in enumerate(nodes_requested):
+        log.info("Requested node %s with slots %s" % (node, slots_requested[node_idx]))
+        # for simplicity, uniformly distribute the numbers of cpus requested across all the requested nodes
+        slots_required_per_node = -(-slots_requested[node_idx] // node)
+
+        if slots_required_per_node > vcpus:
```
mauri-melato: `slots_required_per_node` is the result of the total number of required slots (cpus for us) of a job divided by the number of requested nodes for the same job. If this is greater than the capacity of our instance (`vcpus`) then the job simply will not be submitted and will remain pending in Slurm. Today this causes a continuous scale up till the max, with this patch we should raise the issue in the log and avoid to scale for those unsatisfiable jobs.

jobwatcher/plugins/torque.py
```diff
@@ -1,30 +1,44 @@
...
+            nodes_requested.append(int(line_arr[5]))
+
+    return get_optimal_nodes(nodes_requested, slots_requested, instance_properties)
+
 
 # get nodes reserved by running jobs
 def get_busy_nodes(instance_properties):
-    command = "/opt/torque/bin/qstat -a"
-    status = ['R']
+    command = "/opt/torque/bin/pbsnodes -x"
```
mauri-melato: Please add an output sample also this

jobwatcher/plugins/slurm.py
```diff
@@ -1,28 +1,37 @@
...
+            slots_requested.append(int(line_arr[3]))
+            nodes_requested.append(int(line_arr[2]))
+
+    return get_optimal_nodes(nodes_requested, slots_requested, instance_properties)
+
 
 # get nodes reserved by running jobs
 def get_busy_nodes(instance_properties):
-    command = "/opt/slurm/bin/squeue -r -h -o '%t %D'"
+    command = "/opt/slurm/bin/sinfo -r -h -o '%D %t'"
```
mauri-melato: Please add an output sample also this

jobwatcher/plugins/utils.py
```diff
@@ -0,0 +1,56 @@
...
+        _output = process.communicate()[0]
+        return _output
+    except sub.CalledProcessError:
+        log.error("Failed to run %s\n" % _command)
+        exit(1)
+    finally:
+        DEV_NULL.close()
+
+
+def get_optimal_nodes(nodes_requested, slots_requested, instance_properties):
```
mauri-melato: This method is worh and perfect for unit testing with multiple test-cases.

jobwatcher/plugins/torque.py
fnubalaj: Done

jobwatcher/plugins/slurm.py
fnubalaj: Done

nodewatcher/nodewatcher.cfg
```diff
@@ -2,4 +2,5 @@
 region = us-east-1
 scheduler = test
 proxy = NONE
-scaledown_idletime = 10
\ No newline at end of file
+scaledown_idletime = 10
+stack_name = test
```
demartinofra: is this a leftover from the tests or it's supposed to be here?

PR #87, opened 2018-12-18
Concurrency issues during scaledown

fnubalaj:
Use optimistic concurrency using hostname as a lock for dynamodb table item.

1. Create a dummy entry in the dynamodb table to keep track of current active nodes

2. To get the current number of active nodes, use information from scheduler instead of just the asg desired value, since the ASG desired value does not reflect bootstrapping.

3. Optimistically update the dummy entry using atomic counter during scaledown in order to prevent excessive termination

4. Keep a count of the total number of nodes as seen by the cluster, not just the ASG

5. For every entry in dynamoDB have a flag indicating if the termination is happening from the nodewatcher

6. If the termination happens outside the nodewatcher for some reason, sqswatcher will decrement the count

### DIFF
--- a/nodewatcher/nodewatcher.py
+++ b/nodewatcher/nodewatcher.py
@@ -18,15 +18,13 @@ import urllib2
 import os
 import time
 import sys
-import tempfile
 import logging
 import boto3
+from boto3.dynamodb.conditions import Attr
 import ConfigParser
 from botocore.exceptions import ClientError
 from botocore.config import Config
 import json
-import atexit
-import errno
 
 
 log = logging.getLogger(__name__)
@@ -34,9 +32,26 @@ _DATA_DIR = "/var/run/nodewatcher/"
 _IDLETIME_FILE = _DATA_DIR + "node_idletime.json"
 _CURRENT_IDLETIME = 'current_idletime'
 
-def getConfig(instance_id):
-    log.debug('reading /etc/nodewatcher.cfg')
 
+def get_asg_name(stack_name, region, proxy_config):
+    asg_conn = boto3.client('autoscaling', region_name=region, config=proxy_config)
+    asg_name = ""
+    no_asg = True
+
+    while no_asg:
+        try:
+            r = asg_conn.describe_tags(Filters=[{'Name': 'value', 'Values': [stack_name]}])
+            asg_name = r.get('Tags')[0].get('ResourceId')
+            no_asg = False
+        except IndexError as e:
+            log.error("No asg found for cluster %s" % stack_name)
+            time.sleep(30)
+
+    return asg_name
+
+
+def get_config():
+    log.debug('reading /etc/nodewatcher.cfg')
     config = ConfigParser.RawConfigParser()
     config.read('/etc/nodewatcher.cfg')
     if config.has_option('nodewatcher', 'loglevel'):
@@ -52,29 +67,14 @@ def getConfig(instance_id):
 
     _scaledown_idletime = int(config.get('nodewatcher', 'scaledown_idletime'))
     _stack_name = config.get('nodewatcher', 'stack_name')
-    try:
-        _asg = config.get('nodewatcher', 'asg')
-    except ConfigParser.NoOptionError:
-        ec2 = boto3.resource('ec2', region_name=_region, config=proxy_config)
-
-        instances = ec2.instances.filter(InstanceIds=[instance_id])
-        instance = next(iter(instances or []), None)
-        _asg = filter(lambda tag: tag.get('Key') == 'aws:autoscaling:groupName', instance.tags)[0].get('Value')
-        log.debug("discovered asg: %s" % _asg)
-        config.set('nodewatcher', 'asg', _asg)
-
-        tup = tempfile.mkstemp(dir=os.getcwd())
-        fd = os.fdopen(tup[0], 'w')
-        config.write(fd)
-        fd.close
-
-        os.rename(tup[1], 'nodewatcher.cfg')
+    _table_name = config.get('nodewatcher', 'table_name')
+    _asg = get_asg_name(_stack_name, _region, proxy_config)
 
     log.debug("region=%s asg=%s scheduler=%s prox_config=%s idle_time=%s" % (_region, _asg, _scheduler, proxy_config, _scaledown_idletime))
-    return _region, _asg, _scheduler, proxy_config, _scaledown_idletime, _stack_name
+    return _region, _asg, _scheduler, proxy_config, _scaledown_idletime, _stack_name, _table_name
 
-def getInstanceId():
 
+def get_instance_id():
     try:
         _instance_id = urllib2.urlopen("http://169.254.169.254/latest/meta-data/instance-id").read()
     except urllib2.URLError:
@@ -85,8 +85,90 @@ def getInstanceId():
 
     return _instance_id
 
-def getHostname():
 
+def load_scheduler_module(scheduler):
+    scheduler = 'nodewatcher.plugins.' + scheduler
+    _scheduler = __import__(scheduler)
+    _scheduler = sys.modules[scheduler]
+
+    log.debug("scheduler=%s" % repr(_scheduler))
+
+    return _scheduler
+
+
+def node_has_jobs(scheduler, hostname):
+
+    _jobs = scheduler.has_jobs(hostname)
+
+    log.info("jobs=%s" % _jobs)
+
+    return _jobs
+
+
+def get_idle_nodes(scheduler):
+    idle_nodes = scheduler.get_idle_nodes()
+    log.info("idle_node_count=%s" % len(idle_nodes))
+
+    return idle_nodes
+
+
+def queue_has_pending_jobs(scheduler):
+    _has_pending_jobs, _error = scheduler.has_pending_jobs()
+    log.info("has_pending_jobs=%s, error=%s" % (_has_pending_jobs, _error))
+    return _has_pending_jobs, _error
+
+
+def get_current_cluster_size(scheduler):
+    current_cluster_size = scheduler.get_current_cluster_size()
+    log.info("current_cluster_size=%s" % current_cluster_size)
+    return current_cluster_size
+
+
+def lock_host(scheduler, hostname):
+    log.info("locking %s" % hostname)
+
+    _r = scheduler.lock_host(hostname, False)
+
+    time.sleep(15) # allow for some settling
+
+    return _r
+
+
+def unlock_host(scheduler,hostname):
+    log.info("unlocking %s" % hostname)
+
+    _r = scheduler.lock_host(hostname, True)
+
+    time.sleep(15) # allow for some settling
+
+    return _r
+
+
+def self_terminate(dynamodb_table, asg_conn, instance_id):
+    log.info("terminating %s" % instance_id)
+    try:
+        dynamodb_table.update_item(
+            Key={'instanceId': instance_id},
+            UpdateExpression='SET #terminated_by_cluster = :terminated',
+            ExpressionAttributeNames={'#terminated_by_cluster': 'terminated_by_cluster'},
+            ExpressionAttributeValues={':terminated': True}
+        )
+        asg_conn.terminate_instance_in_auto_scaling_group(InstanceId=instance_id, ShouldDecrementDesiredCapacity=True)
+        return True
+    except ClientError as ex:
+        log.error("terminate_instance_in_auto_scaling_group failed with exception %s" % ex)
+    return False
+
+
+def get_stack_creation_status(stack_name, region, proxy_config):
+    log.info('Checking for status of the stack %s' % stack_name)
+    cfn_client = boto3.client('cloudformation', region_name=region, config=proxy_config)
+    stacks = cfn_client.describe_stacks(StackName=stack_name)
+    return stacks['Stacks'][0]['StackStatus'] == 'CREATE_COMPLETE' or \
+           stacks['Stacks'][0]['StackStatus'] == 'UPDATE_COMPLETE'
+
+
+def get_hostname():
     try:
         _hostname = urllib2.urlopen("http://169.254.169.254/latest/meta-data/local-hostname").read()
     except urllib2.URLError:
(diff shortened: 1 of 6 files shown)


nodewatcher/nodewatcher.py
```diff
@@ -183,57 +260,94 @@ def main():
...
-        has_jobs = hasJobs(s, hostname)
+        has_jobs = node_has_jobs(scheduler, hostname)
         if has_jobs:
             log.info('Instance has active jobs.')
             data[_CURRENT_IDLETIME] = 0
         else:
-            if maintainSize(asg_name, asg_conn):
+            min_size, current_size = get_desired_sizes(scheduler, asg_name, asg_conn)
+            if current_size <= min_size:
+                log.info('capacity less then or equal to min size. Node not eligible for termination')
```
enrico-usai: Should we reset the `_CURRENT_IDLETIME` in this case too?

nodewatcher/plugins/slurm.py
```diff
@@ -61,5 +114,6 @@ def hasPendingJobs():
 
     return has_pending, error
 
-def lockHost(hostname, unlock=False):
+
+def lock_host(hostname, unlock=False):
```
enrico-usai: We should fould a way to lock the host in the SLURM case too.

nodewatcher/plugins/sge.py
```diff
@@ -36,7 +37,53 @@ def hasJobs(hostname):
...
+    # ip-172-31-56-253.ec2.internal
+    # ip-172-31-56-252.ec2.internal
+    # ip-172-31-56-251.ec2.internal
+
+    _command = shlex.split(command)
+    try:
+        process = subprocess.Popen(_command, env=dict(os.environ, SGE_ROOT='/opt/sge',
+                                                      PATH='/opt/sge/bin:/opt/sge/bin/lx-amd64:/bin:/usr/bin'),
+                                   stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+    except subprocess.CalledProcessError:
```
enrico-usai: From the documentation it seems the CalledProcessError is raised only with `check_call` or `check_output` calls.

Then, what is the behaviour when the host is not attached to the cluster? 
Some sge commands fail with exit code 0 but with an error message, we should check the stderr for the commands.

nodewatcher/nodewatcher.py
```diff
@@ -183,57 +260,94 @@ def main():
...
 
-        has_jobs = hasJobs(s, hostname)
+        has_jobs = node_has_jobs(scheduler, hostname)
         if has_jobs:
             log.info('Instance has active jobs.')
             data[_CURRENT_IDLETIME] = 0
         else:
-            if maintainSize(asg_name, asg_conn):
+            min_size, current_size = get_desired_sizes(scheduler, asg_name, asg_conn)
+            if current_size <= min_size:
```
demartinofra: We should consider resetting CURRENT_IDLETIME in cases when the node is not marked for termination because the min_size of the cluster is reached or when there are pending jobs that are waiting to be submitted. Currently CURRENT_IDLETIME is kept at the previously computed value, meaning the idle time might not be guaranteed under certain circumstances (e.g. the CURRENT_IDLETIME was already bigger than threshold but node is not terminated because min_size is reached. In the following cycles this node will be terminated right away when cluster size grows)

nodewatcher/plugins/sge.py
```diff
@@ -36,7 +37,53 @@ def hasJobs(hostname):
...
+        log.error("Failed to run %s\n" % command)
+        return -1
+
+    output = process.communicate()[0]
+    current_cluster_size = len(filter(None, output.split("\n")))
```
enrico-usai: ` len(filter...` doesn't work.

nodewatcher/plugins/sge.py
demartinofra: Same applies everywhere else the Popen method is called

nodewatcher/nodewatcher.py
```diff
@@ -18,25 +18,40 @@
...
+            no_asg = False
+        except IndexError as e:
+            log.error("No asg found for cluster %s" % stack_name)
+            time.sleep(30)
+
+    return asg_name
+
+
+def get_config():
+    log.debug('reading /etc/nodewatcher.cfg')
```
enrico-usai: We should define an internal variable for the cfg file and use it in the log and in the read action.

nodewatcher/plugins/torque.py
```diff
@@ -51,7 +53,34 @@ def hasJobs(hostname):
...
+
+def get_idle_nodes():
+    command = "/opt/torque/bin/pbsnodes -l free"
+
+    # Command outputs list of free nodes
+    # ip-172-31-53-229     free
+    # ip-172-31-53-223     free
+    # ip-172-31-53-224     free
+
+    _command = shlex.split(command)
```
enrico-usai: We could create the command array avoiding to create and split a string.

sqswatcher/sqswatcher.py
```diff
@@ -216,6 +235,10 @@ def main():
     region, sqsqueue, table_name, scheduler, cluster_user, proxy_config = getConfig()
     q = setupQueue(region, sqsqueue, proxy_config)
     t = setupDDBTable(region, table_name, proxy_config)
+    t.put_item(Item={
```
demartinofra: This can be very dangerous in case the sqswatcher crashes and is restarted by the supervisord. In that case the current_cluster_size is set to 0 and the existing nodes will be never terminated.

nodewatcher/plugins/torque.py
```diff
@@ -18,7 +18,8 @@
 
 log = logging.getLogger(__name__)
 
-def runPipe(cmds):
+
+def run_pipe(cmds):
```
enrico-usai: To have a cleaner patch, it is better to split the function name refactoring from the new functionalities.

nodewatcher/plugins/torque.py
```diff
@@ -87,7 +116,35 @@ def hasPendingJobs():
...
+        log.error("Failed to run %s\n" % command)
+        return -1
+
+    output = process.communicate()[0]
+
+    # if this command did not include MasterServer something has gone wrong
+    if "MasterServer" not in output:
+        return -1
+
+    current_cluster_size = len(filter(None, output.split("\n"))) - 1
```
enrico-usai: Could we avoid the filter function?

sqswatcher/sqswatcher.py
```diff
@@ -185,10 +193,21 @@ def pollQueue(scheduler, q, t, proxy_config):
...
+                    log.info("Fetching the item from DynamoDB table")
                     try:
                         item = t.get_item(ConsistentRead=True, Key={"instanceId": instanceId})
                         if item.get('Item') is not None:
                             hostname = item.get('Item').get('hostname')
+                            terminated_by_cluster = item.get('Item').get('terminated_by_cluster')
+
+                            if not terminated_by_cluster:
+                                log.info("Instance %s was not terminated by the cluster" % instanceId)
+                                t.update_item(
```
enrico-usai: What about creating a new class to manage the db and move all the table management functions on it?

nodewatcher/nodewatcher.py
```diff
@@ -106,68 +95,156 @@ def loadSchedulerModule(scheduler):
...
+def update_table_when_termination_fails(dynamodb_table, instance_id):
+    log.info("Updating table since termination request on this node failed")
+    try:
+        dynamodb_table.update_item(
+            Key={'instanceId': instance_id},
+            UpdateExpression='SET #terminated_by_cluster = :terminated',
+            ExpressionAttributeNames={'#terminated_by_cluster': 'terminated_by_cluster'},
+            ExpressionAttributeValues={':terminated': False}
+        )
+        dynamodb_table.update_item(
```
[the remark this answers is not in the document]
demartinofra: Just be aware that when using transactions the number of reads and writes doubles, so we should take this into account when defining the capacity of the table: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/transaction-apis.html#transaction-capacity-handling

sqswatcher/sqswatcher.py
demartinofra: Read the value from the scheduler

nodewatcher/nodewatcher.py
```diff
@@ -106,68 +95,156 @@ def loadSchedulerModule(scheduler):
...
+    except urllib2.URLError:
+        log.critical('Unable to get hostname from metadata')
+        sys.exit(1)
+
+    log.debug("hostname=%s" % _hostname)
+
+    return _hostname
+
+
+def check_if_terminable(dynamodb_table, min_size):
```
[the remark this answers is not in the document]
fnubalaj: This function checks if the node can be terminated, hence the name.

nodewatcher/nodewatcher.py
```diff
@@ -106,68 +95,156 @@ def loadSchedulerModule(scheduler):
...
+            Key={'instanceId': 'current_cluster_size'},
+            UpdateExpression='SET #count = #count - :count',
+            ConditionExpression=Attr('count').gt(min_size),
+            ExpressionAttributeNames={'#count': 'count'},
+            ExpressionAttributeValues={':count': 1}
+        )
+        return True
+    except ClientError as ex:
+        log.info("Node is not terminable. Failed with ex " % ex)
+        item = dynamodb_table.get_item(ConsistentRead=True, Key={"instanceId": "current_cluster_size"})
```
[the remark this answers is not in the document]
fnubalaj: It should not raise an exception unless the table has been messed with. What we can do in this situation is stop the entire cluster. 

I think the cluster could potentially become unusable if this key is not found.

nodewatcher/nodewatcher.py
```diff
@@ -106,68 +95,156 @@ def loadSchedulerModule(scheduler):
...
+    log.info("current_cluster_size=%s" % current_cluster_size)
+    return current_cluster_size
+
+
+def lock_host(scheduler, hostname):
+    log.info("locking %s" % hostname)
+
+    _r = scheduler.lock_host(hostname, False)
 
     time.sleep(15) # allow for some settling
```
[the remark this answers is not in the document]
fnubalaj: What does this method return? What is _r and how is it being used? That might be a good resource.

nodewatcher/nodewatcher.py
```diff
@@ -106,68 +95,156 @@ def loadSchedulerModule(scheduler):
...
+        if item.get('Item') is not None:
+            count = item.get('Item').get('count')
+            log.info("Current count is %s; Minimum size is %s" % (count, min_size))
+        return False
+
+
+def update_table_when_termination_fails(dynamodb_table, instance_id):
+    log.info("Updating table since termination request on this node failed")
+    try:
+        dynamodb_table.update_item(
```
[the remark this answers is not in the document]
fnubalaj: I don't think that should happen. 

> Provisioned throughput is the maximum amount of capacity that an application can consume from a table or index. If your application exceeds your provisioned throughput capacity on a table or index, it is subject to request throttling.
> 
> Throttling prevents your application from consuming too many capacity units. When a request is throttled, it fails with an HTTP 400 code (Bad Request) and a ProvisionedThroughputExceededException. The AWS SDKs have built-in support for retrying throttled requests (see Error Retries and Exponential Backoff), so you do not need to write this logic yourself.
> 
> 

Maybe we can move from provisioned to on-demand, in that case there needs to be an conversation about the increase in implicit pricing.

nodewatcher/nodewatcher.py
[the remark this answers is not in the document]
fnubalaj: Transactions is a good idea.

nodewatcher/nodewatcher.py
```diff
@@ -106,68 +95,156 @@ def loadSchedulerModule(scheduler):
...
+    except ClientError as ex:
+        log.error("Update table failed with  exception " % ex)
+
+
+def get_desired_sizes(scheduler, asg_name, asg_conn):
+    asg = asg_conn.describe_auto_scaling_groups(AutoScalingGroupNames=[asg_name]) \
+        .get('AutoScalingGroups')[0]
+    asg_desired_capacity = asg.get('DesiredCapacity')
+    log.info("asg_desired=%d" % asg_desired_capacity)
+    total_compute_nodes = get_current_cluster_size(scheduler)
```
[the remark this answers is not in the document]
fnubalaj: That shouldn't happen correct? If this node is not attached to the scheduler correctly, there is no point trying to terminate it. It cannot be terminated through this logic?

nodewatcher/nodewatcher.py
```diff
@@ -183,57 +260,94 @@ def main():
...
+                                        os.remove(_IDLETIME_FILE)
+                                else:
+                                    unlock_host(scheduler, hostname)
+
                             except ClientError as ex:
                                 log.error('Failed to terminate instance: %s with exception %s' % (instance_id, ex))
-                                lockHost(s, hostname, unlock=True)
+                                unlock_host(scheduler, hostname)
                         else:
                             if has_pending_jobs:
```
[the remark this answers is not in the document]
fnubalaj: That should work.

sqswatcher/sqswatcher.py
fnubalaj: Yes, Maurizio and I discussed this over. We should write the current_value to the file or something, but sqswatcher crashing implies no new nodes can join, and the nodes can still terminate outside of CFNCluster ecosystem and the event is not read by the  master. If SQSWatcher crashes, there might be a lot of instability in the system anyways.

nodewatcher/nodewatcher.py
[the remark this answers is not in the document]
demartinofra: It could also fail because dynamo table limits are hit (unless we switch to ondemand)

enrico-usai:
Thank you @fnubalaj for your help and feedbacks! We'll look after this patch.

nodewatcher/nodewatcher.py
```diff
@@ -18,25 +18,40 @@
...
+    no_asg = True
+
+    while no_asg:
+        try:
+            r = asg_conn.describe_tags(Filters=[{'Name': 'value', 'Values': [stack_name]}])
+            asg_name = r.get('Tags')[0].get('ResourceId')
+            no_asg = False
+        except IndexError as e:
+            log.error("No asg found for cluster %s" % stack_name)
+            time.sleep(30)
```
enrico-usai: If the asg name is not found the node will remain up forever. We need a timeout and raise the exception to the parent.

nodewatcher/plugins/sge.py
enrico-usai: Fixed with #116

nodewatcher/plugins/slurm.py
enrico-usai: Fixed with https://github.com/aws/aws-parallelcluster-node/pull/111

sean-smith:
Closing. Should we decide to revisit, we can re-open this pr.

PR #91, opened 2019-01-07
Add UPDATE_ROLLBACK_COMPLETE to stack completion statuses

sean-smith:
See https://github.com/aws/aws-parallelcluster/issues/800

### DIFF
--- a/nodewatcher/nodewatcher.py
+++ b/nodewatcher/nodewatcher.py
@@ -152,9 +152,7 @@ def stackCreationComplete(stack_name, region, proxy_config):
     log.info('Checking for status of the stack %s' % stack_name)
     cfn_client = boto3.client('cloudformation', region_name=region, config=proxy_config)
     stacks = cfn_client.describe_stacks(StackName=stack_name)
-    return stacks['Stacks'][0]['StackStatus'] == 'CREATE_COMPLETE' or \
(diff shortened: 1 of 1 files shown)


nodewatcher/nodewatcher.py
```diff
@@ -152,9 +152,7 @@ def stackCreationComplete(stack_name, region, proxy_config):
     log.info('Checking for status of the stack %s' % stack_name)
     cfn_client = boto3.client('cloudformation', region_name=region, config=proxy_config)
     stacks = cfn_client.describe_stacks(StackName=stack_name)
-    return stacks['Stacks'][0]['StackStatus'] == 'CREATE_COMPLETE' or \
-           stacks['Stacks'][0]['StackStatus'] == 'UPDATE_COMPLETE'
-
+    return stacks['Stacks'][0]['StackStatus'] in ['CREATE_COMPLETE', 'UPDATE_COMPLETE', 'UPDATE_ROLLBACK_COMPLETE']
```
enrico-usai: I think we should rename the function from `stackCreationComplete` to something like `stackReady` or `stackComplete` to enclose the three cases. 

We should also rename the `stack_creation_complete` variable in the main.

sean-smith: done

PR #94, opened 2019-01-25
Avoid to process multiple times a bad host

enrico-usai:
If the hostname was empty or when an exception was raised the related message was not removed from the queue.

The _sqswatcher_ process crashed with an exception and a new sqswatcher process started.

The bad message remained in the queue and after the [SQS visibility timeout](https://docs.aws.amazon.com/en_us/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html) it was processed by the new sqswatcher process, crashing again.

Now we are logging the error, breaking the loop and deleting the message from the queue, for both COMPUTE_READY 417e48015cc1d08a1bfd70a8136c482eac9cee8e and EC2_INSTANCE_TERMINATE  c49e0672797ace0da96a4e5e8bf07775bf2c671b cases.

If a bad host is not added to the cluster, it will be be automatically stopped after the scaledown_idletime
I fixed the sge behaviour with aff0f3121e8a2bf72f889ce72d84460a94b183c3.

Issue: https://github.com/aws/aws-parallelcluster/issues/743

I tested it with:
- sge centos7 
- slurm ubuntu1604
- torque alinux

### DIFF
--- a/nodewatcher/nodewatcher.py
+++ b/nodewatcher/nodewatcher.py
@@ -11,23 +11,18 @@
 # OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 # limitations under the License.
 
-__author__ = 'dougalb'
-
-from datetime import datetime
-import urllib2
+import ConfigParser
+import json
+import logging
 import os
-import time
 import sys
 import tempfile
-import logging
-import boto3
-import ConfigParser
-from botocore.exceptions import ClientError
-from botocore.config import Config
-import json
-import atexit
-import errno
+import time
+import urllib2
 
+import boto3
+from botocore.config import Config
+from botocore.exceptions import ClientError
 
 log = logging.getLogger(__name__)
 _DATA_DIR = "/var/run/nodewatcher/"
@@ -171,8 +166,7 @@ def main():
         if not os.path.exists(_DATA_DIR):
             os.makedirs(_DATA_DIR)
     except OSError as ex:
-        log.critical('Creating directory %s to persist current idle time failed with exception: %s '
-                     % (_DATA_DIR, ex))
+        log.critical('Creating directory %s to persist current idle time failed with exception: %s ' % (_DATA_DIR, ex))
         raise
 
     if os.path.isfile(_IDLETIME_FILE):
--- a/sqswatcher/sqswatcher.py
+++ b/sqswatcher/sqswatcher.py
@@ -127,7 +127,7 @@ def pollQueue(scheduler, q, t, proxy_config):
                     try:
                         eventType = message_attrs.get('detail-type')
                     except KeyError:
-                        log.warn("Unable to read message. Deleting.")
+                        log.warning("Unable to read message. Deleting.")
                         message.delete()
                         break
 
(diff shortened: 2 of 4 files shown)


sqswatcher/sqswatcher.py
```diff
@@ -148,29 +148,33 @@ def pollQueue(scheduler, q, t, proxy_config):
...
+                                wait = wait * 2
                             else:
-                                raise e
+                                log.critical(e.response.get("Error").get("Message"))
+                                break
                         except:
                             log.critical("Unexpected error: %s" % sys.exc_info()[0])
-                            raise
+                            break
+                    message.delete()
```
sean-smith: This effectively adds a re-try for any message that fails. Can you also add the retry logic to `autoscaling:EC2_INSTANCE_TERMINATE`? Say for example the dynamo call fails, we should re-try.

nodewatcher/nodewatcher.py
```diff
@@ -11,23 +11,18 @@
...
+import ConfigParser
+import json
+import logging
 import os
-import time
 import sys
 import tempfile
-import logging
+import time
+import urllib2
```
sean-smith: `urllib2` is just a python 2 package. Breaks interoperability. See https://stackoverflow.com/questions/2792650/import-error-no-module-name-urllib2

sqswatcher/sqswatcher.py
```diff
@@ -122,87 +118,106 @@ def pollQueue(scheduler, q, t, proxy_config):
...
+                        event_type = message_attrs.get('detail-type')
                     except KeyError:
-                        log.warn("Unable to read message. Deleting.")
+                        log.warning("Unable to read message. Deleting.")
                         message.delete()
                         break
 
-                log.info("eventType=%s" % eventType)
-                if eventType == 'autoscaling:TEST_NOTIFICATION':
+                log.info("eventType=%s" % event_type)
```
sean-smith: `log.info("event_type=%s"` to match new case

sqswatcher/sqswatcher.py
```diff
@@ -122,87 +118,106 @@ def pollQueue(scheduler, q, t, proxy_config):
...
+                if event_type == 'autoscaling:TEST_NOTIFICATION':
                     message.delete()
-                elif eventType == 'parallelcluster:COMPUTE_READY':
-                    instanceId = message_attrs.get('EC2InstanceId')
+                elif event_type == 'parallelcluster:COMPUTE_READY':
+                    instance_id = message_attrs.get('EC2InstanceId')
                     slots = message_attrs.get('Slots')
-                    log.info("instanceId=%s" % instanceId)
+                    log.info("instanceId=%s" % instance_id)
```
sean-smith: same

sqswatcher/sqswatcher.py
```diff
@@ -122,87 +118,106 @@ def pollQueue(scheduler, q, t, proxy_config):
...
+                                if hostname:
+                                    log.info("Adding hostname: %s" % hostname)
+                                    scheduler_module.addHost(hostname=hostname, cluster_user=cluster_user, slots=slots)
+
+                                    table.put_item(Item={
+                                        'instanceId': instance_id,
+                                        'hostname': hostname
+                                    })
+                                else:
+                                    log.error("Unable to get the hostname for the instance %s." % instance_id)
```
sean-smith: So here when it can't get the hostname, it deletes the sqs message and continues?

nodewatcher/nodewatcher.py
enrico-usai: I just re-ordered the imports, `urllib2` was already there. 
I think this can be addressed in another patch, if required.

sqswatcher/sqswatcher.py
enrico-usai: done

sqswatcher/sqswatcher.py
enrico-usai: done

sqswatcher/sqswatcher.py
```diff
@@ -122,87 +118,106 @@ def pollQueue(scheduler, q, t, proxy_config):
...
 
-                    log.info("instanceId=%s" % instanceId)
-                    try:
-                        item = t.get_item(ConsistentRead=True, Key={"instanceId": instanceId})
-                        if item.get('Item') is not None:
-                            hostname = item.get('Item').get('hostname')
+                    log.info("instance_id=%s" % instance_id)
+                    retry = 0
+                    wait = 15
+                    while retry < 3:
```
demartinofra: Shall we export the retry logic and the error handling logic to a common function since it is the same for all event types? This can be a good reason to start breaking this huge function into smaller ones :)

Another nice way to implement retries in python is with annotations as also implemented by this library: https://github.com/rholder/retrying

enrico-usai: Done

PR #96, opened 2019-01-31
Improve nodewatcher log

lukeseawalker:
Instance id and hostname are added at the beginning of the log file

### DIFF
--- a/nodewatcher/nodewatcher.py
+++ b/nodewatcher/nodewatcher.py
@@ -29,8 +29,9 @@ _DATA_DIR = "/var/run/nodewatcher/"
 _IDLETIME_FILE = _DATA_DIR + "node_idletime.json"
 _CURRENT_IDLETIME = 'current_idletime'
 
+
 def getConfig(instance_id):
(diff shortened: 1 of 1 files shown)


nodewatcher/nodewatcher.py
```diff
@@ -149,14 +158,15 @@ def stackReady(stack_name, region, proxy_config):
...
+
 def main():
     logging.basicConfig(
         level=logging.INFO,
         format='%(asctime)s %(levelname)s [%(module)s:%(funcName)s] %(message)s'
     )
-    log.info("nodewatcher startup")
     instance_id = getInstanceId()
     hostname = getHostname()
+    log.info('nodewatcher startup for instance %s with hostname %s' % (instance_id, hostname))
```
enrico-usai: I'd split this log in two different logs:
```
log.info("nodewatcher startup")
instance_id = ...
hostname = ...
log.info('Instance = %s, hostname = %s' % (instance_id, hostname))
```
This because the _startup_ log is useful to understand when the nodewatcher has been restarted and if the `getInstanceId` or the `getHostname` fail we don't have this information.

PR #97, opened 2019-02-01
Simplify the logic to detect if a sge compute node has jobs running

lukeseawalker:
This patch revert https://github.com/aws/aws-parallelcluster-node/commit/1adb8579fe5a7f6db700a01ace958a2a5dca014d
in which the check was buggy because it was checking a hostname as a
substring of another hostname

This new code also works for parallel jobs and align its behaviour to
what is done for the other schedulers

It has been tested with normal job, array job and parallel job.

### DIFF
--- a/nodewatcher/plugins/sge.py
+++ b/nodewatcher/plugins/sge.py
@@ -18,8 +18,19 @@ log = logging.getLogger(__name__)
 
 
 def hasJobs(hostname):
-    # Checking for running jobs on the node
+    # Checking for running jobs on the node, with parallel job view expanded (-g t)
     _command = ['/opt/sge/bin/lx-amd64/qstat', '-g', 't', '-l', 'hostname=%s' % hostname, '-u', '*']
+
+    # Command output
+    # job-ID  prior   name       user         state submit/start at     queue                          master ja-task-ID
+    # ------------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


nodewatcher/plugins/sge.py
```diff
@@ -19,31 +19,23 @@
...
-                if hostname.split('.')[0] in host:
-                    _jobs = False
-                    break
+                                  ),
+                                  ).communicate()[0]
     except subprocess.CalledProcessError:
-        log.error("Failed to run %s. Assuming no active jobs." % _command)
+        print ("Failed to run %s\n" % _command)
+
+    if _output == "":
```
demartinofra: If subprocess.Popen fails with exception _ouput is undefined

enrico-usai: There is the same problem with the other schedulers...

lukeseawalker: Thanks, updated

nodewatcher/plugins/sge.py
```diff
@@ -19,31 +19,24 @@
 
 def hasJobs(hostname):
     # Checking for running jobs on the node
```
enrico-usai: I'd add a comment in the code with the expected output of the command and a short explanation of the `-g t` option.

PR #99, opened 2019-02-12
Prevent jobwatcher crashes by using fallback values

enrico-usai:
I removed the `exit(1)` calls because the jobwatcher is managed by the `supervisord` so they are useless.

If the cfn_scheduler_slots configuration parameter is wrong we can  fallback to the `vcpus` default value, log the error and continue to work.

If we are unable to get the number of slots associated to the given instance type the error is critical and the scale up will be blocked.

I also performed minor refactoring in different commits.

### cfn_scheduler_slots error tests

**cfn_scheduler_slots=-1**
```
2019-02-12 11:48:41,205 ERROR [jobwatcher:_get_instance_properties] cfn_scheduler_slots config parameter '-1' is invalid. Assuming 'vcpus'
```

**cfn_scheduler_slots=0**
```
2019-02-12 12:06:46,322 ERROR [jobwatcher:_get_instance_properties] cfn_scheduler_slots config parameter '0' must be greater than 0. Assuming 'vcpus'
```

**cfn_scheduler_slots=test**
```
2019-02-12 11:50:47,821 ERROR [jobwatcher:_get_instance_properties] cfn_scheduler_slots config parameter 'test' is invalid. Assuming 'vcpus'
```

**cfn_scheduler_slots removed from the config file**
```
2019-02-12 11:51:47,882 ERROR [jobwatcher:_get_instance_properties] Required config parameter 'cfn_scheduler_slots' not found in file /opt/parallelcluster/cfnconfig. Assuming 'vcpus'
```

### instance.json tests

**slots number changed to -1 in the instances.json file**
```
2019-02-12 11:58:36,765 CRITICAL [jobwatcher:_get_instance_properties] slots value is invalid. Setting it to 0.
2019-02-12 11:58:36,765 CRITICAL [jobwatcher:main] Error detecting number of slots per instance. The cluster will not scale up.
```

**instance type removed from the instance.json file**
```
2019-02-12 12:16:46,916 ERROR [jobwatcher:_get_vcpus_from_pricing_file] Unable to get vcpus from file /opt/parallelcluster/instances.json. Instance type c4.large not found.
2019-02-12 12:16:46,917 INFO [jobwatcher:_get_instance_properties] Instance c4.large will use number of vcpus as slots based on configuration.
2019-02-12 12:16:46,917 CRITICAL [jobwatcher:_get_instance_properties] slots value is invalid. Setting it to 0.
2019-02-12 12:16:46,917 CRITICAL [jobwatcher:main] Error detecting number of slots per instance. The cluster will not scale up.
```

### DIFF
--- a/jobwatcher/jobwatcher.py
+++ b/jobwatcher/jobwatcher.py
@@ -227,14 +227,23 @@ def main():
 
             # Check to make sure requested number of instances is within ASG limits
             required = running + pending
-            if required > max_size:
-                log.info("%d requested nodes is greater than max %d. Requesting max %d." % (required, max_size, max_size))
-                asg_client.update_auto_scaling_group(AutoScalingGroupName=asg_name, DesiredCapacity=max_size)
-            elif required <= current_desired:
+            if required <= current_desired:
                 log.info("%d nodes desired %d nodes in asg. Noop" % (required, current_desired))
             else:
-                log.info("Setting desired to %d nodes, requesting %d more nodes from asg." % (required, required - current_desired))
-                asg_client.update_auto_scaling_group(AutoScalingGroupName=asg_name, DesiredCapacity=required)
+                if required > max_size:
+                    log.info(
+                        "%d requested nodes is greater than max %d. Requesting max %d."
+                        % (required, max_size, max_size)
+                    )
+                else:
+                    log.info(
+                        "Setting desired to %d nodes, requesting %d more nodes from asg."
+                        % (required, required - current_desired)
+                    )
+                requested = min(required, max_size)
+
+                # update ASG
+                asg_client.update_auto_scaling_group(AutoScalingGroupName=asg_name, DesiredCapacity=requested)
 
         time.sleep(60)
 
--- a/jobwatcher/jobwatcher.py
+++ b/jobwatcher/jobwatcher.py
@@ -88,17 +88,18 @@ def _get_vcpus_from_pricing_file(instance_type):
     Read pricing file and get number of vcpus for the given instance type.
 
     :param instance_type: the instance type to search for.
-    :return: the number of vcpus
+    :return: the number of vcpus or -1 if the instance type cannot be found
     """
     with open(pricing_file) as f:
         instances = json.load(f)
         try:
             vcpus = int(instances[instance_type]["vcpus"])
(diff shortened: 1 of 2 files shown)


jobwatcher/jobwatcher.py
```diff
@@ -176,27 +214,27 @@ def main():
...
 
-            # check to make sure it's in limits
-            desired = running + pending
-            if desired > max:
-                log.info("%d requested nodes is greater than max %d. Requesting max %d." % (desired, max, max))
-                asg_conn.update_auto_scaling_group(AutoScalingGroupName=asg_name, DesiredCapacity=max)
-            elif desired <= current_desired:
-                log.info("%d nodes desired %d nodes in asg. Noop" % (desired, current_desired))
+            # Check to make sure requested number of instances is within ASG limits
+            required = running + pending
```
lukeseawalker: I would prefer to keep "desired" (or new_desired) because we use this value to set the ASG desired capacity, and rather change the log string accordingly

jobwatcher/jobwatcher.py
```diff
@@ -64,61 +83,94 @@ def read_cfnconfig():
...
-            slots = int(cfn_scheduler_slots)
-            log.info("Instance %s will use %s slots based on configuration." % (instance_type, slots))
+    else:
+        log.error("cfn_scheduler_slots config parameter '%s' is invalid. Assuming 'vcpus'" % cfn_scheduler_slots)
+        slots = vcpus
 
-        if not slots > 0:
-            log.critical("cfn_scheduler_slots config parameter '%s' was invalid" % cfn_scheduler_slots)
-            exit(1)
+    if slots <= 0:
```
demartinofra: minor: you can just check < since you are going to set it to 0 in any case.

jobwatcher/jobwatcher.py
enrico-usai: I used `required` in place of `desired` because it is the value that we need, but it will be used for the ASG only if it is lower than `max_size`.

jobwatcher/jobwatcher.py
enrico-usai: done

jobwatcher/jobwatcher.py
enrico-usai: I changed the log messages to reflect the variable names.

PR #108, opened 2019-03-08
Cleanup and refactor daemons code

enrico-usai:
All the daemons:
* Use namedtuple to manage configuration parameters

nodewatcher: 
* do not remove idletime file before self termination. 
If there is an error during the self termination phase we need the idletime file to try to remove the instance in the next daemon execution.
If the instance is deleted correctly the file removal is useless.
* create functions to manage idletime file.
* simplify code by using idletime variable and not a json object.
* fix internal variable names

jobwatcher and nodewatcher:
* Move `get_asg_name` function to `common/utils`. 
From now the function will raise an exception after a timeout if the ASG name can't be found (there was an infinite loop).
In a future patch we have to catch and manage the different exceptions

sqswatcher and jobwatcher:
* avoid global variables

### DIFF
--- a/jobwatcher/jobwatcher.py
+++ b/jobwatcher/jobwatcher.py
@@ -1,8 +1,9 @@
-#!/usr/bin/env python2.6
+#!/usr/bin/env python
 
-# Copyright 2013-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Copyright 2013-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #
-# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the
 # License. A copy of the License is located at
 #
 # http://aws.amazon.com/apache2.0/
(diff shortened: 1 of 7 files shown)


nodewatcher/nodewatcher.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2.6
+#!/usr/bin/env python
```
demartinofra: why are we changing this only in nodewatcher?

nodewatcher/nodewatcher.py
```diff
@@ -236,7 +236,6 @@ def main():
                     else:
                         has_pending_jobs, error = _has_pending_jobs(scheduler_module)
                         if not error and not has_pending_jobs:
-                            os.remove(idletime_file)
                             try:
                                 _self_terminate(asg_name, asg_conn, instance_id)
```
demartinofra: probably not related to this patch but just noticed and wanted to mention: _self_terminate might not terminate the instance if it became essential in the meanwhile but at the same time termination_in_progress is set to True right after meaning this instance will never terminate...

jobwatcher/jobwatcher.py
```diff
@@ -189,7 +190,11 @@ def _get_config():
         "Configured parameters: region=%s scheduler=%s stack_name=%s instance_type=%s pcluster_dir=%s proxy=%s",
         region, scheduler, stack_name, instance_type, pcluster_dir, _proxy
     )
-    return region, scheduler, stack_name, instance_type, pcluster_dir, proxy_config
+
+    JobwatcherConfig = collections.namedtuple(
```
demartinofra: since this is like defining a class we could also have it defined outside of the function scope

nodewatcher/nodewatcher.py
```diff
@@ -168,6 +168,52 @@ def _is_stack_ready(stack_name, region, proxy_config):
...
+    Initialize idletime value (from file if there).
+
+    :param idletime_file: the file to search for
+    :return: the current idletime value (0 if the file doesn't exist)
+    """
+    idletime = 0
+    if os.path.isfile(idletime_file):
+        with open(idletime_file) as f:
+            data = json.loads(f.read())
+            idletime = data["current_idletime"]
```
demartinofra: data.get("current_idletime", 0)?

nodewatcher/nodewatcher.py
enrico-usai: Thank you for highlighting it. It could be part of a future patch.

nodewatcher/nodewatcher.py
enrico-usai: changed in both

jobwatcher/jobwatcher.py
enrico-usai: Done

nodewatcher/nodewatcher.py
enrico-usai: fixed

PR #111, opened 2019-03-19
Slurm: job dependencies and locking logic

demartinofra:
* Implement locking/unlocking logic for Slurm compute nodes
* Added logic to use pending job reason when computing the number of nodes needed to serve all submitted jobs. This fixes the following:
  * when a job is pending because a dependency the cluster won’t scale up
  * when a job exceeds the number of max nodes supported by the cluster (max_queue_size configured by the user) the cluster won’t scale up. Currently what we do is to set the ASG at the max although the job would never be scheduled.
* Check pending jobs in scheduler queue before locking the compute node and checking running job on the specific host. This solves the following issue:
  * nodewatcher locks the host to check for pending and running jobs
  * in the meanwhile jobwatcher checks if cluster requires scaling, sees the nodes as locked (terminating), and requires additional capacity.
  * nodewatcher unlocks the host since it finds pending jobs in the queue.  
* Remove code to handle unregistered "EC2 Instance State-change Notification" CW event. This code was also bugged and had some dead code branches.

### DIFF
--- a/nodewatcher/plugins/slurm.py
+++ b/nodewatcher/plugins/slurm.py
@@ -63,4 +63,28 @@ def hasPendingJobs():
 
 
 def lockHost(hostname, unlock=False):
-    pass
+    # hostname format: ip-10-0-0-114.eu-west-1.compute.internal
+    hostname = hostname.split(".")[0]
+    if unlock:
+        log.info("Unlocking host %s", hostname)
+        command = [
+            "/opt/slurm/bin/scontrol",
+            "update",
+            "NodeName={0}".format(hostname),
+            "State=RESUME",
+            'Reason="Unlocking"',
+        ]
+    else:
+        log.info("Locking host %s", hostname)
+        command = [
+            "/opt/slurm/bin/scontrol",
+            "update",
+            "NodeName={0}".format(hostname),
+            "State=DRAIN",
+            'Reason="Shutting down"',
+        ]
+    try:
+        subprocess.check_call(command, env=dict(os.environ))
+    except subprocess.CalledProcessError as e:
+        # CalledProcessError.__str__ already produces a significant error message
+        log.error(e)
--- a/nodewatcher/plugins/slurm.py
+++ b/nodewatcher/plugins/slurm.py
@@ -14,52 +14,50 @@ import logging
 import shlex
 import os
 
+from common.slurm import PENDING_RESOURCES_REASONS
+
 log = logging.getLogger(__name__)
+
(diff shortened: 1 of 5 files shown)


sqswatcher/sqswatcher.py
```diff
@@ -110,35 +110,22 @@ def _setup_ddb_table(region, table_name, proxy_config):
...
-                {
-                    'AttributeName': 'instanceId',
-                    'AttributeType': 'S'
-                }
-            ],
-            ProvisionedThroughput={
-                'ReadCapacityUnits': 5,
-                'WriteCapacityUnits': 5
-            },
+            KeySchema=[{"AttributeName": "instanceId", "KeyType": "HASH"}],
```
enrico-usai: It is not part of your patch but I think we should remove this code, because we are creating the table in the CFN template

demartinofra: As agreed we'll handle this as part of a separate PR

nodewatcher/plugins/slurm.py
```diff
@@ -14,52 +14,50 @@
...
+from common.slurm import PENDING_RESOURCES_REASONS
+
 log = logging.getLogger(__name__)
 
 
+def _run_command(command):
+    try:
+        if isinstance(command, str):
+            command = shlex.split(command)
+        return subprocess.check_output(command, env=dict(os.environ), universal_newlines=True)
```
enrico-usai: `check_output` is not available in python 2.6.

You can use `from future.moves.subprocess import check_output`
http://python-future.org/compatible_idioms.html#subprocess-check-output

sqswatcher/sqswatcher.py
```diff
@@ -308,8 +289,9 @@ def _poll_queue(scheduler, region, cluster_user, queue, table, proxy_config, max
                         log.info("Unable to query scheduler configuration, discarding %s message" % event_type)
 
                     message.delete()
-
-            results = queue.receive_messages(MaxNumberOfMessages=10)
```
enrico-usai: `while len(results) > 0:` will continue to work with the same `results`.

demartinofra: Good catch! this part of the logic will change completely as part of the upcoming PRs so I'll just leave it as is.

nodewatcher/plugins/slurm.py
demartinofra: 👍👍👍 Thanks!!!

PR #112, opened 2019-03-19
sqswatcher: Improve sqs and dynamodb initialization

enrico-usai:
+ Do not create DynamoDB table if unexistant. The table is created by the CFN template.
+ Rename `_setup_*` internal functions to `_get_*`
+ Improve logging
+ Add try/catch for `ClientError`
+ Add retry

### DIFF
--- a/sqswatcher/sqswatcher.py
+++ b/sqswatcher/sqswatcher.py
@@ -37,6 +37,10 @@ class QueryConfigError(Exception):
     pass
 
 
+class CriticalError(Exception):
+    pass
+
+
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


sqswatcher/sqswatcher.py
```diff
@@ -99,15 +104,19 @@ def _setup_queue(region, queue_name, proxy_config):
...
+        log.critical("Unable to get the SQS queue '%s'. Failed with exception: %s", queue_name, e)
+        raise
 
-    _queue = sqs.get_queue_by_name(QueueName=queue_name)
-    return _queue
+    return queue
 
 
-def _setup_ddb_table(region, table_name, proxy_config):
+@retry(stop_max_attempt_number=3, wait_fixed=5000)
```
demartinofra: you probably don't want to retry on CriticalErrors. This should work:
```
@retry(
...
retry_on_exception=lambda exception: not isinstance(exception, CriticalError)
)
```

enrico-usai: fixed

PR #115, opened 2019-03-20
Sqswatcher: dynamically update cluster size and remove calls to describe_instances

demartinofra:
* Dynamically update the number of configured dummy nodes by retrieving the max capacity from ASG
  * This allows to change cluster size in ASG console or with an update
* Retrieve compute ready from SQS message and drop expensive call to ec2.describe_instances
  * This brought an additional improvements in the scaling speed: the scheduler is able to follow the state of the ASG with a delay of less than 1 minute during both scale-up and scale-down. (tested with a cluster of 200 nodes)

### DIFF
--- a/sqswatcher/sqswatcher.py
+++ b/sqswatcher/sqswatcher.py
@@ -186,13 +186,7 @@ def _retrieve_all_sqs_messages(queue):
     return messages
 
 
-def _parse_sqs_messages(messages, sqs_config, table):
-    instance_ids = _get_compute_ready_instance_ids(messages)
-    hostnames = _get_hostnames(instance_ids, sqs_config.region, sqs_config.proxy_config)
-    if not hostnames:
-        # cannot process messages, bailing out.
-        return []
-
+def _parse_sqs_messages(messages, table):
     update_events = OrderedDict()
     for message in messages:
         message_text = json.loads(message.body)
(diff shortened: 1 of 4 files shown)


sqswatcher/sqswatcher.py
```diff
@@ -139,7 +144,7 @@ def _setup_ddb_table(region, table_name, proxy_config):
 def _retry_on_request_limit_exceeded(func):
     @retry(
         stop_max_attempt_number=3,
-        wait_exponential_multiplier=15000,
+        wait_exponential_multiplier=5000,
```
enrico-usai: We could increase the max_attempt_number since we are decreasing the multiplier.

sqswatcher/sqswatcher.py
```diff
@@ -264,12 +248,14 @@ def _process_instance_terminate_event(message_attrs, sqs_config, message, table)
         return None
 
 
-def _process_sqs_messages(update_events, scheduler_module, sqs_config, table, queue):
-    if not update_events:
+def _process_sqs_messages(
+    update_events, scheduler_module, sqs_config, table, queue, max_cluster_size, update_max_cluster_size
+):
+    if not update_events and not update_max_cluster_size:
```
enrico-usai: I'd add a comment in this.

sqswatcher/sqswatcher.py
demartinofra: increased to 5

sqswatcher/sqswatcher.py
demartinofra: done

PR #127, opened 2019-04-12
slurm: add other reasons for the cluster scale up

enrico-usai:
We were scaling up only when the job was waiting for resources to become available ([Job reason code](https://slurm.schedmd.com/squeue.html) = **Resources**).

In case of a requeued job the cluster didn't scale up because the pending reason is **NodeDown** and then **BeginTime**. I'm adding them to the list of reasons for the scale up:

```
# failed job because node has been retired --> NF = Node Failure
$ squeue -t all  
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
                 2   compute     wrap   ubuntu NF       2:55      1 ip-10-0-2-231

# job requeued
$ scontrol requeue 2

$ squeue -t PD
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
                 2   compute     wrap   ubuntu PD       0:00      1 (NodeDown)
$ squeue -t PD --noheader -o '%r'
NodeDown
...
$ squeue -t PD
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
                 2   compute     wrap   ubuntu PD       0:00      1 (BeginTime)
$ squeue -t PD --noheader -o '%r'
BeginTime
```

I also added **Priority** because I think that we can scale up if the job is waiting for other higher priority jobs.

This fixes https://github.com/aws/aws-parallelcluster/issues/996

### DIFF
--- a/common/slurm.py
+++ b/common/slurm.py
@@ -14,5 +14,8 @@
 
 PENDING_RESOURCES_REASONS = [
     "Resources",
-    "Nodes required for job are DOWN, DRAINED or reserved for jobs in higher priority partitions"
+    "Nodes required for job are DOWN, DRAINED or reserved for jobs in higher priority partitions",
+    "BeginTime",
+    "NodeDown",
+    "Priority"
 ]


cgorgulla:
If I start a pcluster with slurm on AWS, pending jobs with reason "Priority" still don't start. But I installed pcluster in the ordinary way with pip as described in the documentation, so maybe the changes are not in the latest Python package available for pip? How can I run the latest development version of aws-parallelcluster?

cgorgulla:
Or alternatively, how can I just fix this single issue on a running pcluster (slurm)? I have modified the file  /usr/local/lib/python2.7/site-packages/common/slurm.py on a running cluster according to the above changes from you, and have manually started jobwatcher, and jobwatcher now recognizes all the jobs (including "priority" pending jobs) for adding additional nodes, but somehow they are not really added.

lukeseawalker:
@cgorgulla we just released v2.4.0 with this change

cgorgulla:
@lukeseawalker that is awesome, thank you for letting me know.

PR #128, opened 2019-04-17
Enhancements to improve sqswatcher resilience capacity

demartinofra:
- Failed SQS events are re-queued up to 2 times and then discarded. This is done to avoid processing an event that keeps failing forever.
- raise exception when _restart_compute_daemons fails to enable retries in these cases
 - also reduced the wait_fixed retry time to 2 seconds since there is a timeout on the thread pool that consists in 10 seconds per hosts batch
- add item to dynamodb also on failed ADD events. This allows the REMOVE event to successfully remove partially configured nodes from scheduler

### DIFF
--- a/sqswatcher/plugins/slurm.py
+++ b/sqswatcher/plugins/slurm.py
@@ -74,9 +74,7 @@ def _restart_compute_daemons(hostname, cluster_user):
     # This blocks until command completes
     return_code = stdout.channel.recv_exit_status()
     if return_code != 0:
-        error_message = "Failed when restarting slurmd on compute node {0}".format(hostname)
-        log.error(error_message)
-        raise Exception(error_message)
+        raise Exception("Failed when restarting slurmd on compute node %s", hostname)
     ssh_client.close()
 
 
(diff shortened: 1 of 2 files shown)


sqswatcher/plugins/slurm.py
```diff
@@ -61,7 +61,7 @@ def _restart_master_node():
         raise
 
 
-@retry(stop_max_attempt_number=3, wait_fixed=10000)
+@retry(stop_max_attempt_number=3, wait_fixed=2000)
```
lukeseawalker: is 2000 enough?

sqswatcher/plugins/slurm.py
```diff
@@ -83,7 +85,8 @@ def _restart_compute_node_worker(args):
     try:
         _restart_compute_daemons(*args)
         return hostname, True
-    except Exception:
+    except Exception as e:
+        log.error("Failed when restarting compute node %s with error %s", hostname, e)
```
lukeseawalker: I see this as a duplication

demartinofra: removed the log from _restart_compute_daemons

sqswatcher/plugins/slurm.py
demartinofra: 2000 is the wait time for this fast retry at function level while the longer retry wait time is given when re-queuing the message

PR #130, opened 2019-04-17
sqswatcher/jobwatcher: use correct number of slots for dummy nodes and fetch instance type to support updates

demartinofra:
Instance properties containing the number of vcpus to use whan adding a node to the scheduler was computed only when the jobwatcher is initialized and was using a static instance type value read from the jobwatcher config file. In case of an update of the compute instance type the jobwatcher was still using the old cpu values to compute the number of instances to start when scaling up the cluster. Now the instance properties are re-computed at every iteration of the jobwatcher and the instance type is retrieved from the cloudformation stack parameters. A better solution would consist in updating and reloading the node daemons every time a pcluster update operation is performed.

sqswatcher is now also setting the correct number of slots for the dummy-nodes and these are dynamically updated in case of cluster updates.

Additional minor enhancements:
- remove log parameter from all common.utils function. There is no need to pass the specific logger since the logging output is the same if we initialize the logger in the util module
- remove unneeded max_queue_size from sqs_config. This value is fetched from asg config
- remove unused daemons test config files to prevent these from becoming out of sync

### DIFF
--- a/common/utils.py
+++ b/common/utils.py
@@ -11,6 +11,7 @@
 # or in the "LICENSE.txt" file accompanying this file.
 # This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
 # See the License for the specific language governing permissions and limitations under the License.
+import json
 import logging
 import os
 import shlex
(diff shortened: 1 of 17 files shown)


common/utils.py
```diff
@@ -271,6 +271,12 @@ def _get_vcpus_by_instance_type(instances, instance_type):
...
-def get_compute_instance_type(region, proxy_config, stack_name):
-    parameters = get_cloudformation_stack_parameters(region, proxy_config, stack_name)
-    return parameters["ComputeInstanceType"]
+@retry(stop_max_attempt_number=3, wait_fixed=5000)
+def get_compute_instance_type(region, proxy_config, stack_name, fallback):
+    try:
+        parameters = get_cloudformation_stack_parameters(region, proxy_config, stack_name)
+        return parameters["ComputeInstanceType"]
+    except Exception:
+        if fallback:
```
enrico-usai: I'd add some log errors in this case, to have evidence that we are using a fallback value and we are not able to get the value from the stack.

enrico-usai: I just realized we are already logging in the `get_cloudformation_stack_parameters` function 👍

PR #133, opened 2019-04-18
Minor enhancements and refactoring for nodewatcher

demartinofra:
- Fix: correctly handle maintain_size call from self_terminate. If maintain_size was true the node was not terminating but was entering the infinite termination loop.
- Fix: dump idletime to file at every iteration and not only when this is increased as it was done before. There are cases when the idletime is reset to 0 and hence we need to update the stored copy
- Fix: in case _has_pending_jobs fails the log message was saying not terminating instance but the instance was being terminated. Decided to terminate instance in this case to avoid keeping node up and running with a broken scheduler daemon.
- Refactoring: move check for stack readiness outside of main loop
- Enhancement: reset idletime to 0 when the host becomes essential for the cluster (because of min_size of ASG or because there are pending jobs in scheduler queue)
- Refactoring: check for pending jobs before increasing the idletime to also avoid misleading log messages
- Refactoring: terminate nodewatcher daemon when the node is terminating rather than sleeping forever

### DIFF
--- a/nodewatcher/nodewatcher.py
+++ b/nodewatcher/nodewatcher.py
@@ -17,6 +17,7 @@ import ConfigParser
 import json
 import logging
 import os
+import sys
 import time
 import urllib2
 
@@ -138,9 +139,13 @@ def _self_terminate(asg_name, asg_client, instance_id):
     :param asg_client: ASG boto3 client
     :param instance_id: the instnace to terminate
     """
-    if not _maintain_size(asg_name, asg_client):
-        log.info("Self terminating %s" % instance_id)
-        asg_client.terminate_instance_in_auto_scaling_group(InstanceId=instance_id, ShouldDecrementDesiredCapacity=True)
+    if _maintain_size(asg_name, asg_client):
+        log.info("Not terminating due to min cluster size reached")
+        return False
+
+    log.info("Self terminating %s" % instance_id)
+    asg_client.terminate_instance_in_auto_scaling_group(InstanceId=instance_id, ShouldDecrementDesiredCapacity=True)
+    return True
 
 
 def _maintain_size(asg_name, asg_client):
@@ -163,7 +168,8 @@ def _maintain_size(asg_name, asg_client):
         return True
 
 
-def _is_stack_ready(stack_name, region, proxy_config):
+@retry(wait_fixed=10000, retry_on_result=lambda result: result is False)
+def _wait_for_stack_ready(stack_name, region, proxy_config):
     """
     Verify if the Stack is in one of the *_COMPLETE states.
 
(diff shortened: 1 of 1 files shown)


nodewatcher/nodewatcher.py
```diff
@@ -235,44 +243,38 @@ def _poll_instance_status(config, scheduler_module, asg_name, hostname, instance
     :param hostname: current hostname
     :param instance_id: current instance id
     """
+    log.info("Waiting for stack %s to be ready", config.stack_name)
+    _wait_for_stack_ready(config.stack_name, config.region, config.proxy_config)
```
enrico-usai: You moved the `_wait_for_stack_ready` outside the main loop but it can be a problem during an update because the current behaviour is that the compute nodes (and the nodewatcher daemons) remain running during an update, so the loop should be "paused" waiting for the UPDATE_*COMPLETE status.

We can move this function outside when we'll force the compute nodes to scale-down during an update.

nodewatcher/nodewatcher.py
```diff
@@ -163,7 +168,8 @@ def _maintain_size(asg_name, asg_client):
         return True
 
 
-def _is_stack_ready(stack_name, region, proxy_config):
+@retry(wait_fixed=30000, retry_on_result=lambda result: result is False)
```
enrico-usai: I'd reduce the interval, 30 seconds is too much given that the default scaling_down_period is 2 minutes. I think 15 seconds could be a good compromise.

nodewatcher/nodewatcher.py
```diff
@@ -282,15 +284,17 @@ def _poll_instance_status(config, scheduler_module, asg_name, hostname, instance
                         continue
 
                     try:
-                        _self_terminate(asg_name, asg_conn, instance_id)
-                        termination_in_progress = True
+                        succeeded = _self_terminate(asg_name, asg_conn, instance_id)
+                        if succeeded:
+                            sys.exit(0)
+                        idletime = 0
```
enrico-usai: I think we shouldn't reset the idletime when the `_self_terminate` fails, because we have to retry to terminate the instance in the next round.

nodewatcher/nodewatcher.py
demartinofra: very good point! I'll move this back into the main loop!

nodewatcher/nodewatcher.py
demartinofra: if it fails with exception we do not reset the idletime. This gets reset only if we skip termination due to _maintain_size which means the node became essential to the cluster

nodewatcher/nodewatcher.py
demartinofra: We agreed that during an update a node not performing any task is still considered idle hence the termination of such nodes is aligned with the expected behaviour. Leaving it as it is.

nodewatcher/nodewatcher.py
demartinofra: done!

PR #134, opened 2019-04-26
Do not scale up and maintain size when job has invalid CPU requirements

demartinofra:
This fixes the following issues:
- Cluster won't scale up in case a submitted job requires more slots per node than the max available on the specific instance type. The fix is applied to slurm and torque schedulers. SGE does not have the concept of slots reserved per node.
- [slurm only] cluster won't maintain the current size in case the only pending jobs are those with invalid CPU requirements. This fix still needs to be applied to torque scheduler.

Some minimal tests: https://github.com/aws/aws-parallelcluster/pull/1020 

### DIFF
--- a/nodewatcher/plugins/slurm.py
+++ b/nodewatcher/plugins/slurm.py
@@ -33,15 +33,26 @@ def hasJobs(hostname):
 
 
 def hasPendingJobs():
-    command = "/opt/slurm/bin/squeue -t PD --noheader -o '%r'"
+    command = "/opt/slurm/bin/squeue -t PD --noheader -o '%c-%r'"
 
     # Command outputs the pending jobs in the queue in the following format
-    #  Resources
-    #  Priority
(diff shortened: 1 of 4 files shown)


nodewatcher/plugins/slurm.py
```diff
@@ -75,3 +86,10 @@ def lockHost(hostname, unlock=False):
         run_command(command)
     except subprocess.CalledProcessError:
         log.error("Error %s host %s", "unlocking" if unlock else "locking", hostname)
+
+
+def _get_node_slots():
+    hostname = check_command_output("hostname")
+    command = "/opt/slurm/bin/sinfo -o '%c' -n {0} -h".format(hostname)
```
sean-smith: Can you put a comment with example output?

demartinofra: Sure! Will do in a separate commit as part of the next PR in order to close this one

PR #135, opened 2019-05-01
nodewatcher: self-terminate node in case of failures

demartinofra:
* nodewatcher: handle broader Exception in hasPendingJobs - this allows to better handle all exceptions that might happen when checking for pending jobs. In such cases we assume there are no pending jobs
* jobwatcher - slurm: add down nodes to busy nodes count - This is done to speedup cluster recovery in case of failing nodes. jobwatcher will start requesting an additional node if possible while nodewatcher will take care of terminating the faulty instance.
* nodewatcher: add logic to self terminate failing node - nodewatcher is now able to self-teminate the instance in case the node is reported as down by the scheduler or if not attached to scheduler at all. The logs are dumped to the shared /home/logs/compute directory in order to facilitate future investigations.
  * at nodewatcher startup there is a timeout of 3 minutes before termination in order to allow the scheduler to register the new node. Note that now the nodewatcher is started at the end of user-data so this delay is not affected by the execution of userdata or custom post_install script
  * every minute nodewatcher verifies the status of the node and if this is marked as down by the scheduler or the scheduler is unresponsive it terminates the node. Before terminating the node 1 minute is allowed for recovery.

Note: functions to verify if node is down for sge and torque still have to be implemented

### DIFF
--- a/nodewatcher/nodewatcher.py
+++ b/nodewatcher/nodewatcher.py
@@ -14,18 +14,20 @@
 
 import collections
 import ConfigParser
+import errno
 import json
 import logging
 import os
 import sys
+import tarfile
 import time
 import urllib2
 
 import boto3
 from botocore.config import Config
-from botocore.exceptions import ClientError
-from retrying import retry
+from retrying import RetryError, retry
 
+from common.time_utils import minutes, seconds
 from common.utils import CriticalError, get_asg_name, load_module
 
 log = logging.getLogger(__name__)
@@ -131,21 +133,23 @@ def _lock_host(scheduler_module, hostname, unlock=False):
     time.sleep(15)  # allow for some settling
 
 
-def _self_terminate(asg_name, asg_client, instance_id):
+def _self_terminate(asg_client, instance_id, decrement_desired=True):
     """
     Terminate the given instance and decrease ASG desired capacity.
 
-    :param asg_name: ASG name
     :param asg_client: ASG boto3 client
-    :param instance_id: the instnace to terminate
+    :param instance_id: the instance to terminate
+    :param decrement_desired: if True decrements ASG desired by 1
     """
-    if _maintain_size(asg_name, asg_client):
-        log.info("Not terminating due to min cluster size reached")
+    try:
+        log.info("Self terminating %s" % instance_id)
+        asg_client.terminate_instance_in_auto_scaling_group(
+            InstanceId=instance_id, ShouldDecrementDesiredCapacity=decrement_desired
+        )
+        sys.exit(0)
+    except Exception as e:
+        log.error("Failed when self terminating instance with exception %s.", e)
         return False
-
-    log.info("Self terminating %s" % instance_id)
-    asg_client.terminate_instance_in_auto_scaling_group(InstanceId=instance_id, ShouldDecrementDesiredCapacity=True)
-    return True
 
 
 def _maintain_size(asg_name, asg_client):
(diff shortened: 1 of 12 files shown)


nodewatcher/nodewatcher.py
```diff
@@ -156,19 +160,61 @@ def _maintain_size(asg_name, asg_client):
...
 
-@retry(wait_fixed=10000, retry_on_result=lambda result: result is False)
+def _dump_logs(instance_id):
+    """Dump gzipped /var/log dir to /home/logs/compute/$instance_id.tar.gz"""
+    logs_dir = "/home/logs/compute"
+    filename = "{0}/{1}.tar.gz".format(logs_dir, instance_id)
+    try:
+        try:
+            os.makedirs(logs_dir)
+        except OSError as e:
```
sean-smith: Instead of having a double nested `try/except` block, use two except clauses:

```
try:
...
except e1:
...
except e2:
...
```

nodewatcher/nodewatcher.py
```diff
@@ -244,11 +290,13 @@ def _poll_instance_status(config, scheduler_module, asg_name, hostname, instance
     :param instance_id: current instance id
     """
     _wait_for_stack_ready(config.stack_name, config.region, config.proxy_config)
+    _terminate_if_down(scheduler_module, config, asg_name, instance_id, minutes(3))
```
sean-smith: Can we pull these constants to the top?

Like:
```
TERMINATE_MIN_IF_UNRESPONSIVE = 3
```

nodewatcher/nodewatcher.py
```diff
@@ -244,11 +290,13 @@ def _poll_instance_status(config, scheduler_module, asg_name, hostname, instance
...
 
     idletime = _init_idletime()
     while True:
-        _store_idletime(idletime)
         time.sleep(60)
+        _store_idletime(idletime)
+        _terminate_if_down(scheduler_module, config, asg_name, instance_id, minutes(1))
```
sean-smith: same here... maybe even make this user configurable

nodewatcher/plugins/sge.py
```diff
@@ -71,3 +72,9 @@ def lockHost(hostname, unlock=False):
         run_sge_command(command)
     except subprocess.CalledProcessError:
         log.error("Error %s host %s", "unlocking" if unlock else "locking", hostname)
+
+
+def is_node_down():
+    """Check if node is down according to scheduler"""
+    # ToDo: to be implemented
+    return False
```
sean-smith: I'd leave this with `return None` or pass so we don't accidentally call it

nodewatcher/plugins/torque.py
```diff
@@ -88,3 +89,9 @@ def lockHost(hostname, unlock=False):
         run_command(command)
     except subprocess.CalledProcessError:
         log.error("Error %s host %s", "unlocking" if unlock else "locking", hostname)
+
+
+def is_node_down():
```
sean-smith: same here

nodewatcher/nodewatcher.py
demartinofra: this first try..except is to handle the case where the log dir already exists. It is a bit ugly but unfortunately they added a flag to the makedirs method only in later python versions. I can't merge the except cause I still want to dump the files if the exception is due to existing dir.

nodewatcher/nodewatcher.py
demartinofra: I'll pull them out in constants and later we can make them configurable 👍

nodewatcher/plugins/sge.py
demartinofra: actually returning False is done to disable this logic for sge and torque. nodewatcher will still call is_node_down but the node will never result as down (and this preserves the previous behaviour)

nodewatcher/nodewatcher.py
sean-smith: I still don't see it, how about:

```python
try:
    os.makedirs(logs_dir)
except OSError as e:
if e.errno == errno.EEXIST:
    log.info("Dumping logs to %s", filename)
    with tarfile.open(filename, "w|gz") as archive:
        archive.add("/var/log", recursive=True)
except Exception as e:
    log.warning("Failed while dumping logs to %s with exception %s.", filename, e)
```

Seems like that should do the trick. Am I missing something?

demartinofra: the tarfile needs to be created also when os.makedirs succeeds since makedirs is creating the log dir and not the file

nodewatcher/nodewatcher.py
```diff
@@ -131,22 +133,24 @@ def _lock_host(scheduler_module, hostname, unlock=False):
...
+    :param decrement_desired: if True decrements ASG desired by 1
     """
-    if _maintain_size(asg_name, asg_client):
-        log.info("Not terminating due to min cluster size reached")
+    try:
+        log.info("Self terminating %s" % instance_id)
+        asg_client.terminate_instance_in_auto_scaling_group(
+            InstanceId=instance_id, ShouldDecrementDesiredCapacity=decrement_desired
+        )
+        sys.exit(0)
```
enrico-usai: The `sys.exit(0)` is not really useful since the supervisord will restart the daemon, but ok we can leave it.

demartinofra: supervisord by default does not restart the program when exitcode is 0 and we do not override this config value

source: http://supervisord.org/configuration.html - exitcodes (defaults to 0)

PR #154, opened 2019-06-18
Save remote host key in known hosts file

lukeseawalker:
On compute host joining the cluster, query the compute's remote SSH server to retrieve the public keys to be saved into the shared known hosts file.
This will permit SSH master/compute and compute/compute without prompting for the host key check, without changing the StrictHostKeyChecking setting.

### DIFF
--- a/src/common/ssh_keyscan.py
+++ b/src/common/ssh_keyscan.py
@@ -1,0 +1,112 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License").
+# You may not use this file except in compliance with the License.
+# A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file.
+# This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+# See the License for the specific language governing permissions and limitations under the License.
+import base64
+import logging
+import os
+from math import ceil
+from multiprocessing import Pool
+
+from retrying import retry
+
+from common.utils import run_command
+from paramiko import HostKeys, RSAKey, SSHException, Transport
+
+
+@retry(stop_max_attempt_number=3)
+def _get_server_keys(hostname):
+
+    server_keys = []
+
+    # key_type_list = ["ssh-ed25519", "ssh-rsa", "ecdsa-sha2-nistp256"] # default key_type used by ssh-keysca
+    # Supported key_type for OS
+    # alinux     ssh-rsa,ssh-ed25519,ecdsa-sha2-nistp256
+    # ubuntu1404 ssh-rsa,ssh-ed25519,ecdsa-sha2-nistp256
+    # ubuntu1604 ssh-rsa,ssh-ed25519,ecdsa-sha2-nistp256
+    # centos7    ssh-rsa,ssh-ed25519,ecdsa-sha2-nistp256
+    # centos6    ssh-rsa
+    key_type_list = ["ssh-rsa"]
+
+    for key_type in key_type_list:
+        try:
+            transport = Transport(hostname)
+            transport._preferred_keys = [key_type]
+            transport.start_client()
+            server_keys.append(transport.get_remote_server_key())
+        except Exception as e:
+            pass
+        finally:
+            transport.close()
+
+    if not server_keys:
+        logging.error("Failed retrieving server key from host '{0}'".format(hostname))
+
+    return hostname, [(server_key.get_base64(), server_key.get_name()) for server_key in server_keys]
+
+
+def _get_server_key_on_multiple_hosts(hostnames, parallelism=25, timeout=5):
+    if not hostnames:
+        return {}
+
+    pool = Pool(parallelism)
+    try:
+        r = pool.map_async(_get_server_keys, hostnames)
+        # The pool timeout is computed by adding 2 times the command timeout for each batch of hosts that is
+        # processed in sequence. Where the size of a batch is given by the degree of parallelism.
+        results = r.get(timeout=int(ceil(len(hostnames) / float(parallelism)) * (2 * timeout)))
+    finally:
+        pool.terminate()
+
+    return dict(results)
+
+
+def _add_keys_to_known_hosts(server_keys, host_keys_file):
+    try:
+        if not os.path.isfile(host_keys_file):
+            host_keys = HostKeys()
+        else:
+            host_keys = HostKeys(filename=host_keys_file)
+
+        for hostname, key_list in server_keys.items():
+            try:
+                for key_tuple in key_list:
+                    key = RSAKey(data=base64.b64decode(key_tuple[0]))
+                    host_keys.add(hostname=hostname, key=key, keytype=key_tuple[1])
+                    host_keys.add(hostname=hostname + ".*", key=key, keytype=key_tuple[1])
+                    logging.info(
+                        "Adding keys to known hosts file '{0}' for host '{1}'".format(host_keys_file, hostname)
+                    )
+                    host_keys.save(filename=host_keys_file)
+            except Exception as e:
+                logging.error(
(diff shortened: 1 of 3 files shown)


bwbarrett:
Am I reading the patch correctly that you're storing the keys in the default user's known hosts?  If I was a user, I would find that unexpected for a couple of reasons.  First, if I had set up users (like for a group's cluster), they would have different behaviors than if I used the default.  Second, if I had updated the .ssh/known_hosts for the default user, this would add to it, which might be unexpected.  I think it might be more natural (and way less unexpected) to update the system known hosts file instead.

I think the patch will also cause the file to grow without bound as instances are terminated and restarted.  If a cluster does a lot of scaling, this could be a problem.  It is also likely to result in multiple known_host entries for the same IP, and I'm not sure how ssh will deal with that.  It might be nicer to remove entries on termination?

lukeseawalker:
Hi @bwbarrett,
I was fixing the regression introduced since 2.3.1, for slurm and afterwards for sge, where the host key of the joining compute node is not added anymore in the known hosts file, see [here](https://github.com/aws/aws-parallelcluster-node/blob/develop/common/remote_command_executor.py#L38-L44).

You can compare with the old behaviour which is still in place for torque, [here](https://github.com/aws/aws-parallelcluster-node/blob/develop/sqswatcher/plugins/torque.py#L76-L99), which btw has a flow because first is connecting to the host with the AutoAddPolicy, but only later is loading and saving the known_hosts file, which would override the eventually new connected host with an old removed host with the same IP. 

Also, the host key of the master itself is already added to the known hosts file of the default user, this behaviour didn't change, see [here](https://github.com/aws/aws-parallelcluster-cookbook/blob/develop/recipes/_master_base_config.rb#L175-L181).

I agree with you that using the system known hosts file would favour all the users, not only the default one.

The multiple known_host entries for the same IP is not a case, since any existing entry for a (hostname, keytype) pair will be replaced, this is managed by the [HostKeyEntry object](http://docs.paramiko.org/en/2.5/api/hostkeys.html#paramiko.hostkeys.HostKeys.add).

The only potential issue could be the growing size of the known hosts file, but this should be checked. I do not expects any degraded performance btw. Alas the [HostKeys object](http://docs.paramiko.org/en/2.5/api/hostkeys.html#paramiko.hostkeys.HostKeys) doesn't have a method to remove a given host from the table, and this strengthen my expectation.

lukeseawalker:
I did some investigation for the file size question.
In worst case scenario we can have a subnet with /16 mask, this imply 2^16 unique IPs which translate to 131072 entries in the file (since we have 2 entries per file).
I have generated a known_hosts file with 131072 random hostnames.
```
 > cat known_hosts | wc -l                                                                                                             
  131072
```
Then did some query using `ssh-keygen -F`, with both present and not present hosts
```
> time ssh-keygen -F 89.148.3.6  -f known_hosts
# Host 89.148.3.6 found: line 131071 
89.148.3.6 ssh-rsa AAAAB3NzaC1y [...] =
0.23 real         0.18 user         0.03 sys

> time ssh-keygen -F 72.63.231.17 -f known_hosts   
0.21 real         0.17 user         0.02 sys

> time ssh-keygen -F 72.63.231.16 -f known_hosts     
# Host 72.63.231.16 found: line 43 
72.63.231.16 ssh-rsa AAAAB [...] H3S4ZobsMf7Uug/Lw==
0.20 real         0.17 user         0.02 sys

> time ssh-keygen -F 44.30.159.48 -f known_hosts 
# Host 44.30.159.48 found: line 3238 
44.30.159.48 ssh-rsa AAAAB3NzaC1yc2EAAA [...] Uug/Lw==
0.26 real         0.17 user         0.06 sys

```

Unfortunately `host_keys = HostKeys(filename="known_hosts")` object is taking time to load such big file.
```
time spent opening file with 2000 entries: 2.42742919922 sec
time spent looking up for matching in file: 0.00197601318359sec
---
time spent loading file with 3000 entries: 5.12219500542 sec
time spent looking up for matching in file: 0.00486898422241sec
---
time spent loading file with 4000 entries: 9.15202999115 sec
time spent looking up for matching in file: 0.00587701797485sec
---
time spent loading file with 5000 entries: 18.5455899239 sec
time spent looking up for matching in file: 0.00751996040344sec
---
time spent loading file with 6000 entries: 21.2902128696 sec
time spent looking up for matching in file: 0.0094301700592sec
---
time spent loading file with 7000 entries: 28.8285210133 sec
time spent looking up for matching in file: 0.00925183296204sec
---
time spent loading file with 8000 entries: 37.7386338711 sec
time spent looking up for matching in file: 0.0102441310883sec
---
time spent loading file with 9000 entries: 47.4204499722 sec
time spent looking up for matching in file: 0.0112991333008sec
---
time spent loading file with 10000 entries: 58.346050024 sec
time spent looking up for matching in file: 0.012433052063sec
---
time spent opening file with 11000 entries: 71.2257401943 sec
time spent looking up for matching in file: 0.0137209892273sec
---
time spent opening file with 12000 entries: 85.3885440826 sec
time spent looking up for matching in file: 0.0165331363678sec
---
time spent opening file with 13000 entries: 102.463492155 sec
time spent looking up for matching in file: 0.0154161453247sec
```

src/common/ssh_keyscan.py
```diff
@@ -0,0 +1,111 @@
...
+    for hostname in hostnames:
+        command = "ssh-keygen -R " + hostname + " -f " + host_keys_file
+        run_command(command, raise_on_error=False, execute_as_user=user)
+        command = "ssh-keygen -R " + hostname + ". -f " + host_keys_file
+        run_command(command, raise_on_error=False, execute_as_user=user)
+
+
+def ssh_keyscan_host(events, user):
+    host_keys_file = os.path.expanduser("~" + user) + "/.ssh/known_hosts"
+    __remove_keys_to_known_hosts(
```
demartinofra: remove_keys_from...

src/common/ssh_keyscan.py
```diff
@@ -0,0 +1,111 @@
...
+                )
+    except Exception as e:
+        logging.error("Failed adding keys to known hosts file '{0}', with exception: {1}".format(host_keys_file, e))
+
+
+def __remove_keys_to_known_hosts(hostnames, host_keys_file, user):
+    for hostname in hostnames:
+        command = "ssh-keygen -R " + hostname + " -f " + host_keys_file
+        run_command(command, raise_on_error=False, execute_as_user=user)
+        command = "ssh-keygen -R " + hostname + ". -f " + host_keys_file
```
demartinofra: check if there is a missing * after the .

src/common/ssh_keyscan.py
```diff
@@ -0,0 +1,111 @@
...
+    # centos6    ssh-rsa
+    key_type_list = ["ssh-rsa"]
+
+    for key_type in key_type_list:
+        try:
+            transport = Transport(hostname)
+            transport._preferred_keys = [key_type]
+            transport.start_client()
+            server_keys.append(transport.get_remote_server_key())
+            transport.close()
```
demartinofra: optional: use finally or check if with works

src/common/ssh_keyscan.py
```diff
@@ -0,0 +1,111 @@
...
+        except Exception as e:
+            pass
+
+    if not server_keys:
+        logging.error("Failed retrieving server key from host '{0}'".format(hostname))
+
+    return hostname, [(server_key.get_base64(), server_key.get_name()) for server_key in server_keys if server_keys]
```
demartinofra: if server_keys can be dropped

common/ssh_keyscan.py
```diff
@@ -0,0 +1,111 @@
...
+            except Exception as e:
+                logging.error(
+                    "Failed adding keys to known hosts file for host '{0}', with exception: {1}".format(hostname, e)
+                )
+    except Exception as e:
+        logging.error("Failed adding keys to known hosts file '{0}', with exception: {1}".format(host_keys_file, e))
+
+
+def __remove_keys_to_known_hosts(hostnames, host_keys_file, user):
+    for hostname in hostnames:
```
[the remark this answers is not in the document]
lukeseawalker: unfortunately not with ssh-keygen

src/common/ssh_keyscan.py
lukeseawalker: checked

PR #157, opened 2019-07-09
Torque scheduler enhancements

demartinofra:
**sqswatcher - torque: refactoring + performance enhancements**
This commit rewrites the code for the sqswatcher torque plugin in order to achieve the following enhancements:

- Handle nodes add/removal in parallel.
- Trigger scheduling cycle only when the first node gets added to cluster
- Add logic to handle retries on failures

Comments:

- In the previous version the command `/opt/torque/bin/qmgr -c "set server scheduling=true"` which triggers a scheduling cycle was executed every time a new host was added and its state was set to free. With the new logic the command is executed only when a batch of hosts is added to a cluster with no active nodes and moreover we only wait for one of the nodes to become free. The scheduling cycle needs to be enforced only when the cluster transitions out from 0 active nodes. If this is not done the jobs will be kept in pending state.
-  Command `"/opt/torque/bin/pbsnodes -c %s" % hostname` which was clearing offline state from the node has been removed since it was always failing and not bringing any added value. New nodes automatically transition from down,offline,MOM-list-not-sent -> down -> free
- Torque commands have been moved to a common module and output is now parsed from a xml format and not from a tabular format like before. This is done to have a more consistent and robust parsing behaviour and  to avoid having partial data.
- When nodes are added or removed with a qmgr command we need to handle cases when the node was already added to the cluster (Node name already exists) or already removed (Unknown node). `_qmgr_manage_nodes` takes care of parsing qmgr output to understand when a failure can be ignored.
- Code to add SSH host identities to known_hosts file has been removed as done for the other schedulers. This regression will be fixed for all schedulers in #154
- Failures during node add/removal are now retried up to 3 times. The operations to add/remove a node should be idempotent.

------

**sqswatcher - torque: set cluster limits in scheduler config**

The following limits are set in the scheduler config and dynamically updated when required:

- max number of nodes in the cluster: this limit is set at the [queue](http://docs.adaptivecomputing.com/torque/6-1-2/adminGuide/torque.htm#topics/torque/13-appendices/nQueueAttributes/queueAttribReference.htm#resources_available) and [server](http://docs.adaptivecomputing.com/torque/6-1-2/adminGuide/torque.htm#topics/torque/13-appendices/bServerParameters/serverParameters.htm#resources_available) level by setting resources_available.nodect. This blocks jobs submitted with the `-l nodes=X` option that violate max cluster size. 
- max number of slots per node: this sets the number of max slots per node by using [resources_max.ncpus](http://docs.adaptivecomputing.com/torque/6-1-2/adminGuide/torque.htm#topics/torque/13-appendices/nQueueAttributes/queueAttribReference.htm#resources_max) at the queue level. This blocks jobs submitted with the `-l ncpus=X` option that violate max slots per instance. Note: This does not block jobs submitted with `-l nodes=X:ppn=Y` where Y violates slots contraints.
- number of slots assigned to fake master server: the number of np associated to the fake master server that were previously statically set to 1000 is now dinamically computed by taking into account the max cluster size, the max slot per instance and the active nodes attached to the scheduler. This value controls the total ppn allowed for a job (e.g. a job submitted with `-l nodes=2:ppn=5+3:ppn=7` has a total ppn of 31 and will be submitted only if the total number of np slots in the cluster is bigger than the computed value). Note: ppn cannot be lower than 1. 

Useful link: http://docs.adaptivecomputing.com/torque/6-1-2/adminGuide/torque.htm#topics/torque/2-jobs/requestingRes.htm#interpretRequests

------

**nodewatcher - torque: implement support for is_node_down**

Terminate a node that is in "down", "offline" or "unknown" state for more than the configured timeout (3 minutes on node initialization and 1 minute when it fails after it was added to scheduler)

Valid state strings are "active", "all", "busy", "down", "free", "job-exclusive", "job-sharing", "offline", "reserve", "state-unknown", "time-shared", and "up": http://docs.adaptivecomputing.com/torque/6-1-2/adminGuide/torque.htm#topics/torque/commands/pbsnodes.htm

------

**jobwatcher - torque: rewrite get_busy_nodes to use torque_commands**

Align get_busy_nodes with the new torque commands module. No functional changes are expected.

------

**Split node states when parsing Torque hosts**

Simply store torque host states as an array of states rather than a comma separated string. No functional changes are expected.

------

**jobwatcher - torque: count as busy the nodes in offline or state-unknown state**

Marking nodes that are in `offline` or `state-unknown` state as busy so that the jobwatcher starts requesting a replacement for them. Usually offline state is set by nodewatcher to lock the hosts when tearing them down. This improved logic allows a faster reaction time if new nodes are needed to replace the terminated ones.

Due to the fact that when a node is added it transitions from `down,offline,MOM-list-not-sent -> down -> free` then we need to filter out nodes in MOM-list-not-sent.

------

**torque: add functions to retrieve jobs info**

Add functions needed to retrieve the status of jobs in the system. The funcitons are parsing the xml output of the qstat command and also offer some filtering option to implement the ppn limits support (see next commit).

------

**jobwatcher - torque: add support for ncpus and multi nodes resource allocation**

Add support for the following submission options:

- `-l ncpus=x`: the scaling logic is now able to properly behave in case a job is submitted with such option. As reported in the [official docs](http://docs.adaptivecomputing.com/torque/6-1-2/adminGuide/torque.htm#topics/torque/2-jobs/requestingRes.htm#interpretRequests): ncpus=X Creates 1 task that will use *X* lprocs. This means that jobs submitted with this option will require a single node.
- `-l nodes=x1:ppn=y1+nodes=x2:ppn=y2`: jobwatcher is now discarding jobs that require a ppn that is not satisfiable given the cluster instance limits. `get_required_nodes` is treating jobs with heterogeneous requirements as independent jobs when calling the get_optimal_nodes function. For example a job with `-l nodes=3:ppn=2+nodes=2:ppn=1` will produce the following invocation `get_optimal_nodes([3,2], [6,2], instance_properties)`

------

**torque: add filtering options to get_jobs_info command**

`get_jobs_info` can now filter jobs by state and by execution hosts. The filtering is done by us since qstat command does not provide such a feature.

------

**nodewatcher - torque: add support for suspended jobs and max slots per instance**

- node is marked as free if pending jobs have invalid requirements
- node is marked as busy if a suspended job is assigned to it

------

**Filter out MasterServer from compute nodes list**

Remove fake MasterServer from the list of compute nodes retrieved with pbsnodes command.

------


src/common/schedulers/torque_commands.py
```diff
@@ -32,7 +32,9 @@
     "time-shared",
     "state-unknown",
 )
-TORQUE_PENDING_JOBS_STATE = "Q"
+TORQUE_PENDING_JOB_STATE = "Q"
+TORQUE_SUSPENDED_JOB_STATE = "S"
+TORQUE_RUNNING_JOB_STATE = "R"
```
enrico-usai: In future we could add an enum or a mapping with all the states

src/nodewatcher/plugins/torque.py
```diff
@@ -12,74 +12,42 @@
...
-            if idx < 2:
-                continue
-            queue_status = line.split()
-            pending += int(queue_status[5])
-
-        has_pending = pending > 0
-        error = False
+        pending_jobs = get_pending_jobs_info(max_slots_filter=instance_properties.get("slots"))
+        logging.info("Found the following pending jobs:\n%s", pending_jobs)
+        return len(pending_jobs) > 0, False
```
enrico-usai: It is not clear what is the second returned parameter.
I know that it is "error" but we should clarify it in the doc

src/common/schedulers/torque_commands.py
```diff
@@ -0,0 +1,159 @@
...
+    "state-unknown",
+)
+
+
+def _qmgr_manage_nodes(operation, hosts, error_messages_to_ignore, additional_qmgr_args=""):
+    if not hosts:
+        return []
+
+    hostnames = ",".join(hosts)
+    command = '/opt/torque/bin/qmgr -c "{operation} node {hostnames} {additional_args}"'.format(
```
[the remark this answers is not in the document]
demartinofra: makes total sense. Done in https://github.com/aws/aws-parallelcluster-node/pull/157/commits/5f9f8628ddd449aad9f6317026984e0204d43d8f

src/common/schedulers/torque_commands.py
```diff
@@ -0,0 +1,159 @@
...
+    "job-sharing",
+    "busy",
+    "time-shared",
+    "state-unknown",
+)
+
+
+def _qmgr_manage_nodes(operation, hosts, error_messages_to_ignore, additional_qmgr_args=""):
+    if not hosts:
+        return []
```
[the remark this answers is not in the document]
demartinofra: 👍 https://github.com/aws/aws-parallelcluster-node/pull/157/commits/fc80965519bfda223c129b7eb0d4e9ffb7a77bf8

src/nodewatcher/plugins/torque.py
demartinofra: https://github.com/aws/aws-parallelcluster-node/pull/157/commits/981e1f58edab13c03d67b32821fcd1cc6fc8056f

PR #159, opened 2019-07-10
Upgrade to Python 3

sean-smith:
* Remove support for Python 2.
* Add Python 3 support

Signed-off-by: Sean Smith <seaam@amazon.com>


setup.py
```diff
@@ -30,18 +30,15 @@ def read(fname):
...
-    requires.append("cryptography==2.1.4")  # dependency of paramiko
-    requires.append("pycparser==2.18")
-else:
-    requires.append("paramiko>=2.4.2")
+requires = [
+    "requests>=2.21.0",
+    "boto3>=1.7.55",
+    "python-dateutil>=2.6.1",
+    "retrying>=1.3.3",
+    "future>=0.17.1",
```
demartinofra: we can probably drop this

setup.py
```diff
@@ -30,18 +30,15 @@ def read(fname):
...
-    # For python2.6 we have to require argparse since it
-    # was not in stdlib until 2.7.
-    requires.append("argparse>=1.4")
-    requires.append("idna==2.6")
-    requires.append("paramiko==2.3.3")
-    requires.append("cryptography==2.1.4")  # dependency of paramiko
-    requires.append("pycparser==2.18")
-else:
-    requires.append("paramiko>=2.4.2")
+requires = [
```
demartinofra: shall we sync the requirements.txt file with this list?

setup.py
```diff
@@ -30,18 +30,15 @@ def read(fname):
...
+    "requests>=2.21.0",
+    "boto3>=1.7.55",
+    "python-dateutil>=2.6.1",
+    "retrying>=1.3.3",
+    "future>=0.17.1",
+    "configparser>=3.7.4",
+    "paramiko>=2.4.2",
+]
 
 setup(
```
demartinofra: shall we add `python_requires='>=3.4'`?

https://packaging.python.org/guides/dropping-older-python-versions/

setup.py
sean-smith: done

setup.py
sean-smith: done

setup.py
sean-smith: added

setup.py
sean-smith: removed

sean-smith:
> Also I think we can remove all the future usages (e.g. future.moves...)

done

PR #162, opened 2019-07-15
Convert to Python 3

sean-smith:


src/sqswatcher/plugins/slurm.py
```diff
@@ -72,20 +72,19 @@ def _read_node_list():
...
-            fh,
-            "NodeName=dummy-compute[1-{0}] CPUs={1} State=FUTURE\n".format(
-                dummy_nodes_count, instance_properties["slots"]
-            ),
-        )
-    for node in node_list:
-        os.write(fh, "{0}".format(node))
-
-    os.close(fh)
+    with tempfile.NamedTemporaryFile(mode="r+", delete=False) as file:
```
demartinofra: shall we use `w` as mode?

sean-smith: updated

PR #163, opened 2019-07-16
Stop polling for status in the event of failure

jason-mccloskey:
*Issue #, if available:*
I have a customer leveraging 235 compute nodes. Having this many failures, we often see node failures from a lustre mount issue. This results in a cfn status of "CREATE_FAILED". When that happens, that 235 nodes query describe_stacks every 10 seconds for 10 minutes. This swamps the DescribeStacks API and results in throttling downstream.

*Description of changes:*
Add "CREATE_FAILED" and "UPDATE_FAILED" as potential statuses where the stack is ready. These are final statuses when leveraging the "-nr" switch.

### DIFF
--- a/src/nodewatcher/nodewatcher.py
+++ b/src/nodewatcher/nodewatcher.py
@@ -223,7 +223,13 @@ def _wait_for_stack_ready(stack_name, region, proxy_config):
     stacks = cfn_client.describe_stacks(StackName=stack_name)
     stack_status = stacks["Stacks"][0]["StackStatus"]
     log.info("Stack %s is in status: %s", stack_name, stack_status)
-    return stack_status in ["CREATE_COMPLETE", "UPDATE_COMPLETE", "UPDATE_ROLLBACK_COMPLETE"]
+    return stack_status in [
+        "CREATE_COMPLETE",
+        "UPDATE_COMPLETE",
+        "UPDATE_ROLLBACK_COMPLETE",
+        "CREATE_FAILED",
+        "UPDATE_FAILED",
+    ]
 
 
 def _init_data_dir():


src/nodewatcher/nodewatcher.py
```diff
@@ -223,7 +223,8 @@ def _wait_for_stack_ready(stack_name, region, proxy_config):
     stacks = cfn_client.describe_stacks(StackName=stack_name)
     stack_status = stacks["Stacks"][0]["StackStatus"]
     log.info("Stack %s is in status: %s", stack_name, stack_status)
-    return stack_status in ["CREATE_COMPLETE", "UPDATE_COMPLETE", "UPDATE_ROLLBACK_COMPLETE"]
+    return stack_status in ["CREATE_COMPLETE", "UPDATE_COMPLETE", "UPDATE_ROLLBACK_COMPLETE", "CREATE_FAILED",
+                            "UPDATE_FAILED"]
```
sean-smith: can you remove the whitespace before `UPDATE_FAILED`

sean-smith: To do so you can run `tox -e autoformat` like so:

```bash
pip install tox
cd cli
tox -e autoformat
```
Commit and push the changes! That'll make travis happy.

jason-mccloskey: Autoformating ran and branch updated.

sean-smith:
Ok, I've tested with the following config, with and without `vpc_security_group_id` 

```ini
[cluster cfn-test]
key_name =[key]
vpc_settings = public
template_url = [custom_template]
fsx_settings = fs

[fsx fs]
shared_dir = /fsx
storage_capacity = 3600

[vpc public]
vpc_id = [vpc]
master_subnet_id = [subnet]
vpc_security_group_id = [sg with 988 open]
```

both mount `/fsx`

```bash
$ df -h
Filesystem              Size  Used Avail Use% Mounted on
devtmpfs                483M   68K  483M   1% /dev
tmpfs                   493M     0  493M   0% /dev/shm
/dev/xvda1               17G  7.5G  9.2G  45% /
/dev/xvdb                20G   45M   19G   1% /shared
172.31.94.181@tcp:/fsx  3.4T   14M  3.4T   1% /fsx
```

PR #173, opened 2019-09-04
Increase max retrieved SQS messages and improved daemons loop execution rate

demartinofra:
- Subtract daemons loop execution time from overall sleep time: this allows to run daemons iterations at a constant frequency that does not depend on the loop execution time

- Increase max retrieved SQS messages from 50 to 200. Also make this number configurable with a sqswatcher config directive: max_processed_messages.

Tested manually. Need to run integration tests.

### DIFF
--- a/src/nodewatcher/nodewatcher.py
+++ b/src/nodewatcher/nodewatcher.py
@@ -19,6 +19,7 @@ import sys
 import tarfile
 import time
 from contextlib import closing
+from datetime import datetime
 
 import boto3
 from botocore.config import Config
@@ -37,7 +38,7 @@ IDLETIME_FILE = DATA_DIR + "node_idletime.json"
 INITIAL_TERMINATE_TIMEOUT = minutes(3)
 # Timeout used at every iteration of nodewatcher loop.
 TERMINATE_TIMEOUT = minutes(1)
-
+LOOP_TIME = 60
 
 NodewatcherConfig = collections.namedtuple(
     "NodewatcherConfig", ["region", "scheduler", "stack_name", "scaledown_idletime", "proxy_config"]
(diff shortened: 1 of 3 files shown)


src/nodewatcher/nodewatcher.py
```diff
@@ -278,6 +284,22 @@ def _init_idletime():
     return idletime
 
 
+def _lock_and_terminate(scheduler_module, hostname, asg_name, asg_conn, instance_id):
+    _lock_host(scheduler_module, hostname)
+    has_jobs = _has_jobs(scheduler_module, hostname)
```
enrico-usai: minor: we could avoid the `has_jobs` variable since it is used in a single place and the name of the function called is autoexplicative.

src/nodewatcher/nodewatcher.py
```diff
@@ -295,8 +317,16 @@ def _poll_instance_status(config, scheduler_module, asg_name, hostname, instance
 
     idletime = _init_idletime()
     instance_properties = get_instance_properties(config.region, config.proxy_config, instance_type)
+    start_time = None
     while True:
-        time.sleep(60)
+        end_time = datetime.now()
+        if not start_time:
+            start_time = end_time
+        time_delta = (end_time - start_time).total_seconds()
```
enrico-usai: What about creating a common function to use in the three daemons?

Something like:
```python
def wait_something(loop_time=60, start_time=None):
    time_delta = (datetime.now() - start_time).total_seconds() if start_time else 0
    if time_delta < LOOP_TIME:
        time.sleep(LOOP_TIME - time_delta)
```

src/nodewatcher/nodewatcher.py
```diff
@@ -325,24 +355,12 @@ def _poll_instance_status(config, scheduler_module, asg_name, hostname, instance
...
-
-                    if _maintain_size(asg_name, asg_conn):
-                        log.info("Not terminating due to min cluster size reached")
-                        idletime = 0
-                    else:
-                        _self_terminate(asg_conn, instance_id)
-
-                    _lock_host(scheduler_module, hostname, unlock=True)
+                    _lock_and_terminate(scheduler_module, hostname, asg_name, asg_conn, instance_id)
+                    # set idletime to 0 if termination is aborted
```
enrico-usai: Actually we are setting idletime to 0 also when the _self_termination is executed

demartinofra: when self termination is executed successfully the daemon process exits with success

src/nodewatcher/nodewatcher.py
demartinofra: Good idea! done!

PR #174, opened 2019-10-04
Manage possible exceptions from batch hosts addition or removal

lukeseawalker:
Prior the patch the system was unable to recover if one or more host couldn't be added to the scheduler configuration.
The symptoms were that the SQS messages were not re-queued with a decreasing TTL, and the total number of messages grew indefinitely, causing the impossibility of adding new nodes.
This fix https://github.com/aws/aws-parallelcluster/issues/1327

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/sqswatcher/plugins/slurm.py
+++ b/src/sqswatcher/plugins/slurm.py
@@ -109,24 +109,20 @@ def update_cluster(max_cluster_size, cluster_user, update_events, instance_prope
             # Restarting also if already in config cause it might have failed at the previous iteration
             nodes_to_restart.append(event.host.hostname)
 
-    try:
-        _write_node_list(node_list, max_cluster_size, instance_properties)
-        _restart_master_node()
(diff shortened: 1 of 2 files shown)


src/sqswatcher/plugins/sge.py
```diff
@@ -84,18 +84,22 @@ def update_cluster(max_cluster_size, cluster_user, update_events, instance_prope
...
+        succeeded = []
+        failed = []
+        for event in update_events:
+            if event.host.hostname in added_hosts or event.host.hostname in removed_hosts:
+                succeeded.append(event)
+            else:
+                failed.append(event)
+
+        return failed, succeeded
+    except Exception as e:
```
demartinofra: what do you think if we move this to the sqswatcher.py where we call update_cluster so that we have a uniform behaviour for all schedulers without duplication? Otherwise it LGTM

PR #177, opened 2019-10-18
Modification for slurm GPU option support

rexcsn:
* Modified utils.py to retrieve GPU information from instances file
* Modified sqswatcher slurm.py to add GPU information to gres.conf and slurm_parallelcluster_nodes.conf when applicable. Logic to update gres.conf is the same as updating 
slurm_parallelcluster_nodes.conf.
* Added unit tests for sqswatcher to check node update functions


src/sqswatcher/plugins/slurm.py
```diff
@@ -100,16 +171,32 @@ def update_cluster(max_cluster_size, cluster_user, update_events, instance_prope
...
             node_list = [node for node in node_list if node.split()[0] != node_name]
         elif event.action == "ADD":
-            # Add new node
-            new_node = "NodeName={nodename} CPUs={cpus} State=UNKNOWN\n".format(
-                nodename=event.host.hostname, cpus=event.host.slots
-            )
+            # Add new node and gres_node depending on whether the instance has GPU
+            if "gpus" in instance_properties:
+                new_node = "NodeName={nodename} CPUs={cpus} Gres=gpu:tesla:{gpus} State=UNKNOWN\n".format(
+                    nodename=event.host.hostname, cpus=event.host.slots, gpus=instance_properties["gpus"]
```
demartinofra: I'm wondering if we could send the gpu info in the event as we are doing for slots for the following reasons:
- in case of an update of the instance type we are sure we are using the correct value even if the instance properties hasn't updated yet. dummy-nodes are recomputed on updates of instance_properties but real nodes aren't
- it is slightly easier to move to a multi-instance type model

src/sqswatcher/plugins/slurm.py
```diff
@@ -100,16 +171,32 @@ def update_cluster(max_cluster_size, cluster_user, update_events, instance_prope
...
 
+    return node_list, nodes_to_restart
+
+
+def update_cluster(max_cluster_size, cluster_user, update_events, instance_properties):
+    """Update and write node lists( and gres_nodes if instance has GPU); restart relevant nodes."""
+    node_list, nodes_to_restart = update_node_lists(update_events, instance_properties)
     _write_node_list(node_list, max_cluster_size, instance_properties)
+    # Only modify gres.conf if instance has GPU
+    if "gpus" in instance_properties:
```
demartinofra: what if customer updates the cluster by changing instance type and goes from a gpu enabled instance to an instance without gpus? We should support this and clean up the gres.conf as nodes are deallocated. Always writing the gres.conf file might be a potential solution

tests/common/schedulers/test_sge_commands.py
```diff
@@ -20,9 +22,13 @@
     get_jobs_info,
     get_pending_jobs_info,
 )
-from sqswatcher.sqswatcher import Host
 from tests.common import read_text
 
+# Changed Host class definition in sqswatcher to add gpu info for host.
+# Old Host class definition is included in new definition.
+# Using the old Host class definition in this test should not affect test validity.
+Host = collections.namedtuple("Host", ["instance_id", "hostname", "slots"])
```
demartinofra: I'd update the Host initialization below and keep using the Host from sqswatcher rather than a mocked version.

src/sqswatcher/sqswatcher.py
```diff
@@ -263,11 +263,16 @@ def _parse_sqs_messages(messages, table):
     return update_events.values()
 
 
-def _process_compute_ready_event(message_attrs, message):
+def _process_compute_ready_event(sqs_config, message_attrs, message):
     instance_id = message_attrs.get("EC2InstanceId")
+    instance_type = message_attrs.get("EC2InstanceType")
+    # Retrieving instances properties for each event because instance types
+    # from instance and CloudFormation could be out-of-sync
+    instance_properties = get_instance_properties(sqs_config.region, sqs_config.proxy_config, instance_type)
```
demartinofra: we really don't want to retrieve this file for every new node especially since the update of the compute instance type is a very rare event. We could implement some very simple caching mechanism in get_instance_properties:

```
def get_instance_properties(...):
     if not hasattr(get_instance_properties, "cache"):
          cache = {}
     if instance_type in get_instance_properties.cache:
          return cache[instance_type]

     same as now + update cache before return
```

Cache could also be simply the last item.

src/sqswatcher/sqswatcher.py
```diff
@@ -263,11 +263,16 @@ def _parse_sqs_messages(messages, table):
     return update_events.values()
 
 
-def _process_compute_ready_event(message_attrs, message):
+def _process_compute_ready_event(sqs_config, message_attrs, message):
```
demartinofra: let's only pass the required parameters such as region and proxy_config.

As a general comment we should really get rid of region and proxy_config and set them as env variable right after we read the config. No need to do it now since unrelated.

src/sqswatcher/plugins/slurm.py
```diff
@@ -61,55 +64,120 @@ def _reconfigure_nodes():
...
     for event in update_events:
         if event.action == "REMOVE":
             node_name = "NodeName={0}".format(event.host.hostname)
             node_list = [node for node in node_list if node.split()[0] != node_name]
         elif event.action == "ADD":
-            # Add new node
-            new_node = "NodeName={nodename} CPUs={cpus} State=UNKNOWN\n".format(
-                nodename=event.host.hostname, cpus=event.host.slots
+            # Only include GPU info if instance has GPU
+            gpu_info = "Gres=gpu:tesla:{gpus} ".format(gpus=event.host.gpus) if int(event.host.gpus) != 0 else ""
```
demartinofra: can we have gpus stored already as an int in host?

src/sqswatcher/plugins/slurm.py
```diff
@@ -61,55 +64,120 @@ def _reconfigure_nodes():
...
+
+def _add_dummy_to_node_list(node_list, max_cluster_size, instance_properties):
+    dummy_nodes_count = max_cluster_size - len(node_list)
+    if dummy_nodes_count > 0:
+        # Include GPU information for dummy nodes if instance type has GPU
+        if int(instance_properties["gpus"]) == 0:
+            gpu_info = ""
+        else:
+            gpu_info = "Gres=gpu:tesla:{0} ".format(instance_properties["gpus"])
+        node_list.insert(
```
demartinofra: not a big issue in this case (given ths size) but list is not an optimized data structure to add items to the front. A deque might be a more suitable alternative but you need to double check that.

rexcsn: Yes insert to front of list is O(n), and insert to front of deque is O(1) in python. In our case the order of the nodes doesn't matter, proposing to append dummy nodes to the end of our node list, which is O(1).

src/sqswatcher/plugins/slurm.py
rexcsn: setting gpus as int in `get_instance_properties`

src/sqswatcher/plugins/slurm.py
rexcsn: Changing back to `insert(0)` to be consistent with integration test

src/common/utils.py
```diff
@@ -298,7 +321,7 @@ def _get_vcpus_by_instance_type(instances, instance_type):
     """
     try:
         vcpus = int(instances[instance_type]["vcpus"])
-        log.info("Instance %s has %s vcpus." % (instance_type, vcpus))
+        log.info("Instance {0} has {1} vcpus.".format(instance_type, vcpus))
```
demartinofra: minor: this log is probably redundant

src/common/utils.py
```diff
@@ -212,58 +212,80 @@ def _read_cfnconfig():
...
-    else:
-        log.error("cfn_scheduler_slots config parameter '%s' is invalid. Assuming 'vcpus'" % cfn_scheduler_slots)
-        slots = vcpus
 
-    log.info("Number of slots computed for instance %s: %d", instance_type, slots)
-    return {"slots": slots}
+        log.info("Added instance type: {0} to get_instance_properties cache".format(instance_type))
+        get_instance_properties.cache[instance_type] = {"slots": slots, "gpus": int(gpus)}
+
+    log.info(
```
demartinofra: these logs can be replaced by a single one printing the entire dictionary

src/common/utils.py
```diff
@@ -312,6 +335,25 @@ def _get_vcpus_by_instance_type(instances, instance_type):
...
+    Get gpus for the given instance type from the pricing file.
+
+    :param instances: dictionary conatining the content of the instances file
+    :param instance_type: The instance type to search for
+    :return: the number of GPU for the given instance type
+    :raise CriticalError if unable to find the given instance or whatever error.
+    """
+    try:
+        gpus = int(instances[instance_type]["gpu"])
+        log.info("Instance {0} has {1} gpus.".format(instance_type, gpus))
```
demartinofra: same here

src/sqswatcher/plugins/slurm.py
```diff
@@ -61,55 +64,120 @@ def _reconfigure_nodes():
...
+                dummy_nodes_count, instance_properties["gpus"] - 1
+            ),
+        )
+
+
+def update_cluster(max_cluster_size, cluster_user, update_events, instance_properties):
+    """Update and write node lists( and gres_nodes if instance has GPU); restart relevant nodes."""
+    node_list, nodes_to_restart = _update_node_lists(update_events)
+    _add_dummy_to_node_list(node_list, max_cluster_size, instance_properties)
+    _write_node_list_to_file(node_list, PCLUSTER_NODES_CONFIG)
```
demartinofra: it might be useful to have a log message to log that we are updating the file and print the list of nodes we are writing to the file.

Same for gres.

rexcsn: added logging to `_write_node_list_to_file`

PR #180, opened 2019-10-24
Add dummy node when cluster stops

rexcsn:
* Add a dummy node to slurm_parallelcluster_nodes.conf when the cluster is stopped(node_list empty) so that slurm daemon can still start and receive jobs


src/sqswatcher/plugins/slurm.py
```diff
@@ -172,6 +172,9 @@ def update_cluster(max_cluster_size, cluster_user, update_events, instance_prope
     """Update and write node lists( and gres_nodes if instance has GPU); restart relevant nodes."""
     node_list, nodes_to_restart = _update_node_lists(update_events)
     _add_dummy_to_node_list(node_list, max_cluster_size, instance_properties)
+    # Add a dummy node if node_list is empty so that slurm can start when cluster is stopped
+    if not node_list:
+        node_list.append("NodeName=dummy-compute-stop CPUs={1} State=DOWN\n".format(instance_properties["slots"]))
```
demartinofra: let's also add a reason such as `Reason=Cluster is stopped or max size is 0.`

PR #181, opened 2019-10-28
Added memory info to slurm nodes

rexcsn:
* Modified get_instance_properties in utils.py to retrieve memory from instances.json
* Modified sqswatcher.py to include memoryinfo in "Add" events
* Modified slurm plugin to write memory info to nodes


src/common/utils.py
```diff
@@ -301,18 +302,26 @@ def _fetch_pricing_file(region, proxy_config):
...
-    :return: the number of vcpus for the given instance type
+    :return: the number of vcpus for the given instance type, memory(in MiB) for the instance type
     :raise CriticalError if unable to find the given instance or whatever error.
     """
     try:
         vcpus = int(instances[instance_type]["vcpus"])
-        return vcpus
+        log.info("Instance {0} has {1} vcpus.".format(instance_type, vcpus))
+        # Rounding down on memory in each conversion step to avoid "low real_memory" error in slurm
+        memory = instances[instance_type]["memory"]
```
demartinofra: what about printing an error and skipping memory config if not present in the file?

src/common/utils.py
```diff
@@ -225,7 +225,8 @@ def _get_instance_info_from_pricing_file(region, proxy_config, instance_type):
         log.critical(error_msg)
         raise CriticalError(error_msg)
 
-    return _get_vcpus_by_instance_type(instances, instance_type), _get_gpus_by_instance_type(instances, instance_type)
+    vcpus, memory = _get_vcpus_mem_by_instance_type(instances, instance_type)
```
demartinofra: minor: I'd put some order here since it's a bit confusing having a function for vcpus and memory and one for gpu.

- the gpu one can actually be replaced by this 1 liner: `int(instances[instance_type].get("gpu", 0))`
- the vcpu can be also a 1 liner: `int(instances[instance_type]["vcpus"])`
- the memory can actually also be a 1 liner: `int(float(instances[instance_type].get("memory", 0))*1000)`. But then we need to handle memory==0 in the caller

I'd have a single try except around the 3 operations above where we fail with critical and a clear error message (as already done) in case we cannot find the instance in the json doc.

src/common/utils.py
```diff
@@ -301,18 +302,26 @@ def _fetch_pricing_file(region, proxy_config):
...
     :raise CriticalError if unable to find the given instance or whatever error.
     """
     try:
         vcpus = int(instances[instance_type]["vcpus"])
-        return vcpus
+        log.info("Instance {0} has {1} vcpus.".format(instance_type, vcpus))
+        # Rounding down on memory in each conversion step to avoid "low real_memory" error in slurm
+        memory = instances[instance_type]["memory"]
+        # Get memory in GiB
+        memory = float(memory.split(" ")[0])
```
demartinofra: this should not be needed anymore

src/sqswatcher/sqswatcher.py
```diff
@@ -60,8 +60,8 @@ class QueryConfigError(Exception):
     ],
 )
 
-Host = collections.namedtuple("Host", ["instance_id", "hostname", "slots", "gpus"])
-
+Host = collections.namedtuple("Host", ["instance_id", "hostname", "slots", "gpus", "hw_info"])
+HWInfo = collections.namedtuple("HWInfo", ["memory"])
```
demartinofra: why not simply adding memory to the Host tuple?

rexcsn:
closing in favor of https://github.com/aws/aws-parallelcluster-node/pull/179

PR #194, opened 2020-01-16
Make ParallelCluster services more portable

jgleissner:
*Description of changes:*

This patch is intended to make ParallelCluster services work on other platforms, not just systems created by the chef bootstrap mechanism.

Mostly this means removing hard-coded paths and being more flexible about config file locations. I've also added systemd unit files and relaxed the requirements a little bit to allow the package to be installed on SLES 15 SP1.

### DIFF
--- a/src/nodewatcher/plugins/slurm.py
+++ b/src/nodewatcher/plugins/slurm.py
@@ -89,14 +89,7 @@ def is_node_down():
         # Output format:
         # down*
         hostname = socket.get_hostname()
-        command = [
-            "sinfo",
-            "--noheader",
-            "-o",
-            "%T",
-            "-n",
-            hostname
-        ]
+        command = ["sinfo", "--noheader", "-o", "%T", "-n", hostname]
         output = check_command_output(command).strip()
         log.info("Node is in state: '{0}'".format(output))
         if output and all(state not in output for state in ["down", "drained", "fail"]):
--- a/src/nodewatcher/plugins/slurm.py
+++ b/src/nodewatcher/plugins/slurm.py
@@ -10,6 +10,7 @@
 # limitations under the License.
 
 import logging
+import socket
 import subprocess
 
 from common.schedulers.slurm_commands import PENDING_RESOURCES_REASONS, get_pending_jobs_info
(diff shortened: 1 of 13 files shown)


src/sqswatcher/plugins/slurm.py
```diff
@@ -48,7 +49,7 @@ def _restart_master_node():
 
 def _restart_multiple_compute_nodes(hostnames, cluster_user):
     command = (
-        "if [ -f /etc/systemd/system/slurmd.service ]; "
+        "if [ -e /usr/lib/systemd ]; "
```
rjschwei: I think Ubuntu puts systemd in /lib. This would have to cope with both locations. Similar problems in other areas.

src/nodewatcher/plugins/slurm.py
```diff
@@ -99,10 +101,10 @@ def is_node_down():
 
 
 def _get_node_slots():
-    hostname = check_command_output("hostname")
+    hostname = socket.get_hostname()
     # retrieves number of slots for a specific node in the cluster.
     # Output format:
     # 4
-    command = "/opt/slurm/bin/sinfo -o '%c' -n {0} -h".format(hostname)
+    command = "sinfo -o '%c' -n {0} -h".format(hostname)
```
rjschwei: Hmm if for other distros "slurm" is installed from source and ends up in `/opt` then this is a breaking change as `/opt` is usually not on the PATH. Meaning the code either has to deal with multiple locations in some way or `/opt/slurm/bin` has to be added to PATH. Similar problems in other areas.

src/sqswatcher/plugins/slurm.py
jgleissner: Good to know. I guess both locations could be checked.

src/nodewatcher/plugins/slurm.py
jgleissner: Distributors shouldn't install anything in `/opt`. By removing the path to the binary we make the assumption that it is in the search patch, so if the binaries were installed in a non standard directory, it should be ensured that the service process has that to its search path added IMHO.

It seems that the `aws-parallelcluster` orchestration adds `/opt/slurm/bin` to the system wide search path, so this change shouldn't break anything.

demartinofra:
Feeding this PR to our testing infrastructure for a first validation.

demartinofra:
Tests are failing due to the fact that Slurm binaries cannot be found without using an absolute path. We need to investigate why the bash profile isn't loaded when running commands from node daemons.

demartinofra:
Closing since the integration with Slurm job scheduler has been heavily revised in newer versions. It is now possible to set Slurm binaries through the SLURM_BINARIES_DIR env var.

PR #201, opened 2020-03-20
Various enhancements and bug fixes to better handle scaling in bigger clusters

demartinofra:
1. jobwatcher - slurm: do not count down nodes as busy: In case a node is down and it's not in ASG this causes overscaling. Note that this condition is not triggered by nodes that are set to DRAIN by nodewatcher before being terminated. These nodes will preserve the drain state despite being unreachable (drain*). 
The downside to this is that when a node is unhealthy, hence set
to DOWN by the scheduler, its replacement is launched only after the node is terminated and removed from the scheduler, which means waiting for the SlurmdTimeout+nodewatcher timeout+processing of instance terminate event.

2. sqswatcher: prevent message retries from populating slurm nodes config with zombie nodes: Due to the retries on failed SQS events, under certain circumstances the retry on a ADD event can be processed after a successful REMOVE event for the same instance. This causes terminated instance types to be added back to slurm config and never removed.
With this change retries on failed ADD events are dropped and, in case of colliding messages for the same hostname, REMOVE events are always prevailing over ADD ones.

3. nodewatcher: do not dump logs to shared dir anymore in case of failures: CW is available starting pcluster 2.6.0

4. sqswatcher - slurm: increasing parallelism when restarting slurmd on compute nodes: from 10 to 30

5. jobwatcher: retrieve busy nodes before computing required ones: If jobs are scheduled to nodes between the moment the jobwatcher computes the number of requried nodes (based on the pending jobs) and when the list of busy nodes is fetched then the cluster over-scales.
Moving get_busy_nodes before get_required_nodes in order to prevent this from happening.

6. various additional improvements related to sqs messages management


src/sqswatcher/sqswatcher.py
```diff
@@ -333,7 +336,10 @@ def _process_sqs_messages(
 
     for event in failed_events:
         log.error("Failed when processing event %s for instance %s", event.action, event.host)
-        _requeue_message(queue, event.message)
+        if event.action == "REMOVE":
+            _requeue_message(queue, event.message)
```
tilne: The changes in _parse_sqs_messages ensure that during any one iteration of the sqswatcher loop, there won't be a REMOVE message processed after an ADD message for the same hostname. But does that cover stand across loop iterations as well? Is it possible that due to out of order delivery of SQS messages, we process an ADD during one iteration of the loop and then a REMOVE during the next iteration that was for a different instance ID?

src/sqswatcher/sqswatcher.py
```diff
@@ -184,7 +184,7 @@ def _requeue_message(queue, message):
     :param queue: the queue where to send the message
     :param message: the message to requeue
     """
-    max_retries = 2
+    max_retries = 1
```
demartinofra: move to another commit

src/sqswatcher/plugins/slurm.py
```diff
@@ -189,13 +189,13 @@ def update_cluster(max_cluster_size, cluster_user, update_events, instance_prope
...
     _restart_master_node()
-    results = _restart_multiple_compute_nodes(nodes_to_restart, cluster_user)
+    restart_compute_results = _restart_multiple_compute_nodes(nodes_to_restart, cluster_user)
     _reconfigure_nodes()
 
     failed = []
     succeeded = []
     for event in update_events:
-        if results.get(event.host.hostname, True):
+        if restart_compute_results.get(event.host.hostname, True) or event.action == EventType.REMOVE:
```
demartinofra: it was already managed by default True

src/sqswatcher/sqswatcher.py
demartinofra: This is a good point. We are using SQS queue in standard mode meaning there is no guarantee that the order of messages is preserved: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/standard-queues.html#standard-queues-message-order

However the fact that an instance terminate for a given instance usually happens some time after a compute ready makes the reordering very unlikely to happen. In any case switching to a FIFO queue is probably the more correct thing to do. I would just consider this enhancement as out of the scope of this change if you agree.

In any case as long as ADD events are not retried and REMOVE events corresponding to given ADDs are processed after the ADD, the configuration of the scheduler should be kept consistent. Worst case scenario (that is very unlikely) we discard an ADD event for a legitimate node and we need to wait for the nodewatcher to replace it.

src/sqswatcher/plugins/slurm.py
```diff
@@ -47,12 +47,13 @@ def _restart_master_node():
...
 
 def _restart_multiple_compute_nodes(hostnames, cluster_user):
+    log.info("Restarting slurmd on following compute nodes %s", hostnames)
     command = (
         "if [ -f /etc/systemd/system/slurmd.service ]; "
         "then sudo systemctl restart slurmd.service; "
         'else sudo sh -c "/etc/init.d/slurm restart 2>&1 > /tmp/slurmdstart.log"; fi'
     )
-    return RemoteCommandExecutor.run_remote_command_on_multiple_hosts(command, hostnames, cluster_user)
+    return RemoteCommandExecutor.run_remote_command_on_multiple_hosts(command, hostnames, cluster_user, parallelism=30)
```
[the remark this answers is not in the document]
demartinofra: This seemed to be a bottleneck when sqswatcher tries to add compute nodes that have been already terminated (because of the timeout on the ssh connection). I just increased it to a more reasonable number (tested empirically) without raising it too much and risk to hit other limits.

src/jobwatcher/plugins/utils.py
```diff
@@ -53,15 +53,13 @@ def get_optimal_nodes(nodes_requested, resources_requested, instance_properties)
                     resources_remaining_per_node[slot_idx][resource_type] -= job_resources_per_node[resource_type]
                 num_of_nodes -= 1
 
-        log.info("After looking at existing nodes, %s more nodes are needed" % num_of_nodes)
```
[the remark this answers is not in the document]
demartinofra: I added a log message at the end of the for loop.

src/jobwatcher/plugins/slurm.py
```diff
@@ -57,6 +57,6 @@ def get_busy_nodes():
     output = output.split("\n")
     for line in output:
         line_arr = line.split()
-        if len(line_arr) == 2 and (line_arr[1] in ["mix", "alloc", "down", "down*"]):
+        if len(line_arr) == 2 and (line_arr[1] in ["mix", "alloc"]):
```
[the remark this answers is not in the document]
demartinofra: I left a note in the commit message. Shall I also add a comment here?

src/sqswatcher/sqswatcher.py
```diff
@@ -239,27 +239,28 @@ def _parse_sqs_messages(sqs_config_region, sqs_config_proxy, messages, table, qu
...
 
+    update_events = OrderedDict()
+    for event in itertools.chain(add_events, remove_events):
+        log.info("Processing %s event for instance %s", event.action, event.host)
+        hostname = event.host.hostname
+        if hostname in update_events:
+            # events are looped in the order they are fetched and by iterating over REMOVE events after ADD events.
+            # in case of collisions let's always remove the item that comes first in order to always favor REMOVE ops.
+            log.info("Hostname collision. Discarding event %s in favour of event %s", update_events[hostname], event)
+            del update_events[hostname]
```
mauri-melato: What happens to the message associated to the overwritten event? 
I don't see we're removing it from the queue. 
This flaw is not specifically of this patch but that a potential source of for large clusters.

demartinofra: Good catch. colliding messages need to be removed from the queue.
In general to make this more robust we should filter messages based on the actual nodes available in ec2 so that we never process nodes that had been already terminated.

I added a quick fix for this: https://github.com/aws/aws-parallelcluster-node/pull/201/commits/210f6db95248aed59fa2a9828fce9dc113fb284a

PR #207, opened 2020-04-18
Handling EC2 Health Scheduled Events

rexcsn:
* Modified sqswatcher to retrieve and process messages from health queue. Lock instance if there is an EC2 health scheduled event, so that instance can be replaced by nodewatcher ASAP
* Modified nodewatcher to consider nodes that are locked and have no job as down nodes
* Modified unit tests to adapt to above modifications
* This feature can be turned off by specifying `disable_health_check = True` in `/etc/sqswatcher.cfg` and restarting sqswatcher

### DIFF
--- a/src/common/schedulers/slurm_commands.py
+++ b/src/common/schedulers/slurm_commands.py
@@ -12,11 +12,14 @@
 
 import logging
 import math
+import subprocess
 from textwrap import wrap
 
 from common.schedulers.converters import ComparableObject, from_table_to_obj_list
-from common.utils import check_command_output
+from common.utils import check_command_output, run_command
 
+SLURM_NODE_ERROR_STATES = ["down", "drained", "fail"]
+SLURM_NODE_DISABLED_STATES = ["draining", "drained"]
 PENDING_RESOURCES_REASONS = [
     "Resources",
     "Nodes required for job are DOWN, DRAINED or reserved for jobs in higher priority partitions",
@@ -57,6 +60,19 @@ def get_jobs_info(job_state_filter=None):
 
     output = check_command_output(command)
     return SlurmJob.from_table(output)
+
+
+def get_node_state(hostname):
+    # retrieves the state of a specific node
+    # https://slurm.schedmd.com/sinfo.html#lbAG
+    # Output format:
+    # down*
+    try:
+        command = "/bin/bash -c \"/opt/slurm/bin/sinfo --noheader -o '%T' -n {}\"".format(hostname)
+        output = check_command_output(command).strip()
+        return output
+    except Exception as e:
+        logging.error("Failed when checking if node {} state with exception {}.".format(hostname, e))
 
 
 def get_pending_jobs_info(
@@ -267,6 +283,33 @@ def job_runnable_on_given_node(job_resources_per_node, resources_available, exis
             return False
 
     return True
+
+
+def lock_node(hostname, unlock=False, note=None):
+    # hostname format: ip-10-0-0-114.eu-west-1.compute.internal
+    hostname = hostname.split(".")[0]
+    if unlock:
+        logging.info("Unlocking host %s", hostname)
+        command = [
+            "/opt/slurm/bin/scontrol",
+            "update",
+            "NodeName={0}".format(hostname),
+            "State=RESUME",
+            "Reason={}".format(note if note else '"Unlocking"'),
+        ]
+    else:
+        logging.info("Locking host %s", hostname)
+        command = [
+            "/opt/slurm/bin/scontrol",
+            "update",
+            "NodeName={0}".format(hostname),
+            "State=DRAIN",
+            "Reason={}".format(note if note else '"Shutting down"'),
+        ]
+    try:
+        run_command(command)
+    except subprocess.CalledProcessError:
+        logging.error("Error %s host %s", "unlocking" if unlock else "locking", hostname)
 
 
 class SlurmJob(ComparableObject):
--- a/src/common/schedulers/slurm_commands.py
+++ b/src/common/schedulers/slurm_commands.py
@@ -68,11 +68,12 @@ def get_node_state(hostname):
     # Output format:
     # down*
     try:
-        command = "/bin/bash -c \"/opt/slurm/bin/sinfo --noheader -o '%T' -n {}\"".format(hostname)
+        command = "/opt/slurm/bin/sinfo --noheader -o '%T' -n {}".format(hostname)
         output = check_command_output(command).strip()
         return output
     except Exception as e:
-        logging.error("Failed when checking if node {} state with exception {}.".format(hostname, e))
+        logging.error("Failed when checking if node %s state with exception %s.", hostname, e)
+        raise
 
 
 def get_pending_jobs_info(
@@ -285,31 +286,39 @@ def job_runnable_on_given_node(job_resources_per_node, resources_available, exis
     return True
 
 
-def lock_node(hostname, unlock=False, note=None):
+def lock_node(hostname, reason=None):
     # hostname format: ip-10-0-0-114.eu-west-1.compute.internal
     hostname = hostname.split(".")[0]
-    if unlock:
-        logging.info("Unlocking host %s", hostname)
-        command = [
-            "/opt/slurm/bin/scontrol",
-            "update",
-            "NodeName={0}".format(hostname),
-            "State=RESUME",
-            "Reason={}".format(note if note else '"Unlocking"'),
-        ]
-    else:
-        logging.info("Locking host %s", hostname)
-        command = [
-            "/opt/slurm/bin/scontrol",
-            "update",
-            "NodeName={0}".format(hostname),
-            "State=DRAIN",
-            "Reason={}".format(note if note else '"Shutting down"'),
-        ]
+    logging.info("Locking host %s", hostname)
+    command = [
+        "/opt/slurm/bin/scontrol",
+        "update",
+        "NodeName={0}".format(hostname),
+        "State=DRAIN",
+        "Reason={}".format(reason if reason else '"Shutting down"'),
+    ]
     try:
         run_command(command)
     except subprocess.CalledProcessError:
-        logging.error("Error %s host %s", "unlocking" if unlock else "locking", hostname)
+        logging.error("Error locking host %s", hostname)
+        raise
+
+
+def unlock_node(hostname, reason=None):
+    hostname = hostname.split(".")[0]
+    logging.info("Unlocking host %s", hostname)
+    command = [
+        "/opt/slurm/bin/scontrol",
+        "update",
+        "NodeName={0}".format(hostname),
+        "State=RESUME",
+        "Reason={}".format(reason if reason else '"Unlocking"'),
+    ]
+    try:
+        run_command(command)
+    except subprocess.CalledProcessError:
+        logging.error("Error unlocking host %s", hostname)
+        raise
 
 
 class SlurmJob(ComparableObject):
--- a/src/common/schedulers/slurm_commands.py
+++ b/src/common/schedulers/slurm_commands.py
@@ -67,13 +67,9 @@ def get_node_state(hostname):
     # https://slurm.schedmd.com/sinfo.html#lbAG
     # Output format:
     # down*
-    try:
-        command = "/opt/slurm/bin/sinfo --noheader -o '%T' -n {}".format(hostname)
-        output = check_command_output(command).strip()
-        return output
-    except Exception as e:
-        logging.error("Failed when checking if node %s state with exception %s.", hostname, e)
-        raise
+    command = "/opt/slurm/bin/sinfo --noheader -o '%T' -n {}".format(hostname)
+    output = check_command_output(command).strip()
+    return output
 
 
 def get_pending_jobs_info(
@@ -301,7 +297,6 @@ def lock_node(hostname, reason=None):
         run_command(command)
     except subprocess.CalledProcessError:
         logging.error("Error locking host %s", hostname)
-        raise
 
 
 def unlock_node(hostname, reason=None):
@@ -318,7 +313,6 @@ def unlock_node(hostname, reason=None):
         run_command(command)
     except subprocess.CalledProcessError:
         logging.error("Error unlocking host %s", hostname)
-        raise
 
 
 class SlurmJob(ComparableObject):
--- a/src/common/utils.py
+++ b/src/common/utils.py
@@ -39,11 +39,22 @@ class CriticalError(Exception):
 class EventType(Enum):
     ADD = "ADD"
     REMOVE = "REMOVE"
+    HEALTH = "HEALTH"
 
+
+class QueueType(Enum):
+    instance = "instance"
+    health = "health"
+
+
+SUPPORTED_EVENTTYPE_FOR_QUEUETYPE = {
+    QueueType.instance: [EventType.ADD, EventType.REMOVE],
+    QueueType.health: [EventType.HEALTH],
+}
 
 Host = collections.namedtuple("Host", ["instance_id", "hostname", "slots", "gpus"])
 UpdateEvent = collections.namedtuple("UpdateEvent", ["action", "message", "host"])
-INSTANCE_ALIVE_STATE = ["pending", "running"]
+INSTANCE_ALIVE_STATES = ["pending", "running"]
 
 
 def load_module(module):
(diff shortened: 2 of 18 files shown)


src/sqswatcher/sqswatcher.py
```diff
@@ -203,9 +221,9 @@ def _requeue_message(queue, message):
 
 
 def _retrieve_all_sqs_messages(queue, max_processed_messages):
-    log.info("Retrieving messages from SQS queue")
```
demartinofra, line 206: Rather than removing this I would parametrize this log message with the queue name as done below

src/sqswatcher/plugins/slurm.py
```diff
@@ -208,5 +209,50 @@ def update_cluster(max_cluster_size, cluster_user, update_events, instance_prope
...
+    return False
+
+
+def perform_health_actions(health_events):
+    """Update and write node lists( and gres_nodes if instance has GPU); restart relevant nodes."""
+    failed = []
+    succeeded = []
+    for event in health_events:
+        try:
+            # to-do, ignore fail to lock message if node is not in scheduler
```
demartinofra, line 227: I think this is an important one. We need to make sure such actions are idempotent and do not fail if a node is not in scheduler especially since the process instance events might have just removed it.

src/sqswatcher/plugins/slurm.py
```diff
@@ -208,5 +209,50 @@ def update_cluster(max_cluster_size, cluster_user, update_events, instance_prope
...
+            if _is_node_locked(event.host.hostname):
+                log.error(
```
demartinofra, line 229: why is this an error?also do we need this check or we can go ahead an lock the node?

src/sqswatcher/plugins/slurm.py
```diff
@@ -208,5 +209,50 @@ def update_cluster(max_cluster_size, cluster_user, update_events, instance_prope
...
+                    "Instance {}/{} currently in disabled state 'draining/drained'. "
+                    "Risk of lock being released by nodewatcher if locking the node because of scheduled event now. "
+                    "Marking event as failed to retry later.".format(event.host.instance_id, event.host.hostname)
+                )
+                failed.append(event)
+                continue
+            reason = "DO_NOT_CHANGE_SQSWATCHER_SCHEDULED_EVENT_ACTION"
```
demartinofra, line 236: are we using this reason anywhere else?We could also make it more explicit such as "Node requires replacement due to an EC2 scheduled maintenance event."

src/sqswatcher/plugins/slurm.py
```diff
@@ -208,5 +209,50 @@ def update_cluster(max_cluster_size, cluster_user, update_events, instance_prope
...
+            lock_node(event.host.hostname, unlock=False, note=reason)
+            if _is_node_locked(event.host.hostname):
```
demartinofra, line 238: is this required or we can rely on the exit code of the locking command? I'm just trying to minimize the calls we make to the scheduler

src/sqswatcher/plugins/slurm.py
```diff
@@ -208,5 +209,50 @@ def update_cluster(max_cluster_size, cluster_user, update_events, instance_prope
...
+                succeeded.append(event)
+                log.info(
+                    "Successfully locked {} in response to scheduled maintainence event".format(event.host.hostname)
+                )
+            else:
+                failed.append(event)
+                log.info("Failed to lock {} in response to scheduled maintainence event".format(event.host.hostname))
```
demartinofra, line 245: this could be an error actually :)

src/sqswatcher/plugins/slurm.py
```diff
@@ -208,5 +209,50 @@ def update_cluster(max_cluster_size, cluster_user, update_events, instance_prope
...
+def _is_node_locked(hostname):
+    node_state = get_node_state(hostname)
+    log.info("Node {} currently in state {}".format(hostname, node_state))
+    for disable_state in SLURM_NODE_DISABLED_STATES:
+        if disable_state in node_state:
+            return True
+    return False
+
+
+def perform_health_actions(health_events):
```
demartinofra, line 221: minor: what about a more meaningful name such as handle_maintenance_events?

src/sqswatcher/plugins/slurm.py
rexcsn: Right now if locking fails for any reason, we will catch it with exception, mark the event as failed, and re-queue the event. If the instance is already removed we will re-queue the event until it's eventually discarded.

Do we need to handle this case with specific logic or is using the general re-queue when failed logic ok?

src/sqswatcher/sqswatcher.py
```diff
@@ -220,51 +238,104 @@ def _retrieve_all_sqs_messages(queue, max_processed_messages):
...
+            )
+            health_update_events[health_event.host.instance_id].message.delete()
+            del health_update_events[health_event.host.instance_id]
+        health_update_events[health_event.host.instance_id] = health_event
+
+    return health_update_events.values()
+
+
+def _parse_messages_helper(messages, is_health_queue, sqs_config, table=None, queue=None, instance_id_to_hostname=None):
+    parsed_events = {
```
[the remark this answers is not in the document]
rexcsn: Done

src/common/schedulers/slurm_commands.py
```diff
@@ -269,6 +285,33 @@ def job_runnable_on_given_node(job_resources_per_node, resources_available, exis
...
+            "/opt/slurm/bin/scontrol",
+            "update",
+            "NodeName={0}".format(hostname),
+            "State=DRAIN",
+            "Reason={}".format(note if note else '"Shutting down"'),
+        ]
+    try:
+        run_command(command)
+    except subprocess.CalledProcessError:
+        logging.error("Error %s host %s", "unlocking" if unlock else "locking", hostname)
```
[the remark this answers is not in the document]
demartinofra: if we start raising we need to go through all callers and make sure the old behavior is preserved. We don't want the nodewatcher to crash when failing to lock a node but we need to properly handle the error

src/sqswatcher/sqswatcher.py
```diff
@@ -220,51 +238,104 @@ def _retrieve_all_sqs_messages(queue, max_processed_messages):
...
             update_events[hostname].message.delete()
             del update_events[hostname]
         update_events[hostname] = event
 
     return update_events.values()
 
 
+def _parse_health_messages(sqs_config, messages, health_queue):
+    health_update_events = OrderedDict()
+    instance_id_to_hostname = get_cluster_instance_info(sqs_config.stack_name, sqs_config.region, include_master=False)
```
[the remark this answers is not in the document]
demartinofra: I would have personally extracted the instance-ids from all health messages and then performed a single (or a few if needed) ec2 calls by passing an array of instance-ids. 
I expect health messages to be always a few (more like 1 at a time) and, in case of big clusters, describing all instances rather than 1 seems a bit overkill. What do you think?

of if it gets to complicated to parse the messages twice you could also perform a describe instance for each message in the queue since we don't expect them to be many

src/common/utils.py
```diff
@@ -389,3 +401,31 @@ def retrieve_max_cluster_size(region, proxy_config, asg_name, fallback):
...
+                    is_alive = instance.get("State").get("Name") in INSTANCE_ALIVE_STATES
+                    instance_id = instance.get("InstanceId")
+                    if is_alive:
+                        instances_in_cluster.append(instance_id)
+
+        return instances_in_cluster
+
+    except Exception as e:
+        logging.error("Failed retrieving instance_ids for cluster %s with exception: %s", stack_name, e)
+        raise
```
demartinofra, line 431: is the sqswatcher crashing if this fails? if so we should handle failures in handling such messages with the same requeuing mechanism we use for the other events

src/common/utils.py
```diff
@@ -389,3 +401,31 @@ def retrieve_max_cluster_size(region, proxy_config, asg_name, fallback):
...
+        instance_paginator = ec2_client.get_paginator("describe_instances")
+        nodes_to_include = ["Compute", "Master"] if include_master else ["Compute"]
+
+        for page in instance_paginator.paginate(
+            Filters=[
+                {"Name": "tag:Application", "Values": [stack_name]},
+                {"Name": "tag:Name", "Values": nodes_to_include},
+            ]
+        ):
+            for reservation in page.get("Reservations"):
```
demartinofra, line 420: you want to double-check this cause I remember I was not getting Reservations but Instances directly when using the paginator. If I'm correct on this can we understand why this is not caught by integ tests?

src/nodewatcher/plugins/slurm.py
```diff
@@ -54,44 +61,28 @@ def has_pending_jobs(instance_properties, max_size):
...
-        command = [
-            "/opt/slurm/bin/scontrol",
-            "update",
-            "NodeName={0}".format(hostname),
-            "State=DRAIN",
-            'Reason="Shutting down"',
-        ]
-    try:
-        run_command(command)
-    except subprocess.CalledProcessError:
```
demartinofra, line 79: This was previously swallowing the exception in case of failure. Now we are not. Let's make sure we properly handle the error cases

src/nodewatcher/plugins/torque.py
```diff
@@ -56,14 +56,7 @@ def has_pending_jobs(instance_properties, max_size):
...
 def lock_host(hostname, unlock=False):
-    # hostname format: ip-10-0-0-114.eu-west-1.compute.internal
-    hostname = hostname.split(".")[0]
-    mod = unlock and "-c" or "-o"
-    command = [TORQUE_BIN_DIR + "pbsnodes", mod, hostname]
-    try:
-        run_command(command)
-    except subprocess.CalledProcessError:
-        log.error("Error %s host %s", "unlocking" if unlock else "locking", hostname)
+    lock_node(hostname, unlock=unlock)
```
demartinofra, line 59: same here. I'll let you double-check all of them.

src/sqswatcher/sqswatcher.py
```diff
@@ -223,51 +248,108 @@ def _retrieve_all_sqs_messages(queue, max_processed_messages):
...
+            log.info("Unsupported event type %s for %s queue. Discarding message.", event_type, queue_type)
+            message.delete()
+
+    return parsed_events
+
+
+def _process_scheduled_maintenance_event(message_attrs, message, table, instances_in_cluster):
+    instance_id = message_attrs.get("EC2InstanceId")
+    if instance_id in instances_in_cluster:
+        hostname = _retrieve_hostname_from_ddb(instance_id, table)
```
demartinofra, line 344: Probably we want to handle the case when hostname is None. Node could have been already removed. Shall we discard the message?

src/common/utils.py
rexcsn: Response does include Reservations

src/common/utils.py
rexcsn: `sqswatcher` will handle this in `_process_scheduled_maintenance_event`

src/nodewatcher/plugins/slurm.py
rexcsn: Preserving old behavior, locking action now will not raise error, `sqswatcher` and `nodewatcher` can handle locking failures on their own without catching the error

src/nodewatcher/plugins/torque.py
rexcsn: Done, will not raise error

src/sqswatcher/sqswatcher.py
```diff
@@ -206,68 +232,145 @@ def _requeue_message(queue, message):
...
+
+def _process_scheduled_maintenance_event(stack_name, region, proxy_config, message_attrs, message, table, queue):
+    try:
+        instance_id = message_attrs.get("EC2InstanceId")
+        instances_in_cluster = get_cluster_instance_info(
+            stack_name, region, proxy_config, instance_ids=[instance_id], include_master=False
+        )
+        if instance_id in instances_in_cluster:
+            hostname = _retrieve_hostname_from_ddb(instance_id, table)
+            log.info("Relevant EC2 scheduled event for instance:%s in ASG.", instance_id)
```
demartinofra, line 356: minor: remove ASG at the end? also might want to add a space after `:`

src/sqswatcher/plugins/sge.py
```diff
@@ -99,5 +103,52 @@ def update_cluster(max_cluster_size, cluster_user, update_events, instance_prope
...
+
+
+def perform_health_actions(health_events):
+    """Update and write node lists( and gres_nodes if instance has GPU); restart relevant nodes."""
+    failed = []
+    succeeded = []
+    for event in health_events:
+        try:
+            if _is_node_locked(event.host.hostname):
+                log.error(
```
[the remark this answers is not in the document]
rexcsn: Done

src/sqswatcher/plugins/sge.py
```diff
@@ -99,5 +103,52 @@ def update_cluster(max_cluster_size, cluster_user, update_events, instance_prope
...
+                failed.append(event)
+                continue
+            lock_node(event.host.hostname)
+            if _is_node_locked(event.host.hostname):
+                succeeded.append(event)
+                log.info("Successfully locked %s in response to scheduled maintainence event", event.host.hostname)
+            else:
+                failed.append(event)
+                log.error("Failed to lock %s in response to scheduled maintainence event", event.host.hostname)
+        except Exception as e:
```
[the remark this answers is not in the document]
rexcsn: `lock_node` only catches `subprocess.CalledProcessError`. I think it's still worth having this try block here just to be safe. Changed error message.

src/sqswatcher/plugins/torque.py
```diff
@@ -52,6 +55,56 @@ def update_cluster(max_cluster_size, cluster_user, update_events, instance_prope
...
+
+
+def perform_health_actions(health_events):
+    """Update and write node lists( and gres_nodes if instance has GPU); restart relevant nodes."""
+    failed = []
+    succeeded = []
+    for event in health_events:
+        try:
+            if _is_node_locked(event.host.hostname):
+                log.error(
```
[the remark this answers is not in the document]
rexcsn: Done

src/sqswatcher/sqswatcher.py
```diff
@@ -188,7 +210,11 @@ def _requeue_message(queue, message):
     :param message: the message to requeue
     """
     max_retries = 1
-    message_body = json.loads(message.body)
+    # CW event rule can only send messages to SQS with "" enclosing the message.
```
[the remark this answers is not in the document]
rexcsn: Yes; unfortunately that is the case

src/sqswatcher/sqswatcher.py
rexcsn: Done

src/sqswatcher/plugins/sge.py
```diff
@@ -99,5 +103,52 @@ def update_cluster(max_cluster_size, cluster_user, update_events, instance_prope
...
+
+def perform_health_actions(health_events):
+    """Update and write node lists( and gres_nodes if instance has GPU); restart relevant nodes."""
+    failed = []
+    succeeded = []
+    for event in health_events:
+        try:
+            if _is_node_locked(event.host.hostname):
+                log.warning(
+                    "Instance %s/%s currently in disabled state 'd'. "
```
enrico-usai, line 127: in disabled state 'd'.  --> in the message we should use the SGE_DISABLED_STATE constant defined before.

src/sqswatcher/plugins/slurm.py
```diff
@@ -208,5 +209,54 @@ def update_cluster(max_cluster_size, cluster_user, update_events, instance_prope
     return failed, succeeded
 
 
+def _is_node_locked(hostname):
+    try:
+        node_state = get_node_state(hostname)
+        log.info("Node %s currently in state %s", hostname, node_state)
+        for disable_state in SLURM_NODE_DISABLED_STATES:
+            if disable_state in node_state:
```
enrico-usai, line 217: same here.
We can have a states like: State=IDLE+DRAIN (see: https://bugs.schedmd.com/show_bug.cgi?id=291) and the disable_state could be a substring of a more complex state.

Let me know if it's by design.

src/sqswatcher/plugins/slurm.py
```diff
@@ -208,5 +209,54 @@ def update_cluster(max_cluster_size, cluster_user, update_events, instance_prope
...
+
+def perform_health_actions(health_events):
+    """Update and write node lists( and gres_nodes if instance has GPU); restart relevant nodes."""
+    failed = []
+    succeeded = []
+    for event in health_events:
+        try:
+            if _is_node_locked(event.host.hostname):
+                log.warning(
+                    "Instance %s/%s currently in disabled state 'draining/drained'. "
```
enrico-usai, line 234: disabled state 'draining/drained'. --> we should use in the message the constant SLURM_NODE_DISABLED_STATES defined before.

src/sqswatcher/plugins/torque.py
```diff
@@ -52,6 +55,56 @@ def update_cluster(max_cluster_size, cluster_user, update_events, instance_prope
...
+
+def perform_health_actions(health_events):
+    """Update and write node lists( and gres_nodes if instance has GPU); restart relevant nodes."""
+    failed = []
+    succeeded = []
+    for event in health_events:
+        try:
+            if _is_node_locked(event.host.hostname):
+                log.warning(
+                    "Instance %s/%s currently in disabled state 'offline'. "
```
enrico-usai, line 78: same here for the message.

src/sqswatcher/plugins/torque.py
```diff
@@ -52,6 +55,56 @@ def update_cluster(max_cluster_size, cluster_user, update_events, instance_prope
     return failed, succeeded
 
 
+def _is_node_locked(hostname):
+    try:
+        node = get_compute_nodes_info(hostname_filter=[hostname]).get(hostname)
+        if TORQUE_NODE_DISABLED_STATE in node.state:
```
enrico-usai, line 61: same here.

src/sqswatcher/sqswatcher.py
```diff
@@ -130,6 +150,8 @@ def _get_sqs_queue(region, queue_name, proxy_config):
     try:
         queue = sqs.get_queue_by_name(QueueName=queue_name)
         log.debug("SQS queue is %s", queue)
+        # Set queue_name attribute
+        queue.queue_name = queue.url.split("/")[-1]
```
enrico-usai, line 154: Could it raise an "index out of range" error?

src/nodewatcher/plugins/torque.py
```diff
@@ -74,6 +67,17 @@ def is_node_down():
         if node:
             log.info("Node is in state: '{0}'".format(node.state))
             if all(error_state not in node.state for error_state in TORQUE_NODE_ERROR_STATES):
+                # Consider the node down if it is in Disabled state placed by scheduled event
+                # and does not have job
+                if TORQUE_NODE_DISABLED_STATE in node.state and not has_jobs(hostname):
+                    log.warning(
+                        (
+                            "Considering node as down because there is no job running and node is in a disabled state. "
```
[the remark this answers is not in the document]
rexcsn: Putting the warning message in common.utils

src/sqswatcher/plugins/sge.py
```diff
@@ -99,5 +103,52 @@ def update_cluster(max_cluster_size, cluster_user, update_events, instance_prope
     return failed, succeeded
 
 
+def _is_node_locked(hostname):
+    try:
+        node_info = get_compute_nodes_info(hostname_filter=hostname)
+        node = node_info.get(socket.getfqdn(hostname), node_info.get(hostname))
+        if SGE_DISABLED_STATE in node.state:
```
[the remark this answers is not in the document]
rexcsn: Using `in` instead of `==` is by design, because a node can have hybrid states and we just want to check if the locking state is in the hybrid states in order to check if a node is locked or not.

src/sqswatcher/plugins/sge.py
rexcsn: Done

src/sqswatcher/plugins/slurm.py
rexcsn: This is by design, we only want to check if the locking state is in the node states, and we don't care if the node has other states.

src/sqswatcher/plugins/slurm.py
rexcsn: Done, also moving the message to common.utils

PR #215, opened 2020-06-10
Slurm Cloud Bursting Resume/Suspend Scripts

rexcsn:
* Implement slurm resume program, using `run_instances` to provision EC2 instances for slurm cloud bursting nodes.
* Implement slurm suspend program to terminate instances and resume slurm nodes.


src/slurm_cloud_bursting/resume.py
```diff
@@ -0,0 +1,216 @@
...
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("nodes", help="Nodes to burst")
+    args = parser.parse_args()
+    try:
+        # silence module logging
+        for logger in logging.Logger.manager.loggerDict:
+            logging.getLogger(logger).setLevel(logging.WARNING)
+
+        logging.basicConfig(
```
demartinofra, line 204: let's export logging configuration to a logging config file

src/slurm_cloud_bursting/suspend.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+def suspend(arg_nodes, suspend_config):
+    """Suspend and terminate nodes requested by slurm."""
+    logging.info("Suspending nodes:" + arg_nodes)
+
+    # Retrieve SlurmNode objects from slurm nodelist notation
+    slurm_nodes = get_nodes_info(arg_nodes)
+    logging.debug("Slurm_nodes = %s", slurm_nodes)
+    slurm_nodes = list(filter(lambda n: "DRAIN" not in n.state, slurm_nodes))
```
demartinofra, line 137: I think for now it's useless to filter out nodes in drain state

src/slurm_cloud_bursting/suspend.py
```diff
@@ -0,0 +1,174 @@
...
+            _set_nodes_to_down([instance_ids_to_nodename[instance_id] for instance_id in instances])
+
+
+def _get_instance_ids_to_nodename(slurm_nodes, region, boto3_config):
+    """Retrieve dict that maps from instance ids to slurm nodenames."""
+    node_ip_to_name = {node.nodeaddr: node.name for node in slurm_nodes}
+    ec2_client = boto3.client("ec2", region_name=region, config=boto3_config)
+    paginator = ec2_client.get_paginator("describe_instances")
+    response_iterator = paginator.paginate(
+        Filters=[{"Name": "private-ip-address", "Values": list(node_ip_to_name.keys())}],
```
demartinofra, line 96: is the paginator taking care also of handling very long lists?

Also, I would feel safer if we filter by some tag (such as ClusterName), just to make sure we are not terminating for any reason nodes that do not belong to parallelcluster

src/slurm_cloud_bursting/suspend.py
```diff
@@ -0,0 +1,174 @@
...
+            f"{SCONTROL} update nodename={node.name} nodehostname={node.name} nodeaddr={node.name} state=resume"
+        )
+        # Do not raise on error.
+        # Failure for resume command will fail if node already in IDLE, ignore failure.
+        run_command(resume_nodes_command, raise_on_error=False)
+
+
+def _set_nodes_to_down(slurm_nodes):
+    """
+    Fall back mechanism to handle failure when removing instances.
```
demartinofra, line 119: when a node is in powering down state it's not used by the scheduler until it is resumed. Rather than setting it as down I would simply resume the node.

src/slurm_cloud_bursting/suspend.py
```diff
@@ -0,0 +1,174 @@
...
+    return SlurmSuspendConfig(region, cluster_name, Config(**boto3_config), max_batch_size)
+
+
+def delete_instances(instance_ids_to_nodename, region, boto3_config, batch_size):
+    """Terminate corresponding EC2 instances."""
+    ec2_client = boto3.client("ec2", region_name=region, config=boto3_config)
+    logging.info("Terminating instances %s", list(instance_ids_to_nodename.keys()))
+    for instances in grouper(instance_ids_to_nodename.keys(), batch_size):
+        try:
+            ec2_client.terminate_instances(InstanceIds=instances,)
```
demartinofra, line 84: If you specify multiple instances and the request fails (for example, because of a single incorrect instance ID), none of the instances are terminated. Not sure if we could better handle the error conditions tbh. I'm wondering if we want to retry on failures at the very least (or verify on what error conditions boto3 client retries automatically)

src/slurm_cloud_bursting/resume.py
```diff
@@ -0,0 +1,262 @@
...
+            )
+            failed_nodes.append(slurm_node)
+
+
+def _update_run_instance_args(run_instances_args, cluster_name, queue, instance_type):
+    # LaunchTemplate is different for every instance type in every queue
+    # LaunchTemplate name format: {cluster_name}-{queue_name}-{instance_type}
+    # Sample LT name: hit-queue1-c5.xlarge
+    run_instances_args["LaunchTemplate"] = {"LaunchTemplateName": f"{cluster_name}-{queue}-{instance_type}"}
+    run_instances_args["InstanceType"] = instance_type
```
demartinofra, line 152: Since now we have a separate LT for each instance type I think we only need to select the right LT, so probably we can get rid of this  function

src/slurm_cloud_bursting/resume.py
```diff
@@ -0,0 +1,262 @@
...
+        private_ip=instance["PrivateIpAddress"],
+        hostname=instance["PrivateDnsName"].split(".")[0],
+    )
+
+
+def _update_slurm_node_addrs(launched_nodes):
+    """Update node information in slurm with info from launched EC2 instance."""
+    for instance, slurm_node in launched_nodes:
+        try:
+            update_nodes(slurm_node, nodeaddr=instance.private_ip, nodehostname=instance.hostname, raise_on_error=True)
```
demartinofra, line 128: can you remind me why we are not batching these calls? If it's to avoid failing the update on all nodes if individual nodes fail I'm wondering if we can do some optimization such as updating in batch and if it fails falling back to individual node updates. The reason is that I expect Slurm might handle batch updates in a more optimized fashion.

src/slurm_cloud_bursting/resume.py
```diff
@@ -0,0 +1,262 @@
...
+
+    If this process is not done explicitly, slurm will wait until ResumeTimeout,
+    then execute this process of setting nodes to DOWN then POWER_DOWN.
+    To save time, should explicitly set nodes to DOWN then POWER_DOWN after encountering failure.
+    """
+    try:
+        log.info("Node %s marked as down and placed into power_down", node_list)
+        _set_nodes_down_and_power_save(node_list)
+    except Exception as e:
+        log.exception("Failed to place nodes %s into down/power_down with exception: %s", node_list, e)
```
demartinofra, line 112: do we need the stack trace here?

src/slurm_cloud_bursting/resume.py
```diff
@@ -0,0 +1,262 @@
...
+    To do this, set node to DOWN, so slurm will automatically re-queue job.
+    Then set node to POWER_DOWN so suspend program will be run.
+    Suspend program needs to properly clean up instances(if any) and set node back to IDLE in all cases.
+
+    If this process is not done explicitly, slurm will wait until ResumeTimeout,
+    then execute this process of setting nodes to DOWN then POWER_DOWN.
+    To save time, should explicitly set nodes to DOWN then POWER_DOWN after encountering failure.
+    """
+    try:
+        log.info("Node %s marked as down and placed into power_down", node_list)
```
demartinofra, line 109: nit: this is a list of nodes so probably we want to append it at the end of the log message

src/slurm_cloud_bursting/resume.py
```diff
@@ -0,0 +1,262 @@
...
+from slurm_cloud_bursting.utils import CONFIG_FILE_PATH
+
+TAG_SPECIFICATIONS = [
+    {
+        "ResourceType": "instance",
+        "Tags": [{"Key": "aws-parallelcluster-slurm-node-type", "Value": "cloud-bursting-compute"}],
+    }
+]
+EC2Instance = collections.namedtuple("EC2Instance", ["id", "private_ip", "hostname"])
+LOG_CONFIG_FILE = "/opt/parallelcluster/configs/slurm/parallelcluster_resume_logging.conf"
```
demartinofra, line 36: as discussed this probably should be in the config file and in case not found default to a default logging config that needs to be shipped with the code itself.

src/slurm_cloud_bursting/suspend.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+class SlurmSuspendConfig:
+    DEFAULT_MAX_RETRY = 5
+    DEFAULT_MAX_INSTANCES_BATCH_SIZE = 100
+
+    def __init__(self, config_file_path=None, **kwargs):
+        if config_file_path:
+            self._get_config(config_file_path)
+        else:
```
demartinofra, line 37: if this is needed only for testing what we can do is to have a test for SlurmSuspendConfig that provides a set of files and makes sure this logic is correct and then use a mock for all other functions that require a SlurmSuspendConfig object.

src/slurm_cloud_bursting/suspend.py
```diff
@@ -0,0 +1,153 @@
...
+    When encountering a failure, update the node state to IDLE so the node can be used for future jobs.
+    Cloud-sync daemon will be responsible for removing orphaned nodes.
+    """
+    log.info("Resuming the following nodes back to IDLE: %s", slurm_nodenames)
+    set_nodes_idle(slurm_nodenames, reset_node_addrs_hostname=True)
+
+
+def _suspend(arg_nodes, suspend_config):
+    """Suspend and terminate nodes requested by slurm."""
+    log.info("Suspending nodes:" + arg_nodes)
```
demartinofra, line 119: nit: space after :

src/slurm_cloud_bursting/suspend.py
```diff
@@ -0,0 +1,153 @@
...
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("nodes", help="Nodes to release")
+    args = parser.parse_args()
+    try:
+        # Configure root logger
+        fileConfig(LOG_CONFIG_FILE, disable_existing_loggers=False)
+        suspend_config = SlurmSuspendConfig(CONFIG_FILE_PATH)
+        _suspend(args.nodes, suspend_config)
+    except Exception as e:
```
demartinofra, line 147: same here for error code

src/slurm_cloud_bursting/suspend.py
```diff
@@ -0,0 +1,153 @@
...
+        log.exception("Encountered exception when suspending instances for %s: %s", args.nodes, e)
+        _set_nodes_idle(args.nodes)
```
demartinofra, line 149: this can be in a finally statement since you always want to run it at the end

src/slurm_cloud_bursting/suspend.py
```diff
@@ -0,0 +1,153 @@
...
+from common.utils import grouper
+from slurm_cloud_bursting.utils import CONFIG_FILE_PATH
+
+LOG_CONFIG_FILE = "/opt/parallelcluster/configs/slurm/parallelcluster_suspend_logging.conf"
+log = logging.getLogger(__name__)
+
+
+class SlurmSuspendConfig:
+    DEFAULT_MAX_RETRY = 5
+    DEFAULT_MAX_INSTANCES_BATCH_SIZE = 100
```
demartinofra, line 32: why not setting to 1000?

src/slurm_cloud_bursting/suspend.py
```diff
@@ -0,0 +1,153 @@
...
+def _delete_instances(instance_ids_to_nodename, region, boto3_config, batch_size):
+    """Terminate corresponding EC2 instances."""
+    ec2_client = boto3.client("ec2", region_name=region, config=boto3_config)
+    log.info("Terminating instances %s", list(instance_ids_to_nodename.keys()))
+    for instances in grouper(instance_ids_to_nodename.keys(), batch_size):
+        try:
+            # Boto3 clients retries on connection errors only
+            # Adding extra layer of retry on all exceptions to try to terminate instances
+            retry(stop_max_attempt_number=3, wait_fixed=5000)(ec2_client.terminate_instances)(InstanceIds=instances,)
+        except ClientError as e:
```
demartinofra, line 83: I might be wrong but I think when using the retry function the error is wrapped in a RetryError exception. 
Btw I know probably this change came from one of my comments but maybe we can drop the retry here and have the daemon take care of unterminated instances.

src/slurm_cloud_bursting/suspend.py
```diff
@@ -0,0 +1,153 @@
...
+            log.error("Failed when terminating instances %s with error %s", instances, e)
```
demartinofra, line 84: also in this case we probably want to exit the script with error. Shall we return an encountered_error flag and handle this in the caller?

tests/slurm_cloud_bursting/test_suspend.py
```diff
@@ -0,0 +1,87 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
demartinofra, line 1: Same comment here as for resume tests.

src/slurm_cloud_bursting/resume.py
```diff
@@ -0,0 +1,262 @@
...
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("nodes", help="Nodes to burst")
+    args = parser.parse_args()
+    try:
+        # Configure root logger
+        fileConfig(LOG_CONFIG_FILE, disable_existing_loggers=False)
+        resume_config = SlurmResumeConfig(CONFIG_FILE_PATH)
+        _resume(args.nodes, resume_config)
+    except Exception as e:
```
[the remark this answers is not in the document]
demartinofra: Actually if we exit with error the ResumeFailProgram might be invoked that is what we want only if all nodes failed. If this is true we could just exit with error when all nodes failed and wait for the suspend to be invoked. We'll have to test this

src/common/schedulers/slurm_commands.py
```diff
@@ -42,6 +43,89 @@
     "cpus-per-tres",
 ]
 SQUEUE_FIELD_STRING = ",".join([field + ":{size}" for field in _SQUEUE_FIELDS]).format(size=SQUEUE_FIELD_SIZE)
+SCONTROL = "/opt/slurm/bin/scontrol"
+SlurmNode = collections.namedtuple("SlurmNode", ["name", "nodeaddr", "nodehostname", "state"])
+
+
+def update_nodes(nodes, nodeaddr=None, nodehostname=None, state=None, reason=None, raise_on_error=True):
```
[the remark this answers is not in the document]
rexcsn: This might be difficult to check if nodes are in nodelist format?

src/slurm_cloud_bursting/resume.py
```diff
@@ -0,0 +1,262 @@
...
+from retrying import retry
+
+from common.schedulers.slurm_commands import get_nodes_info, set_nodes_down, set_nodes_power_down, update_nodes
+from common.utils import grouper
+from slurm_cloud_bursting.utils import CONFIG_FILE_PATH
+
+TAG_SPECIFICATIONS = [
+    {
+        "ResourceType": "instance",
+        "Tags": [{"Key": "aws-parallelcluster-slurm-node-type", "Value": "cloud-bursting-compute"}],
```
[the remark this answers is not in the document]
rexcsn: I think we would, or we are ok with just relying on the Name tag?

src/slurm_cloud_bursting/resume.py
```diff
@@ -0,0 +1,262 @@
...
+        except IOError:
+            log.error(f"Cannot read slurm cloud bursting scripts configuration file: {config_file_path}")
+            raise
+
+        self.region = config.get("slurm_cb_config", "region")
+        self.cluster_name = config.get("slurm_cb_config", "cluster_name")
+        self.max_batch_size = int(
+            config.get("slurm_cb_config", "max_batch_size", fallback=self.DEFAULT_MAX_INSTANCES_BATCH_SIZE)
+        )
+        self.update_node_address = config.getboolean("slurm_cb_config", "update_node_address", fallback=True)
```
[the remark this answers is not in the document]
rexcsn: Good point

src/slurm_cloud_bursting/resume.py
rexcsn: We are calling this function in batches

src/slurm_cloud_bursting/resume.py
rexcsn: Ah ok

src/slurm_cloud_bursting/resume.py
```diff
@@ -0,0 +1,262 @@
...
+    )
+    return [
+        (_parse_ec2_instance(ec2_instance), slurm_node)
+        for ec2_instance, slurm_node in zip(result["Instances"], batch_nodes)
+    ]
+
+
+def _add_instances(node_list, resume_config):
+    """Launch EC2 instances for cloud nodes."""
+    logging.debug(resume_config.region)
```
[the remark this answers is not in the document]
rexcsn: ok

src/slurm_cloud_bursting/resume.py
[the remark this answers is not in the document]
rexcsn: ok

src/slurm_cloud_bursting/suspend.py
rexcsn: Smaller batch size more tolerant?

src/slurm_cloud_bursting/suspend.py
rexcsn: ok

src/slurm_cloud_bursting/resume.py
rexcsn: Verify how slurm handle failure

src/slurm_cloud_bursting/resume.py
rexcsn: Done

src/slurm_cloud_bursting/resume.py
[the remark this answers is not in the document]
rexcsn: Manually check and `exit 1` does not trigger `ResumeFailProgram` at least not immediately. Slurm knows the program failed by the node is still stuck in `power_on` state. Keeping existing behavior

tox.ini
```diff
@@ -14,7 +14,8 @@ commands =
     bash ./tests/test.sh
     python -m unittest discover -s tests/jobwatcher -p "*tests.py"
     py.test -l -v --basetemp={envtmpdir} --cov={envsitepackagesdir}/jobwatcher --cov={envsitepackagesdir}/sqswatcher \
-        --cov={envsitepackagesdir}/nodewatcher --cov={envsitepackagesdir}/common --cov-report=term tests/
+        --cov={envsitepackagesdir}/nodewatcher --cov={envsitepackagesdir}/slurm_cloud_bursting \
```
demartinofra, line 17: I think in the renaming you forgot to rename this

demartinofra: ok I fixed it

PR #220, opened 2020-07-21
Refactor Resume/Suspend functionalities into common module

rexcsn:
* Refactor SlurmNode into a class with added functionalities in anticipation for future work
* Refactor most of resume program functionality into InstanceLauncher in slurm_plugin/common.py
* Move suspend program functionality into slurm_plugin/common.py
* Use regex to capture information from nodename, changed nodename validator to better account for new instance type names
* Add/Refactor unit test to cover the above changes


tox.ini
```diff
@@ -52,11 +52,14 @@ commands =
...
 # The following target sorts the import according to .isort.cfg file.
+# Pin isort to the last release prior to 5.0.0, which resulted in import
+# statements being flagged as improperly sorted where previously they were
+# not. See https://timothycrosley.github.io/isort/docs/major_releases/introducing_isort_5/.
 [testenv:isort]
 basepython = python3
 skip_install = true
 deps =
-    isort
+    isort==4.3.21
```
demartinofra, line 62: Shall we apply changes introduced by isort 5 also in wip/hit branch? We might just commit those in a separate commit

src/slurm_plugin/common.py
```diff
@@ -8,5 +8,183 @@
...
+    """
+    InstanceLauncher class.
+
+    Class to manage instance launching action.
+    Should generally be used when launching instances for slurm integration.
+    """
+
+    def __init__(self, node_list, region, cluster_name, boto3_config, max_batch_size, update_node_address):
+        """Initialize InstanceLauncher with required attributes."""
+        self.node_list = node_list
```
demartinofra, line 77: let's stick to python naming convention by adding a leading underscore to class variables that are not part of the public class interface

src/slurm_plugin/common.py
```diff
@@ -8,5 +8,183 @@
...
+            MaxCount=current_batch_size,
+            # LaunchTemplate is different for every instance type in every queue
+            # LaunchTemplate name format: {cluster_name}-{queue_name}-{instance_type}
+            # Sample LT name: hit-queue1-c5.xlarge
+            LaunchTemplate={"LaunchTemplateName": f"{cluster_name}-{queue}-{instance_type}"},
+        )
+
+        return result["Instances"]
+
+    def _capture_nodename(self, nodename):
```
demartinofra, line 178: `_parse_nodename`?

src/slurm_plugin/common.py
```diff
@@ -8,5 +8,183 @@
...
+    log.info("Terminating instances %s", instance_ids_to_terminate)
+    for instances in grouper(instance_ids_to_terminate, batch_size):
+        try:
+            # Boto3 clients retries on connection errors only
+            ec2_client.terminate_instances(InstanceIds=list(instances),)
+        except ClientError as e:
+            log.error("Failed when terminating instances %s with error %s", instances, e)
+
+
+class InstanceLauncher:
```
demartinofra, line 67: What I don't like about this class is that you can use it only once, meaning you cannot initialize an InstanceLauncher object and then call launch_instances (or add_instances_for_nodes) multiple times. Can we slightly change it so that we keep at the class level the data that would not change across multiple invocations (e.g. region, cluster_name..) and move attributes such as node_list to the actual method that gets called to launch new instances. Also failed_nodes needs to be re-initialized.

Also would it make sense to make this more generic and include the termination functionalities defined above? Also the module name could be updated since here we are grouping all functions related to instance management if I'm not wrong

src/slurm_plugin/common.py
```diff
@@ -8,5 +8,183 @@
...
+        instance_hostnames = []
+        for instance in launched_instances:
+            instance_ids.append(instance["InstanceId"])
+            instance_ips.append(instance["PrivateIpAddress"])
+            instance_hostnames.append(instance["PrivateDnsName"].split(".")[0])
+        return instance_ids, instance_ips, instance_hostnames
+
+    def _parse_requested_instances(self):
+        """
+        Parse out instance type from slurm NodeName.
```
demartinofra, line 146: is this method really doing this? It is more counting how many instances of a certain type for each queue are requested.

src/slurm_plugin/common.py
```diff
@@ -8,5 +8,183 @@
...
+        """
+        instances_to_launch = collections.defaultdict(lambda: collections.defaultdict(list))
+        for node in self.node_list:
+            capture = self._capture_nodename(node)
+            if capture:
+                queue_name, instance_type = capture
+                instances_to_launch[queue_name][instance_type].append(node)
+            else:
+                log.warning("Discarding NodeName with invalid format: %s", node)
+        log.info("instances_to_launch = %s", instances_to_launch)
```
demartinofra, line 159: Let's make log messages more meaningful.

src/slurm_plugin/common.py
```diff
@@ -8,5 +8,183 @@
...
+
+        Valid NodeName format: {queue_name}-{static/dynamic}-{instance_type}-{number}
+        """
+        nodename_capture = re.match(r"^([a-z0-9\-_]+)-(static|dynamic)-([a-z0-9-]+.[a-z0-9-]+)-\d+$", nodename)
+        if not nodename_capture:
+            log.error("Invalid nodename format for node %s", nodename)
+            self.failed_nodes.append(nodename)
+            return None
+
+        return nodename_capture.group(1, 3)
```
demartinofra, line 190: it is not ideal to have functions that return a different number of values based on some conditions. Also having a simple function that needs to parse nodenames also update the global state of failed_nodes makes it more complicated and harder to test. Shall we simply raise a InvalidNodenameError when a parsing failure occurs and have the caller handle it?

src/slurm_plugin/resume.py
```diff
@@ -217,7 +111,7 @@ def _resume(arg_nodes, resume_config):
     node_list = [node.name for node in get_nodes_info(arg_nodes)]
     log.info("Retrieved nodelist: %s", node_list)
 
-    _add_instances(node_list, resume_config)
+    failed_nodes = _add_instances(node_list, resume_config)
     success_nodes = [node for node in node_list if node not in failed_nodes]
     log.info("Successfully launched nodes %s", success_nodes)
     if failed_nodes:
```
demartinofra, line 117: line below looks more like an error than an info message, right?

src/slurm_plugin/suspend.py
```diff
@@ -122,14 +89,14 @@ def _suspend(arg_nodes, suspend_config):
     slurm_nodes = get_nodes_info(arg_nodes)
     log.debug("Slurm_nodes = %s", slurm_nodes)
 
-    instance_ids_to_nodename = _get_instance_ids_to_nodename(
-        slurm_nodes, suspend_config.region, suspend_config.cluster_name, suspend_config.boto3_config
+    terminate_associated_instances(
```
demartinofra, line 92: let's make sure we log somewhere the instance ids of the instances we are trying to terminate and ideally the associated slurm node

tests/slurm_plugin/test_common.py
```diff
@@ -0,0 +1,610 @@
...
+
+from assertpy import assert_that
+from common.schedulers.slurm_commands import SlurmNode
+from slurm_plugin.common import InstanceLauncher, _get_instance_ids_to_nodename, delete_instances
+from tests.common import MockedBoto3Request
+
+
+@pytest.fixture()
+def boto3_stubber_path():
+    # we need to set the region in the environment because the Boto3ClientFactory requires it.
```
demartinofra, line 14: I think this does not apply in the node package. I think it comes from pcluster batch CLI tests

src/common/schedulers/slurm_commands.py
```diff
@@ -162,6 +228,28 @@ def get_nodes_info(nodes):
     return _parse_nodes_info(nodeinfo_str)
 
 
+def get_partition_info():
+    """
+    Retrieve SlurmNode list from slurm nodelist notation.
```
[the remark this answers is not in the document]
rexcsn: Done

src/slurm_plugin/common.py
rexcsn: Done

src/slurm_plugin/common.py
rexcsn: Tried

src/slurm_plugin/common.py
rexcsn: Done

src/slurm_plugin/common.py
rexcsn: Done

src/slurm_plugin/resume.py
rexcsn: Done

src/slurm_plugin/suspend.py
rexcsn: Added log entry in `terminate_associated_instances`

tox.ini
rexcsn: Sure, doing a separate commit

src/slurm_plugin/common.py
demartinofra: Regarding InstanceLauncher is not about performance but about limiting a class to be used only once without even enforcing this. I understand the reason why this class was created in the first place, which is mostly to have a shared state across the different methods, but I don't think this is something that prevents us from having the class behavior consistent across multiple invocations. If we don't want to do that, for the correctness of the class API we should at least raise an exception when the method to launch instance is invoked the second time. Personally I'd just reset the shared state when the method is invoked and also we could consider returning the list of failed nodes rather than reading it separately from the class attributes, making it an internal state only.

For the termination it was more for consistency but I'm fine leaving it as is.

rexcsn: Ok that makes sense, I will revise InstanceLauncher, and see if I can potentially include termination functions as well. If not I will just leave termination functions as is.

tests/slurm_plugin/test_common.py
rexcsn: This fixture is used by the boto3 stubber fixture to specify the path of the boto3 import to mock, so I think it is required.

demartinofra: the fixture is required. I was referring to the comment and the line below related to Boto3ClientFactory

src/slurm_plugin/common.py
```diff
@@ -8,5 +8,184 @@
...
+    def __init__(self, region, cluster_name, boto3_config):
+        """Initialize InstanceLauncher with required attributes."""
+        self._region = region
+        self._cluster_name = cluster_name
+        self._boto3_config = boto3_config
+        self.failed_nodes = []
+
+    def clear_failed_nodes(self):
+        """Clear and reset failed nodes list."""
+        self.failed_nodes = []
```
demartinofra, line 55: why not just doing this as first thing in `add_instances_for_nodes`, without exposing any additional function the caller needs to be aware of?

rexcsn: Done

tests/slurm_plugin/test_common.py
rexcsn: The region is actually required, or else we will get an error like this: https://travis-ci.org/github/aws/aws-parallelcluster-node/jobs/712808915

PR #222, opened 2020-07-29
Implement Cluster Management Daemon, clustermgtd

rexcsn:
* Cluster management daemon is used to sync scheduler states with EC2 states, and has the following functionalities:
* Clean up INACTIVE partitions
* Maintain any node with unhealthy EC2 states
* Maintain any node with unhealthy scheduler states
* Clean up orphaned resources not included in scheduler configuration
* Add unit tests to cover newly added functionalities

### DIFF
--- a/src/slurm_plugin/clustermgtd.py
+++ b/src/slurm_plugin/clustermgtd.py
@@ -243,6 +243,16 @@ class ClusterManager:
             timestamp_file.write(f"{self.current_time}")
 
     @staticmethod
+    @retry(stop_max_attempt_number=3, wait_fixed=1000)
+    def _get_node_info_with_retry(nodes):
+        return get_nodes_info(nodes, command_timeout=5)
+
+    @staticmethod
+    @retry(stop_max_attempt_number=3, wait_fixed=1000)
+    def _get_partition_info_with_retry():
+        return get_partition_info(command_timeout=5)
+
+    @staticmethod
     def _get_node_info_from_partition():
         """
         Retrieve node belonging in UP/INACTIVE partitions.
@@ -252,9 +262,9 @@ class ClusterManager:
         try:
             inactive_nodes = []
             active_nodes = []
-            partitions = get_partition_info()
+            partitions = ClusterManager._get_partition_info_with_retry()
             for part in partitions:
-                nodes = get_nodes_info(part.nodes)
+                nodes = ClusterManager._get_node_info_with_retry(part.nodes)
                 if "INACTIVE" in part.state:
                     inactive_nodes.extend(nodes)
                 else:
--- a/src/slurm_plugin/clustermgtd.py
+++ b/src/slurm_plugin/clustermgtd.py
@@ -28,7 +28,7 @@ from common.schedulers.slurm_commands import (
 )
 from common.time_utils import seconds
 from common.utils import sleep_remaining_loop_time
-from slurm_plugin.common import CONFIG_FILE_DIR, EC2_INSTANCE_HEALTHY_STATUSES, InstanceManager
+from slurm_plugin.common import CONFIG_FILE_DIR, EC2_HEALTH_STATUS_UNHEALTHY_STATES, InstanceManager
 
 LOOP_TIME = 30
 log = logging.getLogger(__name__)
@@ -148,8 +148,6 @@ class ClustermgtdConfig:
         self._get_launch_config(config)
         self._get_terminate_config(config)
 
-        # Initialize instance manager class that will be used to launch/terminate instance
-        self.instance_manager = InstanceManager(self.region, self.cluster_name, self.boto3_config)
         # Log configuration
         log.info(self.__repr__())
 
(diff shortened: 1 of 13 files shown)


src/slurm_plugin/common.py
```diff
@@ -180,14 +190,58 @@ def get_instance_ids_to_nodename(self, slurm_nodes):
...
         log.info("Terminating instances %s", instance_ids_to_terminate)
-        for instances in grouper(instance_ids_to_terminate, batch_size):
+        for instances in grouper(instance_ids_to_terminate, terminate_batch_size):
             try:
                 # Boto3 clients retries on connection errors only
                 ec2_client.terminate_instances(InstanceIds=list(instances),)
             except ClientError as e:
                 log.error("Failed when terminating instances %s with error %s", instances, e)
+
+    def get_instance_health_states(self, instance_ids):
```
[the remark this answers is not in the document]
rexcsn: Added unit test, failure are tested in `test_manage_cluster_boto3`

src/slurm_plugin/common.py
```diff
@@ -180,14 +190,58 @@ def get_instance_ids_to_nodename(self, slurm_nodes):
...
+                        instance.get("SystemStatus"),
+                        instance.get("Events"),
+                    )
+                    for instance in response
+                ]
+            )
+
+        return instance_health_states
+
+    def get_cluster_instances(self, include_master=False, alive_states_only=True):
```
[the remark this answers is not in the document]
rexcsn: Added unit test, failures are tested in `test_manage_cluster_boto3`

src/slurm_plugin/clustermgtd.py
```diff
@@ -0,0 +1,552 @@
...
+        try:
+            if (
+                # Check instance status
+                instance_health_state.instance_status.get("Status") not in EC2_INSTANCE_HEALTHY_STATUSES
+                and ClusterManager._time_is_up(
+                    instance_health_state.instance_status.get("Details")[0].get("ImpairedSince"),
+                    self.current_time,
+                    self.sync_config.health_check_timeout,
+                )
+            ) or (
```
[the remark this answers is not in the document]
rexcsn: We cannot do this because the `ImpairedSince ` time for instance status and system status are different

src/slurm_plugin/clustermgtd.py
```diff
@@ -0,0 +1,624 @@
...
+        Call is made by filtering on tags and includes non-terminating instances only
+        Instances returned will not contain instances previously terminated in _clean_up_inactive_partition
+        """
+        try:
+            return self.instance_manager.get_cluster_instances(include_master=False, alive_states_only=True)
+        except Exception as e:
+            log.error("Unable to retrieve instances from EC2 with exception: %s", e)
+            raise
+
+    def _perform_health_check_actions(self, cluster_instances, ip_to_slurm_node_map):
```
demartinofra, line 282: For each of these functions called from cluster_manager shall we add a small informative logging message that tells us the operation that is in progress. This could help following the flow of clustermgtd in case of errors

src/slurm_plugin/clustermgtd.py
```diff
@@ -0,0 +1,624 @@
...
+            if not self.sync_config.disable_scheduled_event_health_check:
+                self._handle_health_check(
+                    instance_health_states,
+                    id_to_instance_map,
+                    ip_to_slurm_node_map,
+                    health_check_type=ClusterManager.HealthCheckTypes.scheduled_event,
+                )
+        except Exception as e:
+            log.error(
+                "Unable to perform instance health check actions with exception: %s\nContinuing other operations...", e
```
demartinofra, line 306: one of the downsides of catching the exception only here is that it's hard for us to understand where it generated from (we don't have a stack trace). If we want to add more granular logging we could use a decorator like the following (probably it's incomplete but it's enough to give the idea):

```
def log_exception(logger, desc, log_level=logging.ERROR, exception=Exception):
    def _log_exception(function):
        @functools.wraps(function)
        def wrapper(*args, **kwargs):
            try:
                return function(*args, **kwargs)
            except exception as e:
                logger.log(log_level, "Failed when %s with exception %s", desc, e)
                raise
        return wrapper

    return _log_exception


@log_exception(logger, "retrieving EC2 data")
def my function(...):
  ...
```

src/slurm_plugin/clustermgtd.py
```diff
@@ -0,0 +1,624 @@
...
+    def _update_static_nodes_in_replacement(self, slurm_nodes):
+        """Update self.static_nodes_in_replacement by removing nodes that finished replacement and is up."""
+        nodename_to_slurm_nodes_map = {node.name: node for node in slurm_nodes}
+        nodes_not_in_replacement = set()
+        for nodename in self.static_nodes_in_replacement:
+            node = nodename_to_slurm_nodes_map.get(nodename)
+            # Remove nodename from static_nodes_in_replacement if node is no longer an active node or node is up
+            if not node or node.is_up():
+                nodes_not_in_replacement.add(nodename)
+        self.static_nodes_in_replacement.difference_update(nodes_not_in_replacement)
```
demartinofra, line 394: minor: instead of creating a new list of nodes to remove and then compute the difference you could have computed the update list directly (by using the complementary condition `if node and not node.is_up()`) and then assign the new list.

src/slurm_plugin/clustermgtd.py
```diff
@@ -0,0 +1,624 @@
...
+                instances_to_terminate.append(backing_instance.id)
+        log.info(
+            "Terminating instances %s that are backing unhealthy static nodes: %s",
+            instances_to_terminate,
+            unhealthy_static_nodes,
+        )
+        self.instance_manager.delete_instances(
+            instances_to_terminate, terminate_batch_size=self.sync_config.terminate_max_batch_size
+        )
+        log.info("Launching new instances for unhealthy static nodes: %s", unhealthy_static_nodes)
```
demartinofra, line 518: I'm leaving it here but as a general comment we should review the logs cause I'm afraid they are really verbose and significantly grow with the cluster size

src/slurm_plugin/clustermgtd.py
```diff
@@ -0,0 +1,624 @@
...
+            log.info(
+                "Found the following unhealthy static nodes: %s\nFound the following unhealthy dynamic nodes: %s",
+                unhealthy_static_nodes,
+                unhealthy_dynamic_nodes,
+            )
+            if unhealthy_dynamic_nodes:
+                ClusterManager._handle_unhealthy_dynamic_nodes(unhealthy_dynamic_nodes)
+            if unhealthy_static_nodes:
+                self._handle_unhealthy_static_nodes(unhealthy_static_nodes, private_ip_to_instance_map)
+        except Exception as e:
```
demartinofra, line 552: do we want that a failure in _handle_unhealthy_dynamic_nodes also affects _handle_unhealthy_static_nodes?

src/slurm_plugin/clustermgtd.py
```diff
@@ -0,0 +1,624 @@
...
+    def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
+        """Terminate instance not associated with any node and running longer than orphaned_instance_timeout."""
+        try:
+            instances_to_terminate = []
+            for instance in cluster_instances:
+                if instance.private_ip not in ips_used_by_slurm and ClusterManager._time_is_up(
+                    instance.launch_time, self.current_time, self.sync_config.orphaned_instance_timeout
+                ):
+                    instances_to_terminate.append(instance.id)
+            log.info("Terminating the following orphaned instances: %s", instances_to_terminate)
```
demartinofra, line 566: why not moving this after the if below?

src/slurm_plugin/clustermgtd.py
```diff
@@ -0,0 +1,624 @@
...
+        self.instance_manager = InstanceManager(
+            self.sync_config.region, self.sync_config.cluster_name, self.sync_config.boto3_config
+        )
+
+    def manage_cluster(self, sync_config):
+        """Manage cluster by syncing scheduler states with EC2 states and performing node maintenance actions."""
+        # Initialization
+        self.set_sync_config(sync_config)
+        self.set_current_time(datetime.now(tz=timezone.utc))
+        self.initialize_instance_manager()
```
demartinofra, line 197: sorry if I asked already. Can this be initialized in the constructor?

src/slurm_plugin/clustermgtd.py
```diff
@@ -0,0 +1,624 @@
...
+                nodes = get_nodes_info(part.nodes)
+                if "INACTIVE" in part.state:
+                    inactive_nodes.extend(nodes)
+                else:
+                    active_nodes.extend(nodes)
+
+            return active_nodes, inactive_nodes
+        except Exception as e:
+            log.error("Unable to retrieve node states from scheduler with exception: %s", e)
+            raise
```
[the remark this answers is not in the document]
rexcsn: Ok, I see. For some reason my impression was that we want clustermgtd to crash on critical failures, if say slurmctld is down, but I can definitely do that, so the clustermgtd is always running and can handle failure when getting scheduler states or ec2 states

src/slurm_plugin/clustermgtd.py
rexcsn: Sure

src/slurm_plugin/clustermgtd.py
rexcsn: Will look through this

src/slurm_plugin/clustermgtd.py
rexcsn: Good catch

src/slurm_plugin/clustermgtd.py
rexcsn: So we know this function is called?

src/slurm_plugin/common.py
```diff
@@ -180,14 +193,57 @@ def get_instance_ids_to_nodename(self, slurm_nodes):
...
                 ec2_client.terminate_instances(InstanceIds=list(instances),)
             except ClientError as e:
                 log.error("Failed when terminating instances %s with error %s", instances, e)
+
+    def get_instance_health_states(self, instance_ids):
+        """Get health status for instances."""
+        instance_health_states = []
+        ec2_client = boto3.client("ec2", region_name=self._region, config=self._boto3_config)
+        for batch in grouper(instance_ids, self.GET_HEALTH_STATES_BATCH_SIZE):
+            response = ec2_client.describe_instance_status(InstanceIds=list(batch)).get("InstanceStatuses")
```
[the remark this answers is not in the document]
rexcsn: Good point!

src/slurm_plugin/common.py
```diff
@@ -180,14 +193,57 @@ def get_instance_ids_to_nodename(self, slurm_nodes):
...
+        ec2_client = boto3.client("ec2", region_name=self._region, config=self._boto3_config)
+        paginator = ec2_client.get_paginator("describe_instances")
+        args = {
+            "Filters": [{"Name": "tag:ClusterName", "Values": [self._cluster_name]}],
+        }
+        if alive_states_only:
+            args["Filters"].append({"Name": "instance-state-name", "Values": list(EC2_INSTANCE_ALIVE_STATES)})
+        if not include_master:
+            args["Filters"].append({"Name": "tag:aws-parallelcluster-node-type", "Values": ["Compute"]})
+        response_iterator = paginator.paginate(**args)
```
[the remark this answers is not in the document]
rexcsn: Let me check

src/slurm_plugin/clustermgtd.py
demartinofra: gotcha, then probably the setter for the sync_config should trigger a re-initialzation of the instance_manager.

src/slurm_plugin/common.py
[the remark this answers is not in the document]
demartinofra: Thanks for checking! I'd rather have a big number we are aware of rather than relying on an unclear page size

PR #226, opened 2020-08-04
Implement computemgtd

rexcsn:
* Computemgtd will self-terminate instance if clustermgtd is offline for more than clustermgtd_timeout
* Revised clustermgtd implementation slightly
* Add unit tests for computemgtd

### DIFF
--- a/src/slurm_plugin/common.py
+++ b/src/slurm_plugin/common.py
@@ -13,6 +13,7 @@ import functools
 import logging
 import re
 import subprocess
+from datetime import timezone
 
 import boto3
 from botocore.exceptions import ClientError
@@ -41,6 +42,9 @@ EC2_SCHEDULED_EVENT_CODES = [
 # PageSize parameter used for Boto3 paginated calls
 # Corresponds to MaxResults in describe_instances and describe_instance_status API
 BOTO3_PAGINATION_PAGE_SIZE = 1000
+# timestamp used by clustermgtd and computemgtd should be in default ISO format
+# YYYY-MM-DDTHH:MM:SS.ffffff+HH:MM[:SS[.ffffff]]
+TIMESTAMP_FORMAT = "%Y-%m-%d %H:%M:%S.%f%z"
 
 log = logging.getLogger(__name__)
 
@@ -78,9 +82,6 @@ class InstanceManager:
     Class implementing instance management actions.
     Used when launching instance, terminating instance, and retrieving instance info for slurm integration.
     """
-
-    # Max 100 instance ids in 1 describe_instance_status call
-    GET_HEALTH_STATES_BATCH_SIZE = 100
 
     class InvalidNodenameError(ValueError):
         r"""
@@ -308,3 +309,18 @@ class InstanceManager:
             )
             for instance_info in filtered_iterator
         ]
+
+
+def time_is_up(initial_time, current_time, grace_time):
+    """Check if timeout is exceeded."""
+    # Localize datetime objects to UTC if not previously localized
+    # All timestamps used in this function should be already localized
+    # Assume timestamp was taken from UTC is there is no localization info
+    if not initial_time.tzinfo:
+        log.warning("Timestamp %s is not localized. Please double check that this is expected, localizing to UTC.")
+        initial_time = initial_time.replace(tzinfo=timezone.utc)
+    if not current_time.tzinfo:
+        log.warning("Timestamp %s is not localized. Please double check that this is expected, localizing to UTC")
+        current_time = current_time.replace(tzinfo=timezone.utc)
+    time_diff = (current_time - initial_time).total_seconds()
+    return time_diff >= grace_time
--- a/src/slurm_plugin/computemgtd.py
+++ b/src/slurm_plugin/computemgtd.py
@@ -14,8 +14,8 @@ from common.utils import check_command_output, sleep_remaining_loop_time
 from slurm_plugin.common import CONFIG_FILE_DIR, TIMESTAMP_FORMAT, InstanceManager, log_exception, time_is_up
 
 LOOP_TIME = 60
-# Computemgtd config is under /opt/slurm/etc/pcluster; all compute nodes share a config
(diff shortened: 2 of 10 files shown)


src/slurm_plugin/common.py
```diff
@@ -64,7 +68,7 @@ def _clear_failed_nodes(self):
         """Clear and reset failed nodes list."""
         self.failed_nodes = []
 
-    def add_instances_for_nodes(self, node_list, launch_batch_size, update_node_address=True):
+    def add_instances_for_nodes(self, node_list, launch_batch_size=100, update_node_address=True):
```
demartinofra, line 71: I'd avoid setting these defaults here (same for the functions below) so that we always force the caller to specify the value. Default can be set in the caller and read from config.

src/slurm_plugin/computemgtd.py
```diff
@@ -0,0 +1,138 @@
...
+
+from botocore.config import Config
+from configparser import ConfigParser
+from retrying import retry
+
+from common.time_utils import seconds
+from common.utils import check_command_output, sleep_remaining_loop_time
+from slurm_plugin.common import CONFIG_FILE_DIR, TIMESTAMP_FORMAT, InstanceManager, time_is_up
+
+LOOP_TIME = 30
```
demartinofra, line 14: let's increase this to 1 min

src/slurm_plugin/computemgtd.py
```diff
@@ -0,0 +1,138 @@
...
+            )
+        # Check heartbeat
+        try:
+            last_heartbeat = _get_clustermgtd_heartbeat(computemgtd_config.clustermgtd_heartbeat_file_path)
+            log.info("Latest heartbeat from clustermgtd: %s", last_heartbeat)
+        except Exception as e:
+            log.error("Unable to retrieve clustermgtd heartbeat with exception: %s", e)
+        finally:
+            # Any non-positive clustermgtd_timeout will disable self-termination
+            if computemgtd_config.clustermgtd_timeout > 0 and time_is_up(
```
demartinofra, line 113: why don't we use a separate parameter to disable computemgtd functionalities and we check for it at the beginning of the loop so that we don't perform any other operation?

src/slurm_plugin/computemgtd.py
```diff
@@ -0,0 +1,138 @@
...
+                last_heartbeat, current_time, computemgtd_config.clustermgtd_timeout
+            ):
+                log.error(
+                    "Clustermgtd has been offline since %s. Current time is %s. Timeout of %s seconds has expired!",
+                    last_heartbeat,
+                    current_time,
+                    computemgtd_config.clustermgtd_timeout,
+                )
+                _self_terminate(computemgtd_config)
```
demartinofra, line 122: If I'm not wrong here we decided to have 2 timeouts/checks:
1. we check for the hearthbeat coming from clustermgtd
2. when the hearthbeat check fails I'd say we check for the status of the node in the scheduler. This because we do not want to terminate a node still connected to the scheduler even if the clustermgtd is temporarily offline. When checking the status of the scheduler we could issue a `scontrol show node nodename` where the nodename could be read from the computemgtd config and written in the chef recipes.

Also probably we want 2 timeouts:
1. on the hearth-beat not received in the last 5 mins (3 I think it's a bit too little in case clustermgtd is very busy)
2. on slurm failing to respond (we might want to retry here for a minute or so but with a very low rate, e.g. every 20 secs for 3 times)

demartinofra: we need to handle failures in self_terminate by probably logging the error and having it retry at the next iteration rather than propagating the exception to the main program

src/slurm_plugin/computemgtd.py
rexcsn: Done

src/slurm_plugin/computemgtd.py
rexcsn: I will argue against the second slurm timeout, `SlurmdTimeout` should already take that into account, and I don't think we want to add a second timeout. Plus, timing on scheduler state is not straightforward, we will need another "pre-termination" loop to constantly check scheduler state(like nodewatcher), which is not that nice.

I think we should check the scheduler state only once whenever we see that the heartbeat is expired, and if the state is `down` or if we cannot retrieve scheduler state we will terminate.

demartinofra: yes sorry I should have been more specific. The second timeout was referred to the fact that Slurm might be temporary unresponsive so we want a big enough timeout or at least retry for some time if the scheduler command fails (for example with a timeout). we cannot be super aggressive and terminate right away if get_nodes_info fails because of a timeout. We should give some time to recover as we do today in nodewatcher. If you recall we had to increase the timeout in nodewatcher to account for transient issues that make the scheduler unresponsive (e.g. network overload or scheduler overload)

src/slurm_plugin/computemgtd.py
```diff
@@ -0,0 +1,189 @@
...
+def _is_self_node_down(slurm_nodename_file):
+    """
+    Check if self node is down in slurm.
+
+    This check prevents termination of a node that is still well-attached to the scheduler.
+    Note: a node that is not attached to the scheduler will be put into DOWN* automatically after slurmd.
+    """
+    try:
+        log.info("Reading self nodename from %s", slurm_nodename_file)
+        with open(slurm_nodename_file, "r") as nodename_file:
```
demartinofra, line 119: why don't we read this when we load the config?

src/slurm_plugin/computemgtd.py
```diff
@@ -0,0 +1,189 @@
...
+                e,
+            )
+        # Check heartbeat
+        try:
+            last_heartbeat = _get_clustermgtd_heartbeat(computemgtd_config.clustermgtd_heartbeat_file_path)
+            log.info("Latest heartbeat from clustermgtd: %s", last_heartbeat)
+        except Exception as e:
+            log.error("Unable to retrieve clustermgtd heartbeat with exception: %s", e)
+        finally:
+            if computemgtd_config.disable_computemgtd_actions:
```
demartinofra, line 170: minor: I'd just move this before checking the heartbeat

PR #231, opened 2020-08-20
Scaling Optimization

rexcsn:
* Optimize scaling by minimizing API calls and adjusting timeouts in slurm_plugin package
* Adjust scaling architecture, overall strategy is to have clustermgtd handle all termination and replacement, and make Resume/SuspendProgram execute quickly
* SuspendProgram will not terminate instance, but only resume nodes
* ResumeProgram only places nodes failed to launch in DOWN, but does not place nodes in POWER_DOWN
* Modify logging messages so they are easier to read in large scale settings
* Update unit tests
* Use /etc/parallelcluster/slurm_plugin as directory for configs instead of /opt/parallelcluster/...


src/slurm_plugin/suspend.py
```diff
@@ -114,16 +70,17 @@ def main():
...
                 suspend_config.logging_config,
                 default_log_file,
                 e,
             )
-        _suspend(args.nodes, suspend_config)
-    except Exception as e:
-        log.exception("Encountered exception when suspending instances for %s: %s", args.nodes, e)
-    finally:
+        log.info(
+            "Setting following nodes into DOWN so that clustermgtd can terminate associated instances: %s", args.nodes
```
demartinofra, line 79: aren't we setting them to IDLE?

src/common/schedulers/slurm_commands.py
```diff
@@ -159,14 +161,14 @@ def update_nodes(
...
-        run_command(f"{update_cmd} {node_info}", raise_on_error=raise_on_error, timeout=command_timeout)
+        run_command(f"{update_cmd} {node_info}", raise_on_error=raise_on_error, timeout=command_timeout, shell=True)
 
 
 def update_partitions(partitions, state):
     succeeded_partitions = []
     for partition in partitions:
         try:
-            run_command(f"{SCONTROL} update partitionname={partition} state={state}", raise_on_error=True)
+            run_command(f"{SCONTROL} update partition name={partition} state={state}", raise_on_error=True, shell=True)
```
demartinofra, line 171: I think the previous version was the correct one

PR #234, opened 2020-08-26
Add DNS conf params to clustermgtd conf and add possibility to use private hostnames

enrico-usai:
The clustermgtd daemon starts the static instances so it is required to have all the DNS info.

We're also setting use_private_hostname configuration parameters to both clustermgtd and resume conf to have a way to force the
use of private hostnames in place of the custom ones. 

Signed-off-by: Enrico Usai <usai@amazon.com>

### DIFF
--- a/src/slurm_plugin/clustermgtd.py
+++ b/src/slurm_plugin/clustermgtd.py
@@ -137,6 +137,10 @@ class ClustermgtdConfig:
         "disable_scheduled_event_health_check": False,
         "disable_all_cluster_management": False,
         "health_check_timeout": 180,
+        # DNS domain configs
+        "hosted_zone": None,
+        "dns_domain": None,
+        "use_private_hostname": False,
     }
 
     def __init__(self, config_file_path):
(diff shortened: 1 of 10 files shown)


src/slurm_plugin/clustermgtd.py
```diff
@@ -137,6 +137,11 @@ class ClustermgtdConfig:
         "disable_scheduled_event_health_check": False,
         "disable_all_cluster_management": False,
         "health_check_timeout": 180,
+        # DNS domain configs
+        "dynamodb_table": None,
+        "hosted_zone": None,
+        "dns_domain": None,
+        "use_private_hostname": "no",
```
demartinofra, line 144: let's translate this in the cookbook to true/false

src/slurm_plugin/clustermgtd.py
```diff
@@ -137,6 +137,11 @@ class ClustermgtdConfig:
         "disable_scheduled_event_health_check": False,
         "disable_all_cluster_management": False,
         "health_check_timeout": 180,
+        # DNS domain configs
+        "dynamodb_table": None,
```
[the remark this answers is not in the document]
enrico-usai: Done

src/slurm_plugin/clustermgtd.py
enrico-usai: Changed, we're now using True/False in the node code, and "true"/"false" in the cookbook one.

I tested it with `"use_private_hostname": "true"`  and `"use_private_hostname": True` and with the default value.

demartinofra (approved):

PR #236, opened 2020-08-27
Use st/dy in place of static/dynamic as node type identifier

enrico-usai:
## Use st/dy in place of static/dynamic as node type identifier

Added extra tests to the parse_nodename function

## Use parse_nodename to evaluate if a node is a static one

The test we were using `'-st-' in self.name` was not safe.
Now we're parsing the slurm nodename when initializing the `SlurmNode` object and saving the info in the `is_static` variable.

The same methods will be used by clustermgtd and instance manager.

Added tests for `is_static_node`

### DIFF
--- a/src/common/schedulers/slurm_commands.py
+++ b/src/common/schedulers/slurm_commands.py
@@ -71,17 +71,10 @@ class SlurmNode:
     def __init__(self, name, nodeaddr, nodehostname, state):
         """Initialize slurm node with attributes."""
         self.name = name
+        self.is_static = is_static_node(name)
         self.nodeaddr = nodeaddr
         self.nodehostname = nodehostname
         self.state = state
-
-    def is_static_node(self):
-        """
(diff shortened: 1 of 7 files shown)


src/common/schedulers/slurm_commands.py
```diff
@@ -278,7 +278,7 @@ def get_nodes_info(nodes, command_timeout=5):
     """
     Retrieve SlurmNode list from slurm nodelist notation.
 
-    Sample slurm nodelist notation: queue1-dynamic-c5_xlarge-[1-3],queue2-static-t2_micro-5.
+    Sample slurm nodelist notation: queue1-dyc5_xlarge-[1-3],queue2-st-t2_micro-5.
```
lukeseawalker, line 281: dyc5_xlarge -> dy-c5_xlarge

demartinofra (approved):

PR #239, opened 2020-08-31
Fix to clustermgtd

rexcsn:
* Change node_replacement_timeout=3600 to give more time for static nodes to spin up in case of long pre/post-install. This value should always be the same as ResumeTimeout
* Minor fix to logging
* Related to https://github.com/aws/aws-parallelcluster-cookbook/pull/710

### DIFF
--- a/src/common/schedulers/slurm_commands.py
+++ b/src/common/schedulers/slurm_commands.py
@@ -333,26 +333,22 @@ def _parse_partition_name_and_state(partition_info):
 
 def _get_partition_nodes(partition_name, command_timeout=5):
     """Get up nodes in a parition by querying sinfo, and filtering out power_down nodes."""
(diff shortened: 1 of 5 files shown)


demartinofra (commented):
LGTM.

Can you also do the following since `_handle_failed_nodes` also accepts string and not only lists
```
@@ -103,10 +103,10 @@ def _handle_failed_nodes(node_list):
     Clustermgtd will be responsible for running full DOWN -> POWER_DOWN process.
     """
     try:
-        log.info("Setting following failed nodes into DOWN state: %s", print_with_count(node_list))
+        log.info("Setting following failed nodes into DOWN state: %s", node_list)
         set_nodes_down(node_list, reason="Failure when resuming nodes")
     except Exception as e:
-        log.error("Failed to place nodes %s into down with exception: %s", print_with_count(node_list), e)
+        log.error("Failed to place nodes %s into down with exception: %s", node_list, e)
```

demartinofra (approved):

PR #244, opened 2020-09-04
Reload computemgtd config every 10 iterations

demartinofra:


src/slurm_plugin/computemgtd.py
```diff
@@ -26,6 +26,7 @@
 from slurm_plugin.common import CONFIG_FILE_DIR, TIMESTAMP_FORMAT, InstanceManager, log_exception, time_is_up
 
 LOOP_TIME = 60
+RELOAD_CONFIG_ITERATIONS = 10
```
rexcsn, line 29: Maybe make this smaller so that time between reloading is small than clustermgtd_timeout? Otherwise it's impossible to change clustermgtd_timeout and "save" a node

rexcsn (approved):

src/slurm_plugin/computemgtd.py
demartinofra: For now I think it's ok. New nodes will pull the new config.

PR #255, opened 2020-09-15
Upload cluster metrics to CloudWatch for slurm

agobeaux:
With this commit, cluster metrics are uploaded to CloudWatch. In order not to impact the job of the current clustermgtd, a separate thread was created to upload those metrics. Threads were used instead of processes as the functions here are I/O bound, threads are thus suitable. Some code has been added to the current deamon so that it saves the information it requests from the scheduler or from ec2. This information is then used by the second thread to upload the cluster metrics.
Testing: still to do, just modified so that it passed current tests but test coverage is less good as the new functions are not tested (and some code has been added in the old deamon)
Handling exception: working on it as exceptions from a sub-thread is not raised to the thread.

Signed-off-by: Alexandre Gobeaux <gobeaa@amazon.com>

### DIFF
--- a/src/slurm_plugin/clustermgtd.py
+++ b/src/slurm_plugin/clustermgtd.py
@@ -52,6 +52,14 @@ LOOP_TIME = 60
 log = logging.getLogger(__name__)
 
 
+class ThreadException(Exception):
+    """
+    Raised if one of the sub-threads created by main raises an exception
+    """
+
+    pass
+
+
 class ComputeFleetStatus(Enum):
     """Represents the status of the cluster compute fleet."""
 
@@ -380,7 +388,7 @@ class ClusterManager:
                 self._compute_fleet_status,
             )
 
-    def manage_cluster(self, cloudwatch_metrics):
+    def manage_cluster(self, cluster_metrics):
         """Manage cluster by syncing scheduler states with EC2 states and performing node maintenance actions."""
         # Initialization
         log.info("Managing cluster...")
@@ -396,8 +404,8 @@ class ClusterManager:
             try:
                 log.info("Retrieving nodes info from the scheduler")
                 active_nodes, inactive_nodes = self._get_node_info_from_partition()
-                cloudwatch_metrics._slurm_active_nodes = active_nodes
-                cloudwatch_metrics._slurm_inactive_nodes = inactive_nodes
+                cluster_metrics.slurm_active_nodes = active_nodes
+                cluster_metrics.slurm_inactive_nodes = inactive_nodes
             except ClusterManager.SchedulerUnavailable:
                 log.error("Unable to get partition/node info from slurm, no other action can be performed. Sleeping...")
                 return
@@ -407,7 +415,7 @@ class ClusterManager:
                 time.sleep(5)
                 log.info("Retrieving list of EC2 instances associated with the cluster")
                 cluster_instances = self._get_ec2_instances()
-                cloudwatch_metrics._ec2_nodes = cluster_instances
+                cluster_metrics.ec2_nodes = cluster_instances
             except ClusterManager.EC2InstancesInfoUnavailable:
                 log.error("Unable to get instances info from EC2, no other action can be performed. Sleeping...")
                 return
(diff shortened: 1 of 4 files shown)


src/slurm_plugin/clustermgtd.py
```diff
@@ -395,8 +396,8 @@ def manage_cluster(self):
             try:
                 log.info("Retrieving nodes info from the scheduler")
                 active_nodes, inactive_nodes = self._get_node_info_from_partition()
-                log.debug("Current active slurm nodes in scheduler: %s", active_nodes)
-                log.debug("Current inactive slurm nodes in scheduler: %s", inactive_nodes)
+                cloudwatch_metrics._slurm_active_nodes = active_nodes
```
enrico-usai, line 399: if `_slurm_active_nodes` is a private attribute (underscore prefix) we should not access to it.
We should have data encapsulation.

Let's use python properties: see for example this [guide](https://www.python-course.eu/python3_properties.php) or the [official documentation](https://docs.python.org/3/library/functions.html?highlight=property#property).

We're already using it somewhere in the CLI (e.g. for the region).

src/slurm_plugin/clustermgtd.py
```diff
@@ -379,7 +380,7 @@ def _manage_compute_fleet_status_transitions(self):
                 self._compute_fleet_status,
             )
 
-    def manage_cluster(self):
+    def manage_cluster(self, cloudwatch_metrics):
```
enrico-usai, line 383: cloudwatch_metrics --> cluster_metrics.
The name of the parameter should not depend by the fact we'll store this info in cloudwatch. We can store them somewhere else.

src/slurm_plugin/clustermgtd.py
```diff
@@ -879,17 +881,16 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
             )
 
 
-def _run_clustermgtd():
+def _run_clustermgtd(clustermgtd_config_file, config, custom_metrics):
```
enrico-usai, line 884: custom_metrics --> cluster_metrics

src/slurm_plugin/clustermgtd.py
```diff
@@ -907,7 +908,69 @@ def _run_clustermgtd():
...
+
+class ClusterMetrics:
+    def __init__(self):
+        self._has_received = False
+        self._slurm_active_nodes = None
+        self._slurm_inactive_nodes = None
+        self._ec2_nodes = None
+        self._config = None
+
+    def get_config(self):
```
enrico-usai, line 924: instead of getter/setter let's use properties

src/slurm_plugin/clustermgtd.py
```diff
@@ -907,7 +908,69 @@ def _run_clustermgtd():
...
         # Manage cluster
-        cluster_manager.manage_cluster()
+        cluster_manager.manage_cluster(custom_metrics)
+        custom_metrics._has_received = True
+        sleep_remaining_loop_time(config.loop_time, start_time)
+
+
+class ClusterMetrics:
+    def __init__(self):
+        self._has_received = False
```
enrico-usai, line 918: Instead of having an attribute for each metric we could simply have a dict called `metrics`.

Here, we can define an empty dict and in the `manage_cluster` we can put metrics in the dict.

Then in the get_values we just need to cycle the metrics dict and push the metrics we have.
In this way we can get rid of the `_has_received` parameter and we can easily add code somewhere to save new metrics, without the need to change the ClusterMetrics class or the upload method.

Another OO alternative could be to define a Metric class, that could have name and value attribute and in the ClusterMetrics we can manage a list of Metric objects, but for our initial case I think a dict is enough because it also permit to easily update an already existing metric because we'll use the name as dict key.

src/slurm_plugin/clustermgtd.py
```diff
@@ -907,7 +908,69 @@ def _run_clustermgtd():
...
+                {"name": "slurm_number_active_nodes", "value": len(self._slurm_active_nodes)},
+                {"name": "slurm_number_inactive_nodes", "value": len(self._slurm_inactive_nodes)},
+                {"name": "number_ec2_nodes", "value": len(self._ec2_nodes)},
+            ]
+        return []
+
+
+def _upload_cluster_metrics(custom_metrics):
+
+    # FIXME 'StorageResolution' should not be specified I think, here just for faster tests
```
enrico-usai, line 939: fixme?

src/slurm_plugin/clustermgtd.py
```diff
@@ -907,7 +908,69 @@ def _run_clustermgtd():
...
+    log.info("beginning of collect and upload custom metrics")
```
enrico-usai, line 940: minor: the log message should start with a capitalized letter, to be clear it is the initial part of a message and not a broken part of a message.

minor 2: it's not collect upload, but just upload.

minor 3: upload where? I'd call it `_push_metrics_to_cloudwatch` or something more explicit.

enrico-usai: It should be simple to add an unit test for this function. You just need to mock the put_metric_data boto3 call.

src/slurm_plugin/clustermgtd.py
```diff
@@ -907,7 +908,69 @@ def _run_clustermgtd():
...
+    config = custom_metrics.get_config()
+    while True:
+        start_time = datetime.now(tz=timezone.utc)
+
+        # get most recent values ([] if no values were uploaded at all (should happen only the first time)
+        metrics_to_upload = custom_metrics.get_values()
```
enrico-usai, line 946: metrics_to_upload --> metrics ?

src/slurm_plugin/clustermgtd.py
```diff
@@ -879,46 +906,227 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+        subthreads = []
+
+        try:
+            _create_and_monitor_threads(subthreads, clustermgtd_config_file, cluster_metrics, exceptions_list)
+        except CriticalThreadException:
+            _terminate_threads(subthreads)
+
     except Exception as e:
-        log.exception("An unexpected error occurred: %s", e)
+        log.error("An unexpected error occurred: %s", e)
```
demartinofra, line 1129: in this case we want to log it as an exception since it's something unexpected and the stack trace is useful in this situation

src/slurm_plugin/clustermgtd.py
```diff
@@ -879,46 +906,227 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+            # performance problems (instance very busy?), then, it would make it even more difficult for the instance
+            # to do its work.
+            sleep_remaining_loop_time(config.loop_time, start_time)
+    except Exception as e:
+        exceptions_list.append(ClusterMetricsException(e))
+        return
+    log.info("Leaving _push_metrics_to_cloudwatch")
+
+
+class RestartableThread(Thread):
```
demartinofra, line 1028: have you considered configuring these threads with the `daemon=true` flag so that if the main program exits it does not hang but simply exits all daemon threads?

src/slurm_plugin/clustermgtd.py
agobeaux: At first it was a private attribute, forgot to change it back, thanks for the catch and the links.

src/slurm_plugin/clustermgtd.py
agobeaux: Done but changed to metrics_collector now.

src/slurm_plugin/clustermgtd.py
agobeaux: Done but same as above.

src/slurm_plugin/clustermgtd.py
agobeaux: Done for config.

src/slurm_plugin/clustermgtd.py
agobeaux: We now have a timestamp and check if it has not been too much time since we last got new metrics. We also have two dictionaries now (metrics and metrics_provider) that enable us to add new metrics in a flexible manner.

src/slurm_plugin/clustermgtd.py
agobeaux: Done. Removed the FIXME and StorageResolution: misunderstood its use.

src/slurm_plugin/clustermgtd.py
agobeaux: Minor: fixed.
Still TODO: testing

src/slurm_plugin/clustermgtd.py
agobeaux: Done.

src/slurm_plugin/clustermgtd.py
agobeaux: Fixed in the last force push, should work now even for children threads (changed the approach to log those with the trace)

src/slurm_plugin/clustermgtd.py
```diff
@@ -51,6 +52,24 @@
 log = logging.getLogger(__name__)
 
 
+class ClusterMgtdException(Exception):
```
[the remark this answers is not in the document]
agobeaux: We no longer have this kind of exception.

src/slurm_plugin/clustermgtd.py
```diff
@@ -51,6 +59,12 @@
 log = logging.getLogger(__name__)
 
 
+class CriticalThreadException(Exception):
+    """Raised if the main function failed when executing _create_and_monitor_threads."""
```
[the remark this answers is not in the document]
agobeaux: Done, understood. Changed the sentences according to what was already done for other exceptions in the project.

src/slurm_plugin/clustermgtd.py
```diff
@@ -144,6 +158,8 @@ class ClustermgtdConfig:
         "hosted_zone": None,
         "dns_domain": None,
         "use_private_hostname": False,
+        # Disable pushing cluster metrics to cloudwatch
```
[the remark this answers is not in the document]
agobeaux: Deleted

src/slurm_plugin/clustermgtd.py
```diff
@@ -144,6 +158,8 @@ class ClustermgtdConfig:
...
+        "disable_push_metrics_to_cloudwatch": False,
```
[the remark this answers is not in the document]
agobeaux: Done

src/slurm_plugin/clustermgtd.py
```diff
@@ -879,44 +907,235 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+                )
+            # Manage cluster
+            cluster_manager.manage_cluster(metrics_collector)
+            metrics_collector.timestamp = datetime.now(tz=timezone.utc)  # FIXME is it ok to call it again? I guess so
+            sleep_remaining_loop_time(config.loop_time, start_time)
+    except Exception as e:
+        failed_threads.append("scaling")
+        log.exception("An unexpected error occurred: %s", e)
+        return
+    log.info("Leaving _run_clustermgtd")
```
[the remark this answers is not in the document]
agobeaux: Converted to log.debug

src/slurm_plugin/clustermgtd.py
```diff
@@ -879,44 +907,235 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+
+
+class MetricsProvider(ABC):
```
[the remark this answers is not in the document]
agobeaux: New python file created: metrics.py

src/slurm_plugin/clustermgtd.py
```diff
@@ -879,44 +907,235 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+        self.metrics[name] = value
+
+    def add_metrics_provider(self, name, metrics_provider):
+        self.metrics_providers[name] = metrics_provider
+
+    def get_values(self):
+        time_since_last_metric = int((datetime.now(tz=timezone.utc) - self.timestamp).total_seconds())
+        time_between_metrics_update = self.config.loop_time if self.config else -1
+        # TODO: create a config parameter that replaces '3', the factor
+        if time_since_last_metric < 3 * time_between_metrics_update:
```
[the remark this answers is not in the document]
agobeaux: Named it MISSED_METRICS_UPDATE_FACTOR and added a little comment about it

src/slurm_plugin/clustermgtd.py
```diff
@@ -879,44 +907,235 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+        config = metrics_collector.config
+        current_thread = currentThread()
+        while getattr(current_thread, "keep_running", True):
+            start_time = datetime.now(tz=timezone.utc)
+
+            # get most recent values ([] if no values were uploaded at all (should happen only the first time)
+            metrics = metrics_collector.get_values()
+
+            # upload metrics
+            metricdata = [
```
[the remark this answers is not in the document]
agobeaux: Done

src/slurm_plugin/clustermgtd.py
```diff
@@ -879,44 +907,235 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+                }
+                for metric_name, metric_value in metrics.items()
+            ]
+            if metricdata:
+                log.info("Pushing metrics to CloudWatch")
+                config = metrics_collector.config
+                cw_client = boto3.client("cloudwatch", region_name=config.region, config=config.boto3_config)
+                # TODO: Change Namespace to Cluster Metrics or a better name
+                cw_client.put_metric_data(
+                    Namespace="Custom Namespace",
```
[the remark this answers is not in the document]
agobeaux: Done

src/slurm_plugin/clustermgtd.py
```diff
@@ -879,44 +907,235 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+                    MetricData=metricdata,
+                )
+            else:
+                log.warning(
+                    "No metrics to upload: Either no metric was uploaded or the clustermgtd deamon"
+                    " has not uploaded new values for some time"
```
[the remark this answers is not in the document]
agobeaux: Global var = MISSED_METRICS_UPDATE_FACTOR, tried to give more useful info

src/slurm_plugin/clustermgtd.py
```diff
@@ -879,44 +907,235 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+                )
+
+            # FIXME What happens if the put_metric_data call takes too much time (more than config.loop_time) ?
```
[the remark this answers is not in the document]
agobeaux: [STILL TODO] Let's talk about this

src/slurm_plugin/clustermgtd.py
```diff
@@ -879,44 +907,235 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+            # At the moment, the delay would simply add up every time... Should we have this or should we
+            # upload faster if we are late? This could be an idea but a problem is that if it happens because of
+            # performance problems (instance very busy?), then, it would make it even more difficult for the instance
+            # to do its work.
+            sleep_remaining_loop_time(config.loop_time, start_time)
+    except Exception as e:
+        failed_threads.append("metrics uploader")
+        log.exception("An unexpected error occurred: %s", e)
+        return
+    log.info("Leaving _push_metrics_to_cloudwatch")
```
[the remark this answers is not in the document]
agobeaux: Used debug

src/slurm_plugin/clustermgtd.py
```diff
@@ -879,44 +907,235 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+    # signal children threads they should stop
+    for subthread in subthreads.values():
+        subthread.keep_running = False
+
+    # wait that children threads stop (join is a blocking action)
+    for subthread in subthreads.values():
+        # Check if thread is alive as calling join() on a unstarted thread raises an exception
+        if subthread.is_alive():
+            subthread.join()
+    log.info("Stopping main, will retry later")
```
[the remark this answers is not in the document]
agobeaux: used debug, same for log.info("Killing children threads")

src/slurm_plugin/clustermgtd.py
```diff
@@ -879,44 +907,235 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+
+
+def _create_and_start_threads(subthreads, clustermgtd_config_file, metrics_collector, failed_threads):
+    # cluster management deamon
+    subthreads["scaling"] = RestartableThread(
```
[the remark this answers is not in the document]
agobeaux: Done, used SCALING_THREAD_NAME and METRICS_UPLOADER_THREAD_NAME

src/slurm_plugin/clustermgtd.py
```diff
@@ -879,44 +907,235 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
-        except Exception as e:
-            log.warning(
-                "Unable to configure logging from %s, using default logging settings.\nException: %s",
-                config.logging_config,
-                e,
-            )
-        # Manage cluster
-        cluster_manager.manage_cluster()
-        sleep_remaining_loop_time(config.loop_time, start_time)
+        thread_names = []
```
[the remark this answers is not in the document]
agobeaux: Done, also changed thread_name --> failed_thread_name near it

src/slurm_plugin/clustermgtd.py
```diff
@@ -879,44 +907,235 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+            thread_name = failed_threads.pop()
+            thread_names.append(thread_name)
+            subthreads[thread_name] = subthreads[thread_name].clone()
+
+        # Restarting failed threads
+        for thread_name in thread_names:
+            log.info("Restarting %s thread", thread_name)
+            subthreads[thread_name].start()
+
+        sleep_remaining_loop_time(metrics_collector.config.loop_time // 2, start_time)
```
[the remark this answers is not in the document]
agobeaux: Added a little comment

src/slurm_plugin/clustermgtd.py
agobeaux: Using this now, thanks for the suggestion!

enrico-usai:
We're not going to implement it for now.

PR #256, opened 2020-09-16
Stop using S3 configuration files

tilne:
Instead of fetching vCPU and GPU data from instances.json, use the
DescribeInstanceTypes API.

Note that these changes will only affect single instance type (SIT) clusters (i.e., clusters using sge or torque as the scheduler). Heterogeneous instance type (HIT) clusters (those created using v2.9.0 or later and using slurm as the scheduler) already do not use the config files in S3.

Signed-off-by: Tim Lane <tilne@amazon.com>

### DIFF
--- a/src/common/utils.py
+++ b/src/common/utils.py
@@ -11,7 +11,6 @@
 # See the License for the specific language governing permissions and limitations under the License.
 import collections
 import itertools
-import json
 import logging
 import os
 import pwd
@@ -23,6 +22,7 @@ from datetime import datetime, timezone
 from enum import Enum
 
 import boto3
+from botocore.exceptions import ClientError
 from retrying import retry
 
 from common.time_utils import seconds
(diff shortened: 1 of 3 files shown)


src/common/utils.py
```diff
@@ -241,20 +241,45 @@ def _read_cfnconfig():
...
+    ec2_client = boto3.client("ec2", region_name=region, config=proxy_config)
+    try:
+        return ec2_client.describe_instance_types(InstanceTypes=[instance_type]).get("InstanceTypes")[0]
+    except (ClientError, Exception) as exception:
+        error_msg = "Error when calling DescribeInstanceTypes for instance type {0}: {1}".format(
+            instance_type,
+            exception.response.get("Error").get("Message") if isinstance(exception, ClientError) else exception,
+        )
         log.critical(error_msg)
         raise CriticalError(error_msg)
```
demartinofra, line 269: if we always raise CriticalError on ClientError I'm afraid the retry is never applied

demartinofra (commented):
LGTM. just dropped a single comment on the retry in case of throttling

src/common/utils.py
tilne: good call, i missed that

tilne: should be addressed in the changes i just pushed

demartinofra (approved):

PR #257, opened 2020-10-06
Add node memory size to slurm config

cread:
*Issue #, if available:*

None, as it looks like issues are disabled for this repo.

*Description of changes:*

I want to use memory as a consumable resource in my slurm configuration. 
This adds the required configuration to slurm on a per node basis as new nodes come online. This is required for trying to do this with mixed node types.

This will have no impact on anyone running the current configuration. 

### DIFF
--- a/src/common/utils.py
+++ b/src/common/utils.py
@@ -41,7 +41,7 @@ class EventType(Enum):
     REMOVE = "REMOVE"
 
 
-Host = collections.namedtuple("Host", ["instance_id", "hostname", "slots", "gpus"])
+Host = collections.namedtuple("Host", ["instance_id", "hostname", "slots", "gpus", "memory"])
 UpdateEvent = collections.namedtuple("UpdateEvent", ["action", "message", "host"])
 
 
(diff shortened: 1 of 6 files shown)


rexcsn (commented):
Hi @cread,

Thank you for opening this PR.
I believe @nstornetta have communicated with you offline that this approach is not compatible with the our new slurm architecture that is applicable for aws-parallelcluster>=2.9.0.

Under the new architecture, old scaling daemons jobwatcher/sqswatcher/nodewatcher are no longer used in our slurm integration. An action item from our side would be to remove slurm modules from these old daemons.

#### Here is a brief explanation on the difference between old and new slurm integration architecture:
As you are probably aware, our old architecture would add and remove node entries from slurm configuration and restart slurm daemons to accomplish scale up/down process.

Under the new architecture, we no longer modify slurm configuration or restart slurm daemons during cluster runtime. Our integration is based on slurm's cloud bursting plugins.
We start with a complete configuration, already containing all nodes that could potentially be in the system.
During scale up, we would simply "enable" nodes by configuring nodes with EC2 backing instances through ResumeProgram. During scale down, we would "disable" nodes by clearing the node-instance association and resetting the node. This would normally be done through SuspendProgram. However, we decided to perform this action in our daemon, clustermgtd, for better performance at scale.
In addition to resume and suspend programs, we also implement daemons to maintain the cluster and make sure that scheduler node states are consistent with EC2. Aforementioned clustermgtd does most of this work. You can find some high level description about our daemons [here in our official doc](https://docs.aws.amazon.com/parallelcluster/latest/ug/processes.html#clustermgtd)
Source code for resume/suspend program and our daemons are under the [slurm_plugin directory](https://github.com/aws/aws-parallelcluster-node/tree/develop/src/slurm_plugin) in the node repository.

#### Where to introduce changes required for memory support
Under the new architecture, we generate the complete slurm configuration based on an internal representation of queue settings we get from customer configuration in cookbook part of the product. Note that the slurm configuration is now only generated on cluster creation and cluster update, as we no longer need to modify slurm configuration during scale up/down.

First, we would need to make sure that our internal representation of queue settings contain memory information about an instance type.
Properties about a compute resources are retrieved dynamically during cluster creation in the main package, and managed by [`QueueJsonSection`](https://github.com/aws/aws-parallelcluster/blob/v2.9.1/cli/pcluster/config/json_param_types.py#L223) parameter. We need to add logic to retrieve memory in addition to logic that retrieves cpu and gpu info already in place.

After that, we can modify the config generator to generate configuration containing memory settings. Config generator is implemented based on jinja templating and is [here](https://github.com/aws/aws-parallelcluster-cookbook/tree/develop/files/default/slurm) in our cookbook repository.
This is the place we would modify and add memory configuration to different nodes in slurm configuration.

#### Finally
Thank you again for opening this PR, but we cannot accept it as it is incompatible with our new slurm integration architecture.
Please let us know if you have any question. We look forward to seeing your future contributions!

cread:
Thanks for the detailed explanation. I'll do some additional testing and see how things go.

PR #258, opened 2020-10-07
Fix cfn_scheduler_slots in pcluster update

yuleiwan:
* Fix retrieving an old value of compute instance slots by cfn_scheduler_slots when compute instance type was updated
* Use DescribeInstanceType to retrieve updated compute instance vcpus
* Remove name from_pricing_file
* Add unit test
* [cli link](https://github.com/aws/aws-parallelcluster/pull/2166) and [integration test link](https://github.com/aws/aws-parallelcluster/pull/2114)

Signed-off-by: Yulei Wang <yuleiwan@amazon.com>


src/common/utils.py
```diff
@@ -327,6 +330,14 @@ def get_instance_properties(region, proxy_config, instance_type):
                     )
                 )
                 slots = vcpus
+
+            if slots > vcpus:
```
demartinofra, line 334: there might be some remote cases when you want a node to appear bigger than it is.  we might want to avoid that check since using an integer in cfn_scheduler_slots is already a strong customization

yuleiwan: Removed

demartinofra (approved):

PR #260, opened 2020-10-26
Enable all or nothing scaling on batch level

rexcsn:
* Add new ResumeProgram config parameter, `all_or_nothing_batch`, to configure behavior of the RunInstances calls
* With `all_or_nothing_batch = True`, RunInstances calls will not succeed unless all capacity for the batch can be launched
* Updated unit tests
* Increase default timeouts for slurm commands from 5 to 10 seconds, which is consistent with timeout explicitly used in `clustermgtd`


src/common/schedulers/slurm_commands.py
```diff
@@ -317,7 +317,7 @@ def set_nodes_down_and_power_save(node_list, reason):
     set_nodes_power_down(node_list, reason=reason)
 
 
-def get_nodes_info(nodes, command_timeout=5):
+def get_nodes_info(nodes, command_timeout=15):
```
demartinofra, line 320: can we leave a trace on why it was necessary to increase he timeout? Also if I'm not wrong Slurm commands timeout after 10 seconds

nit: it would be nice to set this default timeout as a module constant so that we can change it in a single place when necessary

tests/slurm_plugin/test_resume.py
```diff
@@ -56,3 +69,142 @@ def test_resume_config(config_file, expected_attributes, test_datadir, mocker):
...
+    mock_handle_failed_nodes = mocker.patch("slurm_plugin.resume._handle_failed_nodes", auto_spec=True)
+    # patch slurm calls
+    mocker.patch("slurm_plugin.common.update_nodes", auto_spec=True)
+    mocker.patch("slurm_plugin.resume.get_nodes_info", return_value=mock_node_lists, auto_spec=True)
+    # patch DNS related functions
+    mocker.patch.object(slurm_plugin.common.InstanceManager, "_store_assigned_hostnames", auto_spec=True)
+    mocker.patch.object(slurm_plugin.common.InstanceManager, "_update_dns_hostnames", auto_spec=True)
+
+    _resume("some_arg_nodes", mock_resume_config)
+    mock_handle_failed_nodes.assert_called_with(expected_failed_nodes)
```
demartinofra, line 210: shall we assert also other mock are called as expected?

demartinofra (commented):
LGTM! dropped a couple of minor comments

src/common/schedulers/slurm_commands.py
demartinofra: oh you are right it was 1 min. Thanks

demartinofra (approved):

PR #265, opened 2020-11-04
Fix clustermgtd and computemgtd sleep time computation

demartinofra:
This was broken when the system timezone was not set to UTC because it was forcing the timezone to be UTC when unset. In Python's datetime object when timezone is unset the so called naive datetime relies on the system local time.

This addresses: https://github.com/aws/aws-parallelcluster/issues/2162


rexcsn (approved):
LGTM!

PR #272, opened 2020-12-08
Improve error handling in slurm plugin processes when clustermgtd is down

rexcsn:
* Computemgtd: Add logic to consider self node as down if node appears to be in power saving. When a node is in appears to be in power saving but the backing instance is up the node is not correctly attached to the scheduler. Computemgtd should consider self node as node and self-terminate the instance in this case
* ResumeProgram: Add logic to prevent launching new instances when unable to retrieve relevant clustermgtd heartbeat
* SuspendProgram: Print additional warning message when unable to retrieve relevant clustermgtd heartbeat
* Fix unit tests for above changes

### DIFF
--- a/src/slurm_plugin/resume.py
+++ b/src/slurm_plugin/resume.py
@@ -12,6 +12,7 @@
 
 import logging
 import os
+from datetime import datetime, timezone
 from logging.config import fileConfig
 
 import argparse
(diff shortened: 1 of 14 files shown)


src/slurm_plugin/resume.py
```diff
@@ -115,6 +129,18 @@ def _handle_failed_nodes(node_list):
...
+    if not is_clustermgtd_heartbeat_valid(
+        current_time, resume_config.clustermgtd_timeout, resume_config.clustermgtd_heartbeat_file_path
+    ):
+        log.error(
+            "No valid clustermgtd heartbeat detected, clustermgtd is down!\n"
+            "Please check clustermgtd log for error.\n"
+            "Not launching nodes %s",
+            arg_nodes,
+        )
+        raise Exception("Clustermgtd down")
```
demartinofra, line 143: nit: can we handle it here rather than raising and having this caught in the generic unhandled exception block of the main function?

demartinofra (approved):

PR #273, opened 2020-12-18
Add retries to change_resource_record_sets Route53 API call

demartinofra:
change_resource_record_sets Route53 API has an account why throttling. I'm adding some retry to slurm_resume in order to better handle such failures.

### DIFF
--- a/src/slurm_plugin/common.py
+++ b/src/slurm_plugin/common.py
@@ -18,6 +18,7 @@ import subprocess
 from datetime import datetime, timezone
 
 import boto3
+from botocore.config import Config
 from botocore.exceptions import ClientError
 
 from common.schedulers.slurm_commands import InvalidNodenameError, parse_nodename, update_nodes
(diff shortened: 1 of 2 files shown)


src/slurm_plugin/common.py
```diff
@@ -244,7 +245,13 @@ def _update_dns_hostnames(self, nodes):
             # Submit calls to change_resource_record_sets in batches of 500 elements each.
             # change_resource_record_sets API call has limit of 1000 changes,
             # but the UPSERT action counts for 2 calls
-            route53_client = boto3.client("route53", region_name=self._region, config=self._boto3_config)
+            # Also pick the number of retries to be the max between the globally configured one and 3.
+            # This is done to address Route53 API throttling without changing the configured retries for all API calls.
+            configured_retry = self._boto3_config.retries.get("max_attempts", 0) if self._boto3_config.retries else 0
+            boto3_config = self._boto3_config.merge(
```
enrico-usai, line 251: What about moving that logic in a common function?  We're overriding `boto3_config` to be sure to have a max attempts greater than "a specified value" and the code is very specific and strictly related to the `boto3_config.retries` structure.

enrico-usai (approved):
I left a comment but the code is good to me.

PR #278, opened 2021-02-22
Use cloud_reg_addrs option and do not manually reset nodeaddr on power_down

rexcsn:
* Using cloud_reg_addrs and let slurm reset nodeaddr automatically. This option is available in Slurm 20.11.
* Keep reset nodes for inactivate partitions because nodes are not power_down in this case

### DIFF
--- a/src/slurm_plugin/clustermgtd.py
+++ b/src/slurm_plugin/clustermgtd.py
@@ -758,8 +758,13 @@ class ClusterManager:
                 and self._is_node_state_healthy(node, private_ip_to_instance_map)
             )
         else:
-            return ClusterManager._is_backing_instance_valid(
-                node, instance_ips_in_cluster=list(private_ip_to_instance_map.keys())
+            # Do not handle powering_down nodes
+            # powering_down nodes should be already handled by _handle_powering_down_nodes
+            return (
+                node.is_powering_down()
+                or ClusterManager._is_backing_instance_valid(
+                    node, instance_ips_in_cluster=list(private_ip_to_instance_map.keys())
+                )
             ) and self._is_node_state_healthy(node, private_ip_to_instance_map)
 
     @log_exception(log, "maintaining unhealthy dynamic nodes", raise_on_error=False)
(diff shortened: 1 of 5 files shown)


src/slurm_plugin/clustermgtd.py
```diff
@@ -798,8 +798,6 @@ def _handle_powering_down_nodes(self, slurm_nodes, private_ip_to_instance_map):
                 powering_down_nodes.append(node)
 
         if powering_down_nodes:
-            log.info("Resetting powering down nodes: %s", print_with_count(powering_down_nodes))
-            reset_nodes(nodes=[node.name for node in powering_down_nodes])
```
demartinofra, line 802: when is this resetting happening? as soon as the powering down is initiated? If not are there some downstream consequences if we don't reset it here and Slurm hasn't reset this value yet?

demartinofra: let's also update the docstring by mentioning that we don't reset the nodeaddr since Slurm takes care of it

rexcsn: Docstring updated. Made modification to code to account for the fact that nodeaddr are not reset for powering_down nodes.

demartinofra (commented):
how are now powering down nodes handled?do they result as orphaned instances? If so we might want to adjust the logging message making it clear that we are terminating not only unexpected orphaned instances but also powering down nodes.

src/slurm_plugin/clustermgtd.py
```diff
@@ -790,16 +795,16 @@ def _handle_powering_down_nodes(self, slurm_nodes, private_ip_to_instance_map):
...
         """
         powering_down_nodes = []
         for node in slurm_nodes:
             if not node.is_static and node.is_nodeaddr_set() and (node.is_power() or node.is_powering_down()):
                 powering_down_nodes.append(node)
 
         if powering_down_nodes:
-            log.info("Resetting powering down nodes: %s", print_with_count(powering_down_nodes))
-            reset_nodes(nodes=[node.name for node in powering_down_nodes])
             instances_to_terminate = ClusterManager._get_backing_instance_ids(
```
demartinofra, line 808: can we still retrieve the mapping here?when is the nodeaddr unset by Slurm?

demartinofra (commented):
how are now powering down nodes handled?do they result as orphaned instances? If so we might want to adjust the logging message making it clear that we are terminating not only unexpected orphaned instances but also powering down nodes.

demartinofra (dismissed):

src/slurm_plugin/clustermgtd.py
rexcsn: Added some comments to clarify this

demartinofra (approved):

PR #288, opened 2021-04-05
Transition from IMDSv1 to IMDSv2

chenwany:
Signed-off-by: chenwany <chenwany@amazon.com>

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/slurm_plugin/computemgtd.py
+++ b/src/slurm_plugin/computemgtd.py
@@ -21,7 +21,7 @@ from subprocess import CalledProcessError
 from botocore.config import Config
 from common.schedulers.slurm_commands import get_nodes_info
 from common.time_utils import seconds
-from common.utils import check_command_output, sleep_remaining_loop_time
+from common.utils import get_metadata, sleep_remaining_loop_time
 from retrying import retry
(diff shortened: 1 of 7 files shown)


src/slurm_plugin/computemgtd.py
```diff
@@ -114,8 +114,13 @@ def _self_terminate(computemgtd_config):
     instance_manager = InstanceManager(
         computemgtd_config.region, computemgtd_config.cluster_name, computemgtd_config.boto3_config
     )
+    token = check_command_output(  # nosec
```
rexcsn, line 117: Should we use the `requests` based way to get the metadata here as well? We can put a utility function to get metadata in `common/utils.py` so `nodewatcher` and `computemgtd` can use the same function?

chenwany: From the previous code in `computemgtd` and `nodewatcher`, it looks like  in src/nodewatcher/nodewatcher.py, it is running a python codes to get the metadata.
In `computemgtd` def _self_terminate(computemgtd_config): each commands are execute shell commands and retrieve command output.
I am not sure whether we want to change it be run python code

chenwany: After a second thought, I think it is a good idea to move it and put it in utils, I will test it after change

rexcsn (approved):

rexcsn (approved):

PR #289, opened 2021-04-07
3.0: Drop deprecated features and use inclusive language

enrico-usai:
* Remove every reference to Torque and SGE schedulers
* Remove jobwatcher, sqswatcher and nodewatcher code.
* Remove nodewatcher, sqswatcher, jobwatcher references.
* Remove ASG utilities.
* Use inclusive language: from master to head node.


src/common/utils.py
```diff
@@ -306,7 +260,7 @@ def _get_instance_info(region, proxy_config, instance_type, additional_instance_
 
 def get_instance_properties(region, proxy_config, instance_type, additional_instance_types_data=None):
```
demartinofra, line 261: we could also remove this logic since not used in slurm plugin if I recall correctly

enrico-usai: Done, I removed also other unused utility functions.

tox.ini
```diff
@@ -15,8 +15,7 @@ deps =
     -r tests/requirements.txt
     cov: codecov
 commands =
-    bash ./tests/test.sh
-    nocov: python -m unittest discover -s tests/jobwatcher -p "*tests.py"
+    nocov: python -m unittest discover -p "*tests.py"
```
demartinofra, line 18: we don't need this cause the only unit test which was not in pytest has been removed

demartinofra (changes requested):
LGTM! one remaining minor thing and then good to go

tox.ini
enrico-usai: Removed, thanks

demartinofra (approved):

PR #294, opened 2021-05-11
Implement Slurm Protected mode

chenwany:
* Identify nodes with bootstrap failure.
* Keep count of failures
*  Reset failure when there is a successful launch of the node type have been failed previously
*  Disable queue and put cluster in protected mode if failure > protected_failure_count, which is set to 10.
*  User can check cluster status with pcluster status and manually reactivate queue with pcluster start. When executing `pcluster start`, the protected failure count will be reset.
*  Reduce "node_replacement_timeout" and Resume Timeout to  1800 seconds instead of 3600 seconds

### DIFF
--- a/src/common/schedulers/slurm_commands.py
+++ b/src/common/schedulers/slurm_commands.py
@@ -15,7 +15,7 @@ import logging
 import re
 from enum import Enum
 
-from common.utils import check_command_output, grouper, run_command
+from common.utils import check_command_output, convert_range_to_list, grouper, run_command
 from retrying import retry
 
 PENDING_RESOURCES_REASONS = [
@@ -47,6 +47,7 @@ SCONTROL = "/opt/slurm/bin/scontrol"
 SINFO = "/opt/slurm/bin/sinfo"
 
 SlurmPartition = collections.namedtuple("SlurmPartition", ["name", "nodes", "state"])
+NodeType = collections.namedtuple("NodeType", ["instance_type", "partition"])
 
 # Set default timeouts for running different slurm commands.
 # These timeouts might be needed when running on large scale
@@ -71,6 +72,10 @@ class SlurmNode:
     SLURM_SCONTROL_DRAIN_STATE = "DRAIN"
     SLURM_SCONTROL_POWERING_DOWN_STATE = "POWERING_DOWN"
     SLURM_SCONTROL_POWER_STATE = "IDLE+CLOUD+POWER"
+    SLURM_SCONTROL_POWER_UP_STATE = "#"
+    SLURM_SCONTROL_ONLINE_STATES = {"IDLE+CLOUD", "MIXED+CLOUD", "ALLOCATED+CLOUD", "COMPLETING+CLOUD"}
+    SLURM_SCONTROL_POWER_WITH_JOB_STATE = "MIXED+CLOUD+POWER"
+    SLURM_SCONTROL_RESUME_FAILED_STATE = "DOWN*+CLOUD+POWER"
 
     def __init__(self, name, nodeaddr, nodehostname, state, partitions=None):
         """Initialize slurm node with attributes."""
@@ -118,6 +123,30 @@ class SlurmNode:
     def is_up(self):
         """Check if slurm node is in a healthy state."""
         return not self._is_drain() and not self.is_down() and not self.is_powering_down()
+
+    def is_powering_up(self):
+        """Check if slurm node is in powering up state."""
+        return self.SLURM_SCONTROL_POWER_UP_STATE in self.state
+
+    def is_online(self):
+        """Check if slurm node is online with backing instance."""
+        return self.state in self.SLURM_SCONTROL_ONLINE_STATES
+
+    def is_configuring_job(self):
+        """Check if slurm node is configuring with job and haven't begun to run a job."""
+        return self.is_powering_up() and self.has_job()
+
+    def is_power_with_job(self):
+        """Dynamic nodes allocated a job but power up process has not started yet."""
+        return self.state == self.SLURM_SCONTROL_POWER_WITH_JOB_STATE
+
+    def is_running_job(self):
+        """Check if slurm node is running a job but not in configuring job state."""
+        return not self.is_powering_up() and self.has_job() and not self.is_power_with_job()
+
+    def is_resume_failed(self):
+        """Check if node resume timeout expires."""
+        return self.state == self.SLURM_SCONTROL_RESUME_FAILED_STATE
 
     def __eq__(self, other):
         """Compare 2 SlurmNode objects."""
@@ -440,3 +469,51 @@ def _parse_nodes_info(slurm_node_info):
         if lines:
             slurm_nodes.append(SlurmNode(*lines))
     return slurm_nodes
+
+
+def get_nodes_type(nodes):
+    """Get node type given nodename. Node type format: NodeType(instancetype, queue_name)."""
+    nodes_types = set()
+    for node in nodes:
+        queue_name, _, instance_name = parse_nodename(node.name)
+        nodes_types.add(NodeType(instance_name, queue_name))
+    return nodes_types
+
+
+def get_inactive_partition_names(partitions):
+    """Get the name of all inactive partitions from partitions and name mapping."""
+    inactive_partitions_name = []
+    for partition_name, partition in partitions.items():
+        if partition.state == "INACTIVE":
+            inactive_partitions_name.append(partition_name)
+    return inactive_partitions_name
+
+
+def retrieve_partitions_from_node_types(node_types):
+    """Retrieve partitions for node types."""
+    return {node_type.partition for node_type in node_types}
+
+
+def get_partition_node_names(nodenames):
+    """
+    Convert partition nodenames read from partition to a list of nodes.
+
+    Example input nodenames: "queue1-st-c5xlarge-[1,3,4-5],queue1-st-c5large-20"
+    Example output [queue1-st-c5xlarge-1, queue1-st-c5xlarge-3, queue1-st-c5xlarge-4, queue1-st-c5xlarge-5,
+    queue1-st-c5large-20]
+    """
+    if type(nodenames) is str:
+        matches = re.findall(r"((([a-z0-9\-]+)-(st|dy)-([a-z0-9]+)-)(\[[\d+,-]+\]|\d+))", nodenames)
+        # [('queue1-st-c5xlarge-[1,3,4-5]', 'queue1-st-c5xlarge-', 'queue1', 'st', 'c5xlarge', '[1,3,4-5]'),
+        # ('queue1-st-c5large-20', 'queue1-st-c5large-', 'queue1', 'st', 'c5large', '20')]
+    partition_node_names = []
+    for match in matches:
+        nodename, prefix, _, _, _, nodes = match
+        if "[" not in nodes:
+            # Single nodename
+            partition_node_names.append(nodename)
+        else:
+            # Multiple nodenames
+            node_list = convert_range_to_list(nodes.strip("[]"))
+            partition_node_names += [prefix + str(n) for n in node_list]
+    return partition_node_names
--- a/src/slurm_plugin/clustermgtd.py
+++ b/src/slurm_plugin/clustermgtd.py
@@ -24,13 +24,18 @@ from boto3.dynamodb.conditions import Attr
 from botocore.config import Config
 from common.schedulers.slurm_commands import (
     PartitionStatus,
+    get_inactive_partition_names,
     get_nodes_info,
+    get_nodes_type,
     get_partition_info,
+    get_partition_node_names,
     reset_nodes,
+    retrieve_partitions_from_node_types,
     set_nodes_down,
     set_nodes_down_and_power_save,
     set_nodes_drain,
     update_all_partitions,
+    update_partitions,
 )
 from common.time_utils import seconds
 from common.utils import sleep_remaining_loop_time
@@ -59,6 +64,8 @@ class ComputeFleetStatus(Enum):
     STARTING = "STARTING"  # clustermgtd is handling the start request.
     STOP_REQUESTED = "STOP_REQUESTED"  # A request to stop the fleet has been submitted.
     START_REQUESTED = "START_REQUESTED"  # A request to start the fleet has been submitted.
+    # PROTECTED indicates that some partitions have consistent bootstrap failures. Affected partitions are inactive.
+    PROTECTED = "PROTECTED"
 
     def __str__(self):
         return str(self.value)
@@ -74,6 +81,10 @@ class ComputeFleetStatus(Enum):
     @staticmethod
     def is_stop_in_progress(status):
         return status in {ComputeFleetStatus.STOP_REQUESTED, ComputeFleetStatus.STOPPING}
+
+    @staticmethod
+    def is_protected_status(status):
+        return status == ComputeFleetStatus.PROTECTED
 
 
 class ComputeFleetStatusManager:
@@ -130,7 +141,7 @@ class ClustermgtdConfig:
         # Terminate configs
         "terminate_max_batch_size": 1000,
         # Timeout to wait for node initialization, should be the same as ResumeTimeout
-        "node_replacement_timeout": 3600,
+        "node_replacement_timeout": 1800,
         "terminate_drain_nodes": True,
         "terminate_down_nodes": True,
         "orphaned_instance_timeout": 120,
@@ -143,6 +154,7 @@ class ClustermgtdConfig:
         "hosted_zone": None,
         "dns_domain": None,
         "use_private_hostname": False,
+        "protected_failure_count": 10,
     }
 
     def __init__(self, config_file_path):
@@ -233,6 +245,9 @@ class ClustermgtdConfig:
         self.orphaned_instance_timeout = config.getint(
             "clustermgtd", "orphaned_instance_timeout", fallback=self.DEFAULTS.get("orphaned_instance_timeout")
         )
+        self.protected_failure_count = config.getint(
+            "clustermgtd", "protected_failure_count", fallback=self.DEFAULTS.get("protected_failure_count")
+        )
 
     def _get_dns_config(self, config):
         """Get config option related to Route53 DNS domain."""
@@ -287,6 +302,10 @@ class ClusterManager:
         This state is required because we need to ignore static nodes that might have long bootstrap time
         """
         self._static_nodes_in_replacement = set()
+        self._bootstrap_failure_nodes_types = set()
+        self._instance_ips_in_cluster = set()
+        self._partitions_protected_failure_count_map = {}
+        self._inactive_partitions = []
         self._compute_fleet_status = ComputeFleetStatus.RUNNING
         self._current_time = None
         self._config = None
@@ -371,12 +390,41 @@ class ClusterManager:
                 )
                 if partitions_activated_successfully:
                     self._update_compute_fleet_status(ComputeFleetStatus.RUNNING)
+                    # Reset protected failure
+                    self._partitions_protected_failure_count_map = {}
+                    self._bootstrap_failure_nodes_types = set()
         except ComputeFleetStatusManager.ConditionalStatusUpdateFailed:
             log.warning(
                 "Cluster status was updated while handling a transition from %s. "
                 "Status transition will be retried at the next iteration",
                 self._compute_fleet_status,
             )
+
+    def _handle_successfully_launched_nodes(self, healthy_nodes):
+        """
+        Handle nodes have been failed in bootstrap are launched successfully during this iteration.
+
+        Includes resetting partition bootstrap failure count for these nodes type and updating the
+        bootstrap_failure_nodes_types.
+        If a node has been failed successfully launched, the partition count of this node will be reset.
+        So the successfully launched nodes type will be remove from bootstrap_failure_nodes_types. If there's node types
+        failed in this partition later, it will keep count again.
+        """
+        online_nodes_types = self._get_online_nodes_types(healthy_nodes)
+        bootstrap_failure_nodes_types_successfully_launched = online_nodes_types.intersection(
+            self._bootstrap_failure_nodes_types
+        )
+
+        if bootstrap_failure_nodes_types_successfully_launched:
+            # Reset bootstrap failure count of successfully launched partitions
+            successful_launched_partitions = retrieve_partitions_from_node_types(
+                bootstrap_failure_nodes_types_successfully_launched
+            )
+            for partition in successful_launched_partitions:
+                self._reset_partition_failure_count(partition)
+
+            # Remove successfully launched node types from the set of bootstrap failure nodes.
+            self._bootstrap_failure_nodes_types -= bootstrap_failure_nodes_types_successfully_launched
 
     def manage_cluster(self):
         """Manage cluster by syncing scheduler states with EC2 states and performing node maintenance actions."""
@@ -389,13 +437,19 @@ class ClusterManager:
         if not self._config.disable_all_cluster_management and self._compute_fleet_status in {
             None,
             ComputeFleetStatus.RUNNING,
+            ComputeFleetStatus.PROTECTED,
         }:
             # Get node states for nodes in inactive and active partitions
             try:
                 log.info("Retrieving nodes info from the scheduler")
-                active_nodes, inactive_nodes = self._get_node_info_from_partition()
+                partitions = {
+                    partition.name: partition for partition in ClusterManager._get_partition_info_with_retry()
+                }
+                log.debug("Partitions: %s", partitions)
+                active_nodes, inactive_nodes = self._get_node_info_from_partition(partitions)
                 log.debug("Current active slurm nodes in scheduler: %s", active_nodes)
                 log.debug("Current inactive slurm nodes in scheduler: %s", inactive_nodes)
+                self._inactive_partitions = get_inactive_partition_names(partitions)
             except ClusterManager.SchedulerUnavailable:
                 log.error("Unable to get partition/node info from slurm, no other action can be performed. Sleeping...")
                 return
@@ -416,6 +470,7 @@ class ClusterManager:
                 cluster_instances = self._clean_up_inactive_partition(inactive_nodes, cluster_instances)
             log.debug("Current cluster instances in EC2: %s", cluster_instances)
             private_ip_to_instance_map = {instance.private_ip: instance for instance in cluster_instances}
+            self._instance_ips_in_cluster = set(private_ip_to_instance_map.keys())
             log.debug("private_ip_to_instance_map %s", private_ip_to_instance_map)
             # Clean up orphaned instances and skip all other operations if no active node
             if active_nodes:
@@ -423,7 +478,7 @@ class ClusterManager:
                 if not self._config.disable_all_health_checks:
                     self._perform_health_check_actions(cluster_instances, ip_to_slurm_node_map)
                 # Maintain slurm nodes
-                self._maintain_nodes(private_ip_to_instance_map, active_nodes)
+                self._maintain_nodes(private_ip_to_instance_map, active_nodes, partitions)
             # Clean up orphaned instances
             self._terminate_orphaned_instances(cluster_instances, ips_used_by_slurm=list(ip_to_slurm_node_map.keys()))
         # Write clustermgtd heartbeat to file
@@ -448,7 +503,7 @@ class ClusterManager:
         return get_partition_info(get_all_nodes=True)
 
     @staticmethod
-    def _get_node_info_from_partition():
+    def _get_node_info_from_partition(partitions):
         """
         Retrieve node belonging in UP/INACTIVE partitions.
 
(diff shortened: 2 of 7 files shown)


src/slurm_plugin/clustermgtd.py
```diff
@@ -904,6 +1042,135 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+
+    def _get_bootstrap_failure_nodes_types_successfully_launched(self, slurm_nodes, private_ip_to_instance_map):
+        """Find nodes types which have been failed during bootstrap now become online."""
+        online_nodes = self._find_online_slurm_nodes(slurm_nodes, private_ip_to_instance_map)
+        online_nodes_types = get_nodes_types_from_slurm_nodes(online_nodes)
+
+        bootstrap_failure_nodes_types_successfully_launched = online_nodes_types.intersection(
+            self._bootstrap_failure_nodes_types
+        )
+        return bootstrap_failure_nodes_types_successfully_launched
```
rexcsn, line 1138: What about simplifying this function to only return online node_types?

src/slurm_plugin/clustermgtd.py
```diff
@@ -904,6 +1042,135 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+
+    def _get_partitions_to_disable_and_bootstrap_failure_partitions_have_jobs(
```
rexcsn, line 1140: This function name is a bit too long, can we think of a better name?

src/slurm_plugin/clustermgtd.py
```diff
@@ -904,6 +1042,135 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+        # Get partitions have jobs running
+        partitions_have_active_jobs = ClusterManager._get_partitions_have_active_jobs(slurm_nodes)
+        # Get bootstrap failure partitions
+        bootstrap_failure_partitions = self._get_bootstrap_failure_partitions(
+            active_partitions_protected_failure_count_map
+        )
+        # Only disable bootstrap failure partitions that not having active jobs running
+        partitions_to_disable = set(bootstrap_failure_partitions) - partitions_have_active_jobs
+        bootstrap_failure_partitions_have_jobs = set(bootstrap_failure_partitions) - partitions_to_disable
+        return partitions_to_disable, bootstrap_failure_partitions_have_jobs
```
rexcsn, line 1153: We only use `bootstrap_failure_partitions_have_jobs` to print out a log message. Can we just print out a message here and avoid passing around anything we don't really need?

src/slurm_plugin/clustermgtd.py
```diff
@@ -904,6 +1042,135 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+        bootstrap_failure_partitions = []
+        for p in active_partitions_protected_failure_count_map:
+            if active_partitions_protected_failure_count_map.get(p) >= self._config.protected_failure_count:
+                bootstrap_failure_partitions.append(p)
+        if bootstrap_failure_partitions:
+            log.info(f"Bootstrap failure partitions: {bootstrap_failure_partitions}")
+        return bootstrap_failure_partitions
+
+    @staticmethod
+    def _get_partitions_have_active_jobs(slurm_nodes):
```
rexcsn, line 1116: Very minor: `_get_partitions_with_active_jobs`

src/slurm_plugin/clustermgtd.py
```diff
@@ -904,6 +1042,135 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+                    online_nodes.append(node)
+        return online_nodes
+
+    def _reset_partition_failure_count(self, partition):
+        """Reset bootstrap failure count for partition which has bootstrap failure nodes successfully launched."""
+        log.info(f"Find successfully launched node in partition {partition}, reset partition protected failure count")
+        self._partitions_protected_failure_count_map[partition] = 0
+
+    @staticmethod
+    def _retrieve_partitions_from_node_types(node_types):
```
rexcsn, line 1097: When we parse `node_types` from nodename, it should already be in a data structure that is ready for use by all functions that will be calling it. Why not use a dictionary for node types i.e.:
```
node_types = { partition: instance_type}
```
Here we are parsing `node_types` which is parsed from nodename, not that clean IMO

src/slurm_plugin/clustermgtd.py
```diff
@@ -904,6 +1042,135 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+        if partitions_to_disable:
+            # Change compute fleet status to protected
+            if not self._compute_fleet_status == ComputeFleetStatus.PROTECTED:
+                log.info("Put Cluster into protected mode")
+                self._update_compute_fleet_status(ComputeFleetStatus.PROTECTED)
+            # Place partitions into inactive
+            log.info(f"Placing bootstrap failure partitions to INACTIVE: {partitions_to_disable}")
+            update_partitions(partitions_to_disable, PartitionStatus.INACTIVE)
+        else:
+            log.info("Not entering protected mode since active are jobs running in bootstrap failure partitions")
```
rexcsn, line 1078: What's the difference between this line and the one above?
```
log.info( f"Bootstrap failure partitions {bootstrap_failure_partitions_have_jobs} currently have jobs running, not disabling them." )
```

src/slurm_plugin/clustermgtd.py
```diff
@@ -904,6 +1042,135 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+
+    def _find_online_slurm_nodes(self, slurm_nodes, private_ip_to_instance_map):
+        """Find slurm nodes online. Consider not DOWN static node and running job dynamic node as online nodes."""
+        online_nodes = []
+        for node in slurm_nodes:
+            if node.is_static and not node.is_down():
+                online_nodes.append(node)
+            else:
+                if self._is_node_healthy(node, private_ip_to_instance_map) and node.is_running_job():
```
rexcsn, line 1087: Why do we need this `is_running_job` logic? What if I submit a job and cancel the job later, but the dynamic node is up and running; should I consider this node as an online node?

src/slurm_plugin/clustermgtd.py
```diff
@@ -300,6 +313,8 @@ def set_config(self, config):
             self._config = config
             self._compute_fleet_status_manager = self._initialize_compute_fleet_status_manager(config)
             self._instance_manager = self._initialize_instance_manager(config)
+            self._disable_protected_mode = self._is_protected_mode_disabled(config)
+            self._partitions_protected_failure_count_map = self._initialize_partitions_protected_failure_count()
```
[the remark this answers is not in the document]
chenwany: Do we need to restart clustermgtd to apply the new config after we changing the config? Is it after the clustermgtd restart, the protected failure count will be clear, so it will reset failure count and keep count again?

src/common/schedulers/slurm_commands.py
```diff
@@ -440,3 +457,49 @@ def _parse_nodes_info(slurm_node_info):
...
+                partitions_have_active_jobs.add(partition_name)
+                break
+    return partitions_have_active_jobs
+
+
+def retrieve_partitions_from_node_types(node_types):
+    """Retrieve partitions for node types."""
+    successful_launched_partitions = []
+    for nodes_type in node_types:
+        partitions, _ = re.match(r"^([a-z0-9\-]+)-([a-z0-9]+)", nodes_type).groups()
```
rexcsn, line 503: As we discussed, please try to avoid parsing information we should already have from `parse_nodename` again

src/slurm_plugin/clustermgtd.py
```diff
@@ -904,6 +1042,135 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+    @staticmethod
+    def _get_partitions_have_active_jobs(slurm_nodes):
+        """Retrieve partitions currently have active jobs running."""
+        partitions_have_active_jobs = []
+        for node in slurm_nodes:
+            if node.is_running_job():
+                partitions_have_active_jobs += node.partitions
+        return set(partitions_have_active_jobs)
+
+    def _handle_health_check_bootsrap_failure_node(self, unhealthy_node):
```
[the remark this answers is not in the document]
rexcsn: As we discussed, let's generalize this function and its callers to handle multiple nodes so we don't need to deal with single node case separately

src/slurm_plugin/clustermgtd.py
```diff
@@ -287,6 +296,10 @@ def __init__(self, config):
         This state is required because we need to ignore static nodes that might have long bootstrap time
         """
         self._static_nodes_in_replacement = set()
+        self._bootstrap_failure_nodes_types = set()
+        self._partitions_protected_failure_count_map = None
+        self._inactive_partitions = None
+        self._disable_protected_mode = False
```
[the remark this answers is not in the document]
chenwany: Done

src/slurm_plugin/clustermgtd.py
```diff
@@ -323,6 +338,18 @@ def _initialize_compute_fleet_status_manager(config):
...
         )
 
+    @staticmethod
+    def _is_protected_mode_disabled(config):
+        if config.protected_failure_count == -1:  # disable protected mode when setting protected_failure_count to -1
+            log.info("Disable protected mode")
+            return True
+        return False
+
+    def _initialize_partitions_protected_failure_count(self):
```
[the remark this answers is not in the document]
chenwany: Fixed

tests/slurm_plugin/test_clustermgtd.py
```diff
@@ -1232,14 +1243,21 @@ def test_maintain_nodes(private_ip_to_instance_map, active_nodes, mock_unhealthy
...
     # Run test
-    cluster_manager._maintain_nodes(private_ip_to_instance_map, active_nodes)
+    cluster_manager._maintain_nodes(private_ip_to_instance_map, active_nodes, partitions)
     # Check function calls
     mock_update_replacement.assert_called_with(active_nodes)
     mock_find_unhealthy.assert_called_with(active_nodes, private_ip_to_instance_map)
     mock_handle_dynamic.assert_called_with(mock_unhealthy_nodes[0], private_ip_to_instance_map)
     mock_handle_static.assert_called_with(mock_unhealthy_nodes[1], private_ip_to_instance_map)
     mock_handle_powering_down_nodes.assert_called_with(active_nodes, private_ip_to_instance_map)
+    assert_that(mock_handle_protected_mode_process.call_count).is_equal_to(1)
```
rexcsn, line 1257: Can we add a test case to test that `handle_protected_mode_process` is not called when protected mode is disabled?

tests/slurm_plugin/test_clustermgtd.py
```diff
@@ -2112,6 +2170,794 @@ def test_update_status(self, compute_fleet_status_manager, put_item_response, ex
...
+    cluster_manager._bootstrap_failure_nodes_types = bootstrap_failure_nodes_types
+    mock_reset_partition_failure_count = mocker.patch.object(
+        cluster_manager, "_reset_partition_failure_count", auto_spec=True
+    )
+    # Run test
+    cluster_manager._handle_successfully_launched_nodes(slurm_nodes, private_ip_to_instance_map)
+    # Assert calls
+    mock_get_online_nodes_types.assert_called_with(slurm_nodes, private_ip_to_instance_map)
+    assert_that(cluster_manager._bootstrap_failure_nodes_types).is_equal_to(expected_bootstrap_failure_nodes_types)
+    assert_that(mock_reset_partition_failure_count.call_count).is_equal_to(len(successful_launched_partitions))
```
rexcsn, line 2261: Can we also assert that `partition_failure_count` map is as expected after running the function?
Same comments applies to `test_handle_bootstrap_failure_nodes` and any function that will update  `partition_failure_count`.

tests/slurm_plugin/test_clustermgtd.py
```diff
@@ -2112,6 +2170,794 @@ def test_update_status(self, compute_fleet_status_manager, put_item_response, ex
...
+            {
+                "ip-1": EC2Instance("id-1", "ip-1", "hostname", datetime(2020, 1, 1, 0, 0, 0)),
+            },
+            True,
+            {"queue1-st-c5xlarge-1"},
+            "Node bootstrap error: Replacement timeout expires for node in replacement",
+            True,
+        ),
+        (
+            SlurmNode("queue1-dy-c5xlarge-1", "ip-1", "hostname", "#MIXED", "queue1"),
```
rexcsn, line 2332: Let's try to use real slurm states. Same comment applies to all test cases in this test

tests/slurm_plugin/test_clustermgtd.py
```diff
@@ -2112,6 +2170,794 @@ def test_update_status(self, compute_fleet_status_manager, put_item_response, ex
...
+            SlurmNode("queue1-st-c5xlarge-1", "ip-1", "hostname", "DOWN*+CLOUD", "queue1"),
+            {
+                "ip-1": EC2Instance("id-1", "ip-1", "hostname", datetime(2020, 1, 1, 0, 0, 0)),
+                "ip-2": EC2Instance("id-2", "ip-2", "hostname", "launch_time"),
+            },
+            False,
+            {"some_node_in_replacement"},
+            None,
+            False,
+        ),
```
rexcsn, line 2385: These 2 test cases looks identical.
If we are testing run-instance failures, ip of the node should not be in list of instances

tests/slurm_plugin/test_clustermgtd.py
```diff
@@ -2112,6 +2170,794 @@ def test_update_status(self, compute_fleet_status_manager, put_item_response, ex
...
+    ids=[
+        "static_self_terminate",
+        "static_timeout",
+        "dynamic_self_terminate",
+        "dynamic_timeout",
+        "dynamic_runinstance",
+        "static_runinstance",
+        "static_joined_cluster",
+        "dynamic_reset_incorrect",
+    ],
```
rexcsn, line 2407: Can we try to test some "normal" down nodes and make sure they are not considered as bootstrap failure nodes? For example, dynamic nodes in `DRAIN+CLOUD`or `DOWN+CLOUD+POWERING_DOWN`

tests/slurm_plugin/test_clustermgtd.py
```diff
@@ -2112,6 +2170,794 @@ def test_update_status(self, compute_fleet_status_manager, put_item_response, ex
...
+        ),
+        (
+            [SlurmNode("queue1-st-c5xlarge-1", "ip-1", "hostname", "some_state", "queue1,queue2")],
+            {},
+            {"queue1": 1, "queue2": 1},
+        ),
+    ],
+)
+@pytest.mark.usefixtures("initialize_instance_manager_mock", "initialize_compute_fleet_status_manager_mock")
+def test_increase_partitions_protected_failure_count(nodes, initial_map, expected_map, mocker):
```
rexcsn, line 2468: This function is very straightforward and just increases the count, so we will not get much value by testing it alone. We should test in the caller that partition failure count is as expected when this function is called instead

src/slurm_plugin/clustermgtd.py
```diff
@@ -766,6 +834,58 @@ def _is_node_healthy(self, node, private_ip_to_instance_map):
...
+        """
+        if node.is_static:
+            # Node is currently in replacement and no backing instance
+            if (
+                not ClusterManager._is_backing_instance_valid(
+                    node, instance_ips_in_cluster=list(private_ip_to_instance_map.keys())
+                )
+                and node.name in self._static_nodes_in_replacement
+            ):
+                log.warning("Node bootstrap error: Node is currently in replacement and no backing instance")
```
demartinofra, line 853: Shall we add an information of what node you are referring to for all the log messages in this function?

src/slurm_plugin/clustermgtd.py
```diff
@@ -904,6 +1030,111 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+        if bootstrap_failure_nodes:
+            log.warning("Found the following bootstrap failure nodes: %s", print_with_count(bootstrap_failure_nodes))
+            self._handle_bootstrap_failure_nodes(bootstrap_failure_nodes)
+
+        # Enter protected mode
+        # We will put a partition into inactive state only if the partition satisfies the following:
+        # Partition is not INACTIVE
+        # Partition bootstrap failure count above threshold
+        # Partition does not have job running
+        logging.info(f"Partitions_protected_failure_count_map is {self._partitions_protected_failure_count_map}")
```
demartinofra, line 1111: let's use log and not logging

demartinofra: also let's try to have meaningful log messaged for customers

src/slurm_plugin/clustermgtd.py
```diff
@@ -904,6 +1030,111 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+    def _is_protected_mode_enabled(self):
+        """When protected_failure_count is set to -1, disable protected mode."""
+        if self._config.protected_failure_count == -1:
+            return False
+        return True
+
+    def _handle_protected_mode_process(self, slurm_nodes, partitions, bootstrap_failure_nodes, healthy_nodes):
+        """Handle the process of entering protected mode."""
+        # Handle successfully launched nodes
+        nodename_to_slurm_nodes_map = {node.name: node for node in slurm_nodes}
```
demartinofra, line 1098: You don't need to always initialize this data structure, cause it is only used in the if statement below.
As a general recommendation all these data-structures can contain thousands of nodes based on the potential size of the cluster, which means we should minimize doing all these transformations or copies if possible.

src/slurm_plugin/clustermgtd.py
```diff
@@ -904,6 +1030,111 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
+        ]
+        if partitions_above_threshold:
+            log.info(f"Bootstrap failure partitions: {partitions_above_threshold}")
+            partitions_to_disable = ClusterManager._get_partitions_to_disable(
+                nodename_to_slurm_nodes_map, partitions_above_threshold, partitions
+            )
+            self._enter_protected_mode(partitions_to_disable)
+
+    @staticmethod
+    def _filter_partitions_with_active_jobs(nodename_to_slurm_nodes_map, partitions):
```
demartinofra, line 1125: Dropping this comment here but it is more of a general consideration: the ClusterManger class is becoming too complex and implementing functionalities that could be encapsulated in separate classes. For example it could make sense to have a Partition class which includes as an attribute the list of nodes and that exposes a set of functionalities such has the has_jobs so that you could simply do partition.has_jobs(). The same applies to Node related operations where you could actually have 2 concrete classes for dynamic and static nodes which will allow you to remove all those if/else statements in the code. 
Other than simplifying the code this would allow to reduce the number of times we have to transform, copy or pas s around all the various data structures the ClusterManager is currently using and that have a linear dependency on the number of nodes. Ideally we could identify the following entities: Scheduler, Partition, Node (StaticNode, DynamicNode), EC2Instance that have the following dependencies:
* Scheduler->List[Partition], Scheduler->List[Nodes]
* Partition->List[Nodes]
* Node->EC2Instance

Now I can't asking to refactor the code as part of this PR to achieve all of this but if we could think of some reasonable enhancement to start extracting some of the functionalities outside of the ClusterManager I think it is the right move.

src/slurm_plugin/clustermgtd.py
```diff
@@ -904,6 +1030,111 @@ def _terminate_orphaned_instances(self, cluster_instances, ips_used_by_slurm):
...
 
+    def _enter_protected_mode(self, partitions_to_disable):
+        """Entering protected mode if no active running job in queue."""
+        if partitions_to_disable:
+            # Change compute fleet status to protected
+            if not self._compute_fleet_status == ComputeFleetStatus.PROTECTED:
+                log.info("Place Cluster into protected mode")
+                self._update_compute_fleet_status(ComputeFleetStatus.PROTECTED)
+            # Place partitions into inactive
+            log.info(f"Placing bootstrap failure partitions to INACTIVE: {partitions_to_disable}")
```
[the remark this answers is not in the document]
chenwany: Changed to lazy logging

src/slurm_plugin/clustermgtd.py
chenwany: * Changed into log
* Change the log messages
* Currently the log will for each node if the node have bootstrap failure. `"Node bootstrap error: Node %s is currently in replacement and no backing instance` like we do for unhealthy nodes. In other places, they are log for a set of nodes or partitions

src/common/schedulers/slurm_commands.py
```diff
@@ -440,3 +469,29 @@ def _parse_nodes_info(slurm_node_info):
...
+def get_nodes_type(nodes):
+    """Get node type given nodename. Node type format: NodeType(instancetype, queue_name)."""
+    nodes_types = set()
+    for node in nodes:
+        queue_name, _, instance_name = parse_nodename(node.name)
+        nodes_types.add(NodeType(instance_name, queue_name))
+    return nodes_types
+
+
+def get_inactive_partitions_name(partitions):
```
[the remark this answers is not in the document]
chenwany: Fixed

src/common/schedulers/slurm_commands.py
```diff
@@ -440,3 +469,29 @@ def _parse_nodes_info(slurm_node_info):
...
+            inactive_partitions_name.append(partition_name)
+    return inactive_partitions_name
+
+
+def retrieve_partitions_from_node_types(node_types):
+    """Retrieve partitions for node types."""
+    partitions = set()
+    for node_type in node_types:
+        partitions.add(node_type.partition)
+    return set(partitions)
```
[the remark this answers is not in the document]
chenwany: Good idea! Done

src/slurm_plugin/clustermgtd.py
```diff
@@ -766,6 +834,58 @@ def _is_node_healthy(self, node, private_ip_to_instance_map):
...
+                and time_is_up(
+                    backing_instance.launch_time, self._current_time, grace_time=self._config.node_replacement_timeout
+                )
+            ):
+                log.warning("Node bootstrap error: Replacement timeout expires for node in replacement")
+                return True
+
+        else:
+            # no backing instance + [working state]# in node state
+            if node.is_configuring_job() and not ClusterManager._is_backing_instance_valid(
```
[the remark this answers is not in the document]
chenwany: Have been discussed during our sync, Rex already kept a note for that.

src/slurm_plugin/clustermgtd.py
rexcsn: Will address refactoring in future PR, resolving for now

src/slurm_plugin/clustermgtd.py
```diff
@@ -389,13 +437,19 @@ def manage_cluster(self):
...
                 log.info("Retrieving nodes info from the scheduler")
-                active_nodes, inactive_nodes = self._get_node_info_from_partition()
+                partitions = {
+                    partition.name: partition for partition in ClusterManager._get_partition_info_with_retry()
+                }
+                log.debug("Partitions: %s", partitions)
+                active_nodes, inactive_nodes = self._get_node_info_from_partition(partitions)
                 log.debug("Current active slurm nodes in scheduler: %s", active_nodes)
                 log.debug("Current inactive slurm nodes in scheduler: %s", inactive_nodes)
+                self._inactive_partitions = get_inactive_partition_names(partitions)
```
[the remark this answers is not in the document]
chenwany: Keep a note

src/slurm_plugin/clustermgtd.py
```diff
@@ -660,6 +713,15 @@ def _handle_health_check(
                             unhealthy_node.nodeaddr,
                         )
                         self._static_nodes_in_replacement.remove(unhealthy_node.name)
+                    if self._is_protected_mode_enabled() and unhealthy_node.is_powering_up():
```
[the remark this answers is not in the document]
chenwany: Added

tests/slurm_plugin/test_clustermgtd.py
```diff
@@ -2112,6 +2170,794 @@ def test_update_status(self, compute_fleet_status_manager, put_item_response, ex
...
+        ),
+        (
+            SlurmNode("queue1-dy-c5xlarge-1", "ip-1", "hostname", "some_state", "queue1"),
+            set(),
+            {NodeType("queue1-c5xlarge", "queue1")},
+        ),
+    ],
+)
+@pytest.mark.usefixtures("initialize_instance_manager_mock", "initialize_compute_fleet_status_manager_mock")
+def test_handle_health_check_bootsrap_failure_node(
```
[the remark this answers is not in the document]
chenwany: Right, renamed the function

tests/slurm_plugin/test_clustermgtd.py
```diff
@@ -2112,6 +2170,794 @@ def test_update_status(self, compute_fleet_status_manager, put_item_response, ex
...
+                "ip-2": EC2Instance("id-2", "ip-2", "hostname", "launch_time"),
+            },
+            [],
+            [],
+            {"queue1": SlurmPartition("queue1", "queue1-st-c5xlarge-1, queue1-st-c5xlarge-2", "states")},
+        ),
+    ],
+)
+@pytest.mark.usefixtures("initialize_instance_manager_mock", "initialize_compute_fleet_status_manager_mock")
+def test_handle_protected_mode_process(
```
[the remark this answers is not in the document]
chenwany: Add a reminder for refactoring

tests/slurm_plugin/test_clustermgtd.py
```diff
@@ -595,25 +593,36 @@ def test_fail_scheduled_events_health_check(instance_health_state, expected_resu
...
             {"some_node_in_replacement1"},
             {"some_node_in_replacement1"},
+            SlurmNode("queue1-st-c5xlarge-2", "ip-2", "host-2", "ALLOCATED#+CLOUD", "queue1"),
         ),
         (
             ClusterManager.HealthCheckTypes.ec2_health,
-            [True, False],
-            [False, True],
-            ["queue1-st-c5xlarge-1"],
-            {"some_node_in_replacement1", "queue1-st-c5xlarge-1"},
```
rexcsn, line 613: Are we reducing test coverage here? We were supposed to test the case that if a node in instance replacement fails health check, we take it out of node in replacement so it will be replaced later.

src/slurm_plugin/clustermgtd.py
```diff
@@ -660,6 +713,24 @@ def _handle_health_check(
                             unhealthy_node.nodeaddr,
                         )
                         self._static_nodes_in_replacement.remove(unhealthy_node.name)
+                        if self._is_protected_mode_enabled():
+                            self._handle_bootstrap_failure_nodes([unhealthy_node])
+                            log.warning(
+                                "Node bootstrap error: Node %s(%s) failed health check during replacement, "
+                                f"increase partition failure count: {self._partitions_protected_failure_count_map}, "
```
rexcsn, line 720: There are still f string in the logging, let's try to use `%s`.

src/slurm_plugin/clustermgtd.py
```diff
@@ -660,6 +713,24 @@ def _handle_health_check(
                             unhealthy_node.nodeaddr,
                         )
                         self._static_nodes_in_replacement.remove(unhealthy_node.name)
+                        if self._is_protected_mode_enabled():
+                            self._handle_bootstrap_failure_nodes([unhealthy_node])
+                            log.warning(
+                                "Node bootstrap error: Node %s(%s) failed health check during replacement, "
```
rexcsn, line 719: Minor: `SlurmNode` object has a defined `__str__` function, so we can probably just do `"%s", node` directly as you are doing in other places and the class will print `f"{self.name}({self.nodeaddr})"` automatically.

tests/slurm_plugin/test_clustermgtd.py
chenwany: Right! I didn't realize that, add it back, also add the test for case static node fail bootstrap

tests/slurm_plugin/test_clustermgtd.py
```diff
@@ -595,25 +593,36 @@ def test_fail_scheduled_events_health_check(instance_health_state, expected_resu
...
+            [True, False, True, False],
+            [False, True, True, False],
+            ["queue1-st-c5xlarge-1", "queue1-st-c5xlarge-3"],
             {"some_node_in_replacement1"},
+            {"some_node_in_replacement1"},
+            SlurmNode("queue1-st-c5xlarge-1", "ip-1", "host-1", "ALLOCATED#+CLOUD", "queue1"),
```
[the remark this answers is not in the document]
chenwany: Changed

src/slurm_plugin/clustermgtd.py
chenwany: Changed all to %s

src/slurm_plugin/clustermgtd.py
chenwany: Done

rexcsn (approved):

PR #309, opened 2021-06-25
Use subprocess based method to read shared files

rexcsn:
* Avoid using `open` to read NFS shared files, because when NFS is not available(head node down) `open` call will hang forever.
* Use `subprocess.run`, which has built in timeout functionality, instead.
* This will allow computemgtd to run even if NFS directories are not available

### DIFF
--- a/src/slurm_plugin/common.py
+++ b/src/slurm_plugin/common.py
@@ -21,7 +21,7 @@ import boto3
 from botocore.config import Config
 from botocore.exceptions import ClientError
 from common.schedulers.slurm_commands import InvalidNodenameError, parse_nodename, update_nodes
-from common.utils import grouper
+from common.utils import check_command_output, grouper
 
 CONFIG_FILE_DIR = "/etc/parallelcluster/slurm_plugin"
 EC2Instance = collections.namedtuple("EC2Instance", ["id", "private_ip", "hostname", "launch_time"])
@@ -49,6 +49,7 @@ BOTO3_PAGINATION_PAGE_SIZE = 1000
 # timestamp used by clustermgtd and computemgtd should be in default ISO format
 # YYYY-MM-DDTHH:MM:SS.ffffff+HH:MM[:SS[.ffffff]]
 TIMESTAMP_FORMAT = "%Y-%m-%d %H:%M:%S.%f%z"
+DEFAULT_COMMAND_TIMEOUT = 30
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


src/slurm_plugin/common.py
```diff
@@ -430,16 +431,18 @@ def retrieve_instance_type_mapping(file_path):
...
 
-def _get_clustermgtd_heartbeat(clustermgtd_heartbeat_file_path):
+def get_clustermgtd_heartbeat(clustermgtd_heartbeat_file_path):
     """Get clustermgtd's last heartbeat."""
-    with open(clustermgtd_heartbeat_file_path, "r") as timestamp_file:
-        # Note: heartbeat must be written with datetime.strftime to convert localized datetime into str
-        # datetime.strptime will not work with str(datetime)
-        # Example timestamp written to heartbeat file: 2020-07-30 19:34:02.613338+00:00
-        return datetime.strptime(timestamp_file.read().strip(), TIMESTAMP_FORMAT)
+    heartbeat = check_command_output(
```
demartinofra, line 436: let's use the last line from the stdout and also do a strip() in case there is some additional output prepended when using the shell.

src/slurm_plugin/computemgtd.py
```diff
@@ -100,10 +112,11 @@ def _read_nodename_from_file(nodename_file_path):
         """Read self nodename from a file."""
         try:
             log.info("Reading self nodename from %s", nodename_file_path)
-            with open(nodename_file_path, "r") as nodename_file:
-                nodename = nodename_file.read()
+            nodename = check_command_output(
```
demartinofra, line 115: add comment here too

src/slurm_plugin/common.py
demartinofra: Also do we need shell to be set to True? same everywhere else

src/slurm_plugin/computemgtd.py
```diff
@@ -53,14 +61,18 @@ def __repr__(self):
...
+    @log_exception(log, "reading computemgtd config", catch_exception=Exception, raise_on_error=True)
     def _get_config(self, config_file_path):
         """Get computemgtd configuration."""
         log.info("Reading %s", config_file_path)
         config = ConfigParser()
         try:
-            config.read_file(open(config_file_path, "r"))
-        except IOError:
+            # Use subprocess based method with timeout to read shared file to prevent hanging when NFS is down
+            config_content = check_command_output(
```
demartinofra, line 71: here again I'm concerned that when running this command through the shell we get some unexpected output that breaks our logic (e.g. is the shell profile sourced?can this add unexpected output?)

tilne (approved):

demartinofra (approved):
LGTM. we only need to understand why tests for python 3.6 are failing

demartinofra (approved):

PR #318, opened 2021-07-21
Write temp computemgtd config in slurm plugin dir

tilne:
One of the changes made as part of 22c7e21 was to make a local copy of
the computemgtd config. This local copy was written to a directory owned
by root, which was causing permission issues now that computemgtd is run
as a non-root user.

Signed-off-by: Tim Lane <tilne@amazon.com>

### DIFF
--- a/src/slurm_plugin/computemgtd.py
+++ b/src/slurm_plugin/computemgtd.py
@@ -17,6 +17,7 @@ from configparser import ConfigParser
 from datetime import datetime, timezone
 from logging.config import fileConfig
 from subprocess import CalledProcessError
+from tempfile import mkstemp
 
 from botocore.config import Config
 from common.schedulers.slurm_commands import get_nodes_info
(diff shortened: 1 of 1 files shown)


src/slurm_plugin/computemgtd.py
```diff
@@ -66,12 +67,13 @@ def _get_config(self, config_file_path):
         config = ConfigParser()
         try:
             # Use subprocess based method to copy shared file to local to prevent hanging when NFS is down
+            tmp_config_path = f"{SLURM_PLUGIN_DIR}/.computemgtd_config.local"
```
demartinofra, line 70: the dir used as temporary needs to be on the local fs and not on an nfs share otherwise we don't address the NFS hanging problem

tilne: ah, makes sense. fixing now.

tilne: done.

demartinofra (approved):

PR #322, opened 2021-08-03
Force nodes to down and power_save when stopping the cluster

demartinofra:
If a cluster is stopped while a node is powering-up (alloc#-idle#),
node is kept in the powering-up state on cluster start.
This makes the node unavailable for the entire ResumeTimeout which is 60 minutes.
Slurm is ignoring the transition to power_down if we don't put the node to down first.


enrico-usai:
Closed in favour of https://github.com/aws/aws-parallelcluster-node/pull/325

PR #350, opened 2021-11-03
Upgrade to slurm21.08

chenwany:
* when `set_nodes_down_and_power_save`, update node state to `power_down_force` instead of setting nodes to DOWN and then POWER_DOWN
    * `set_nodes_down_and_power_sav`e is used when terminate unhealthy nodes and reset nodes in `INACTIVE` partition. 
    * `POWER_DOWN_FORCE` cancels all jobs on the node and suspends the node immediately, placing the node in `IDLE` and `POWER_DOWN (!)` state and then `POWERING_DOWN (%)` state. 
    * Previously, we set Node to `DOWN` then `POWER_DOWN`, node will be `DOWN` state, then in `IDLE` and `POWER_DOWN (!)` state, and then `POWERING_DOWN (%)` state. These two steps are equivalent to `POWER_DOWN_FORCE`

* When Slurm Resume Failed, the state is `DOWN+CLOUD+POWERED_DOWN+NOT_RESPONDING`
    * Previously in slurm 20, the state is `DOWN*+CLOUD+POWER`
* State reported by scontrol does not contain anymore symbols
    * `*` in slurm20 becomes `Not_RESPONDING` in slurm21
    * `#` in slurm20 becomes `POWERING_UP` in slurm21
    * sinfo: `down*` scontrol slurm20: `DOWN*+CLOUD` →scontrol slurm21: `DOWN+CLOUD+NOT_RESPONDING` 
    * sinfo: `mix#` scontrol slurm20: `MIXED#+CLOUD` →scontrol slurm21: `MIXED+CLOUD+NOT_RESPONDING+POWERING_UP`
    * sinfo: `mix~` scontrol slurm20: `MIXED+CLOUD+POWER` →scontrol slurm21: `MIXED+CLOUD+POWERED_DOWN`
    * sinfo: `idle~` scontrol slurm20: `IDLE+CLOUD+POWER`  scontrol slurm21: `IDLE+CLOUD+POWERED_DOWN`
    * sinfo: `idle` scontrol slurm20: `IDLE+CLOUD`  scontrol slurm21: `IDLE+CLOUD`
    * sinfo:`idle%` scontrol same in slurm20 and 21: `IDLE+CLOUD+POWERING_DOWN`
* Consider Nodes with `IDLE+CLOUD+NOT_RESPONDING+POWERING_UP` node without backing instance as bootstrap failure nodes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ This file is used to list changes made in each version of the aws-parallelcluste
 - Add possibility to override EC2 RunInstances parameters for instances launched in a Slurm cluster.
 
 **CHANGES**
-- Upgrade Slurm to version 21.08.3.
+- Update Slurm plugin to support version 21.08.
 
 3.0.2
 ------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ This file is used to list changes made in each version of the aws-parallelcluste
 **ENHANCEMENTS**
 - Add possibility to override EC2 RunInstances parameters for instances launched in a Slurm cluster.
 
+**CHANGES**
+- Upgrade Slurm to version 21.08.3.
+
 3.0.2
 ------
 
--- a/src/common/schedulers/slurm_commands.py
+++ b/src/common/schedulers/slurm_commands.py
@@ -140,7 +140,7 @@ def update_all_partitions(state, reset_node_addrs_hostname):
                 log.info(f"Setting partition {part.name} state from {part.state} to {state}")
                 if reset_node_addrs_hostname:
                     log.info(f"Resetting partition nodes {part.nodenames}")
-                    set_nodes_down_and_power_save(part.nodenames, reason="stopping cluster")
+                    set_nodes_power_down(part.nodenames, reason="stopping cluster")
                 partition_to_update.append(part.name)
         succeeded_partitions = update_partitions(partition_to_update, state)
         return succeeded_partitions == partition_to_update
@@ -194,6 +194,7 @@ def set_nodes_drain(nodes, reason):
     update_nodes(nodes, state="drain", reason=reason)
 
 
+@retry(stop_max_attempt_number=3, wait_fixed=1500)
 def set_nodes_power_down(nodes, reason=None):
     """Place slurm node into power_down state and reset nodeaddr/nodehostname."""
     reset_nodes(nodes=nodes, state="power_down_force", reason=reason, raise_on_error=True)
@@ -222,16 +223,6 @@ def set_nodes_idle(nodes, reason=None, reset_node_addrs_hostname=False):
         reset_nodes(nodes, state="resume", reason=reason, raise_on_error=False)
     else:
         update_nodes(nodes=nodes, state="resume", reason=reason, raise_on_error=False)
-
-
-@retry(stop_max_attempt_number=3, wait_fixed=1500)
-def set_nodes_down_and_power_save(node_list, reason):
-    """
-    Set slurm nodes into down -> power_down.
-
-    This is the standard failure recovery procedure to reset a CLOUD node.
-    """
-    set_nodes_power_down(node_list, reason=reason)
 
 
 def get_nodes_info(nodes="", command_timeout=DEFAULT_GET_INFO_COMMAND_TIMEOUT):
--- a/src/common/schedulers/slurm_commands.py
+++ b/src/common/schedulers/slurm_commands.py
@@ -196,7 +196,7 @@ def set_nodes_drain(nodes, reason):
 
 def set_nodes_power_down(nodes, reason=None):
     """Place slurm node into power_down state and reset nodeaddr/nodehostname."""
-    reset_nodes(nodes=nodes, state="power_down", reason=reason, raise_on_error=True)
+    reset_nodes(nodes=nodes, state="power_down_force", reason=reason, raise_on_error=True)
 
 
 def reset_nodes(nodes, state=None, reason=None, raise_on_error=False):
@@ -231,7 +231,8 @@ def set_nodes_down_and_power_save(node_list, reason):
 
     This is the standard failure recovery procedure to reset a CLOUD node.
     """
-    set_nodes_down(node_list, reason=reason)
+    # TODO: check if still required especially when partitions are INACTIVE (stop)
+    # set_nodes_down(node_list, reason=reason)
     set_nodes_power_down(node_list, reason=reason)
 
 
--- a/src/common/schedulers/slurm_commands.py
+++ b/src/common/schedulers/slurm_commands.py
@@ -231,8 +231,6 @@ def set_nodes_down_and_power_save(node_list, reason):
 
     This is the standard failure recovery procedure to reset a CLOUD node.
     """
-    # TODO: check if still required especially when partitions are INACTIVE (stop)
-    # set_nodes_down(node_list, reason=reason)
     set_nodes_power_down(node_list, reason=reason)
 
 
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ This file is used to list changes made in each version of the aws-parallelcluste
 **ENHANCEMENTS**
 - Add possibility to override EC2 RunInstances parameters for instances launched in a Slurm cluster.
 
+**CHANGES**
+- Upgrade Slurm to version 21.08.3.
```
demartinofra: What about: `Update Slurm plugin to support version 21.08`?

src/common/schedulers/slurm_commands.py
```diff
@@ -194,7 +194,7 @@ def set_nodes_drain(nodes, reason):
 
 def set_nodes_power_down(nodes, reason=None):
```
demartinofra, line 198: why don't we keep a single function `set_nodes_power_down` and remove `set_nodes_down_and_power_save` since they are both doing the same? We might want to add the retry annotation on top of `set_nodes_power_down`

src/slurm_plugin/slurm_resources.py
```diff
@@ -148,21 +148,25 @@ def is_drained(self):
...
     def is_power(self):
         """Check if slurm node is in power state."""
-        return self.SLURM_SCONTROL_POWER_STATE == self.state
+        states = self.state.split("+")
+        return all(power_state in states for power_state in self.SLURM_SCONTROL_POWER_STATE.split("+"))
 
     def is_down(self):
         """Check if slurm node is in a down state."""
-        return self.SLURM_SCONTROL_DOWN_STATE in self.state and not self.is_powering_down()
+        states = self.state.split("+")
```
demartinofra: shall we do this split just once in the constructor and assign the result to a set? we could call the string `state_string` and the set `states`? up to you.

Also if you want in some other places like (is_online) you could check the state in the set rather than string which is more performant.

demartinofra (commented):
Good job!

src/common/schedulers/slurm_commands.py
chenwany: Removed `set_nodes_down_and_power_save`

src/slurm_plugin/slurm_resources.py
chenwany: Good idea! Add `self.state_string` and self.states to `SlurmNode`, change other places use `in self.state_string` to `in self.states`

CHANGELOG.md
chenwany: Changed

src/slurm_plugin/slurm_resources.py
```diff
@@ -149,48 +150,57 @@ def is_drained(self):
...
         """Check if slurm node is in powering down state."""
-        return self.SLURM_SCONTROL_POWERING_DOWN_STATE in self.state
+        return any(
+            powering_down_state in self.states for powering_down_state in self.SLURM_SCONTROL_POWERING_DOWN_STATES
+        )
 
     def is_power(self):
         """Check if slurm node is in power state."""
-        return self.SLURM_SCONTROL_POWER_STATE == self.state
+        return all(power_state in self.states for power_state in self.SLURM_SCONTROL_POWER_STATE.split("+"))
```
demartinofra: isn't there a problem in case self.states contains additional states such as NOT_RESPONDING?
either we leave the string match here or we check for set equality something like set(self.SLURM_SCONTROL_POWER_STATE.split("+")) == self.states (if it works)

src/slurm_plugin/slurm_resources.py
```diff
@@ -107,22 +107,23 @@ def __eq__(self, other):
...
+    SLURM_SCONTROL_RESUME_FAILED_STATE = "DOWN+CLOUD+POWERED_DOWN+NOT_RESPONDING"
 
     def __init__(self, name, nodeaddr, nodehostname, state, partitions=None, instance=None):
         """Initialize slurm node with attributes."""
         self.name = name
         self.nodeaddr = nodeaddr
         self.nodehostname = nodehostname
-        self.state = state
+        self.state_string = state
+        self.states = state.split("+")
```
demartinofra: shall we force this to be a set and not a list otherwise you get O(n) and not O(1)

src/slurm_plugin/slurm_resources.py
```diff
@@ -149,48 +150,57 @@ def is_drained(self):
...
+        )
 
     def is_up(self):
         """Check if slurm node is in a healthy state."""
         return not self._is_drain() and not self.is_down() and not self.is_powering_down()
 
     def is_powering_up(self):
         """Check if slurm node is in powering up state."""
-        return self.SLURM_SCONTROL_POWER_UP_STATE in self.state
+        return any(power_up_state in self.states for power_up_state in self.SLURM_SCONTROL_POWER_UP_STATE)
```
demartinofra: optional: if you want to avoid creating sets with a single element you could just do `self.SLURM_SCONTROL_POWER_UP_STATE in self.states` and remove the iteration

src/slurm_plugin/slurm_resources.py
```diff
@@ -149,48 +150,57 @@ def is_drained(self):
...
         """Check if slurm node is running a job but not in configuring job state."""
         return not self.is_powering_up() and self.has_job() and not self.is_power_with_job()
 
     def is_resume_failed(self):
         """Check if node resume timeout expires."""
-        return self.state == self.SLURM_SCONTROL_RESUME_FAILED_STATE
+        return self.state_string == self.SLURM_SCONTROL_RESUME_FAILED_STATE
+
+    def is_poweing_up_idle(self):
+        """Check if node is in IDEL# state."""
```
demartinofra: typo: IDLE

src/slurm_plugin/slurm_resources.py
chenwany: Change back to check `self.SLURM_SCONTROL_POWER_STATE == self.state`

src/slurm_plugin/slurm_resources.py
chenwany: Yes, change to self.states = set(state.split("+"))

src/slurm_plugin/slurm_resources.py
chenwany: Removed creating sets for all single element and changed to check `self.SLURM_SCONTROL_POWER_UP_STATE in self.states`

src/slurm_plugin/slurm_resources.py
chenwany: Fixed

src/slurm_plugin/slurm_resources.py
```diff
@@ -136,61 +137,65 @@ def is_nodeaddr_set(self):
...
+        return self.state_string == self.SLURM_SCONTROL_POWER_WITH_JOB_STATE
 
     def is_running_job(self):
         """Check if slurm node is running a job but not in configuring job state."""
         return not self.is_powering_up() and self.has_job() and not self.is_power_with_job()
 
     def is_resume_failed(self):
         """Check if node resume timeout expires."""
-        return self.state == self.SLURM_SCONTROL_RESUME_FAILED_STATE
+        return self.state_string == self.SLURM_SCONTROL_RESUME_FAILED_STATE
```
demartinofra: one last comment, sorry for not calling out this earlier. Probably it's not even an actual problem at the moment but in case Slurm does not always guarantee the order in the state string we might end up with something like `DOWN+NOT_RESPONDING+CLOUD+POWERED_DOWN` which represents that same state but will fail our check.
To make this logic more robust we could represent the state string as a set and then check for sets equality.
```
SLURM_SCONTROL_RESUME_FAILED_STATE = {DOWN,CLOUD,POWERED_DOWN,NOT_RESPONDING}

return self.stats == self.SLURM_SCONTROL_RESUME_FAILED_STATE
```

Same applies to other few places

chenwany: Thank you! Changed to for SLURM_SCONTROL_RESUME_FAILED_STATE, SLURM_SCONTROL_POWER_WITH_JOB_STATE and SLURM_SCONTROL_POWER_STATE

demartinofra (approved):

PR #369, opened 2021-11-26
Resume POWERING_DOWN nodes after enabling partitions 

demartinofra:
This makes CLOUD nodes available to be powered up immediately after partitions are enabled. 
Without this a node will stay in POWERING_DOWN for the SuspendTimeout (120secs). Also we noticed that in some cases Slurm is not even able to properly reset a node. This change should address such edge case too.

Note that in case you try to set nodes that are not in POWERING_DOWN to RESUME state Slurm will complain. 

### DIFF
--- a/src/common/schedulers/slurm_commands.py
+++ b/src/common/schedulers/slurm_commands.py
@@ -258,6 +258,13 @@ def get_partition_info(command_timeout=DEFAULT_GET_INFO_COMMAND_TIMEOUT, get_all
     ]
 
 
+def resume_powering_down_nodes():
+    """Resume nodes that are powering_down so that are set in power state right away."""
+    log.info("Resuming powering down nodes.")
+    powering_down_nodes = _get_slurm_nodes(states="powering_down")
+    update_nodes(nodes=powering_down_nodes, state="resume", raise_on_error=False)
+
+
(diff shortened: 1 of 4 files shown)


src/common/schedulers/slurm_commands.py
```diff
@@ -258,6 +258,13 @@ def get_partition_info(command_timeout=DEFAULT_GET_INFO_COMMAND_TIMEOUT, get_all
     ]
 
 
+def resume_powering_down_nodes():
+    """Resume nodes that are powering_down so that are set in power state right away."""
+    log.info("Resuming powering down nodes.")
+    powering_down_nodes = _get_slurm_nodes(states="powering_down")
+    update_nodes(nodes=powering_down_nodes, state="resume", raise_on_error=False)
```
chenwany, line 265: Do we want to reset nodeaddr and nodehostname here in case they are not reset and add a reason for the command: `reset_nodes(nodes=powering_down_nodes, state="resume", reason="Resuming powering down nodes.", raise_on_error=False)`
Although in the cases we saw about nodes in `IDLE+CLOUD+POWERING_DOWN`, nodeaddr and nodehostname are already reset

chenwany (approved):

PR #402, opened 2022-04-21
Handle special cases of IDLE~ and DOWN~ nodes

chenwany:
Sometimes nodes reset by power_down_force are in (IDLE~) IDLE+CLOUD+POWER_DOWN+POWERED_DOWN state.

Treat IDLE+CLOUD+POWER_DOWN+POWERED_DOWN node as power node.
Treat DOWN~ (DOWN+CLOUD+POWER_DOWN+POWERED_DOWN+NOT_RESPONDING) nodes as unhealthy nodes

Also we don't need to check ice nodes in unhealthy nodes anymore after we separate ices nodes from unhealthy nodes

Signed-off-by: chenwany <chenwany@amazon.com>

### Description of changes
* Describe *what* you're changing and *why* you're doing these changes.

### Tests
* Add unit test case

### References

### Checklist

### DIFF
--- a/src/slurm_plugin/slurm_resources.py
+++ b/src/slurm_plugin/slurm_resources.py
@@ -112,12 +112,15 @@ class SlurmNode(metaclass=ABCMeta):
     SLURM_SCONTROL_IDLE_STATE = "IDLE"
     SLURM_SCONTROL_DOWN_STATE = "DOWN"
     SLURM_SCONTROL_DRAIN_STATE = "DRAIN"
-    SLURM_SCONTROL_POWERING_DOWN_STATES = {"POWERING_DOWN", "POWER_DOWN"}
-    SLURM_SCONTROL_POWER_STATE = {"IDLE", "CLOUD", "POWERED_DOWN"}
+    SLURM_SCONTROL_POWERED_STATE = "POWERED_DOWN"
+    SLURM_SCONTROL_POWERING_DOWN_STATES = "POWERING_DOWN"
+    SLURM_SCONTROL_POWER_DOWN_STATE = "POWER_DOWN"
     SLURM_SCONTROL_POWER_UP_STATE = "POWERING_UP"
     SLURM_SCONTROL_ONLINE_STATES = {"IDLE+CLOUD", "MIXED+CLOUD", "ALLOCATED+CLOUD", "COMPLETING+CLOUD"}
     SLURM_SCONTROL_POWER_WITH_JOB_STATE = {"MIXED", "CLOUD", "POWERED_DOWN"}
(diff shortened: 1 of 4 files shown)


src/slurm_plugin/slurm_resources.py
```diff
@@ -165,13 +166,14 @@ def is_drained(self):
 
     def is_powering_down(self):
         """Check if slurm node is in powering down state."""
-        return any(
-            powering_down_state in self.states for powering_down_state in self.SLURM_SCONTROL_POWERING_DOWN_STATES
+        return (
+            any(powering_down_state in self.states for powering_down_state in self.SLURM_SCONTROL_POWERING_DOWN_STATES)
+            and self.SLURM_SCONTROL_POWERED_STATE not in self.states
         )
```
demartinofra: Don't we need to assume that a node in `IDLE+CLOUD+POWER_DOWN+POWERED_DOWN` is actually powering down?

src/slurm_plugin/slurm_resources.py
```diff
@@ -165,13 +166,14 @@ def is_drained(self):
...
 
     def is_power(self):
         """Check if slurm node is in power state."""
-        return self.SLURM_SCONTROL_POWER_STATE == self.states
+        return all(power_state in self.states for power_state in self.SLURM_SCONTROL_POWER_STATE)
```
demartinofra: is this condition safe? I'm wondering if you could find the node in a state that is "IDLE", "CLOUD", "POWERED_DOWN" + something else which is not POWER_DOWN.

Shall we add the explicit states to SLURM_SCONTROL_POWER_STATE ({{"IDLE", "CLOUD", "POWERED_DOWN"},{"IDLE", "CLOUD", "POWERED_DOWN", "POWER_DOWN"}}) and check something like `self.states in self.SLURM_SCONTROL_POWER_STATE`?

Also can we add a comment on top of SLURM_SCONTROL_POWERED_STATE saying something like: "Due to a bug in Slurm a powered down node can enter IDLE+CLOUD+POWER_DOWN+POWERED_DOWN state"

chenwany: Yes, I think it is more save to use [{"IDLE", "CLOUD", "POWERED_DOWN"},{"IDLE", "CLOUD", "POWERED_DOWN", "POWER_DOWN"}], changed

src/slurm_plugin/slurm_resources.py
chenwany: After our discussion, treat `POWERING_DOWN + POWERED_DOWN` as powering down node

demartinofra (approved):

PR #404, opened 2022-04-27
[develop] Manage manual power saving node state transitions

lukeseawalker:

### Description of changes
* Do not consider dynamic POWER_DOWN nodes as in powering down
  
  Powering down nodes are only the ones in POWERING_DOWN, because nodes in POWER_DOWN are still up and running. For example a node in POWER_DOWN can be in DRAINING, meaning that jobs are running and will continue to run until completion.

* Handle static nodes in POWERING_DOWN

  Static nodes can be in POWERING_DOWN, for example when setting a manual power save mode from outside.
This condition can happen during the replacement of the nodes following an update of the cluster.

* Do not consider node in DRAIN + POWER_DOWN as drained

  This to let Slurm transition the node from POWER_DOWN to POWERING_DOWN by itself

* Do not consider node in DOWN + POWER_DOWN as down
  
  This to let Slurm transition the node from POWER_DOWN to POWERING_DOWN and then to POWERED_DOWN by itself

### Tests
* Manually tested
* Integration tests will follow up in CLI code

### References
* CLI changes https://github.com/aws/aws-parallelcluster/pull/3976
* Cookbook changes https://github.com/aws/aws-parallelcluster-cookbook/pull/1422

### Checklist


src/slurm_plugin/slurm_resources.py
```diff
@@ -168,9 +167,7 @@ def is_drained(self):
 
     def is_powering_down(self):
         """Check if slurm node is in powering down state."""
-        return self.SLURM_SCONTROL_POWERING_DOWN_STATES in self.states or (
-            self.SLURM_SCONTROL_POWERED_STATE not in self.states and self.SLURM_SCONTROL_POWER_DOWN_STATE in self.states
-        )
+        return self.SLURM_SCONTROL_POWERING_DOWN_STATE in self.states
```
chenwany, line 179: is_powering_down() is also used in checking unhealthy nodes:
```
    def is_down(self):
        """Check if slurm node is in a down state."""
        return self.SLURM_SCONTROL_DOWN_STATE in self.states and not self.is_powering_down()
```
With this change, dynamic nodes state with (down+power_down) and (drain+power_down) will be treated as unhealthy dynamic node,  instances will be terminated and nodes wil be power_down_force by `_handle_unhealthy_dynamic_nodes`

Same for static nodes not in replacement
Should we also modify the is_state_healthy() function?

lukeseawalker: I'm going to add a commit to not consider the node in DRAIN+POWER_DOWN as drained, so to give Slurm time to transition the node from POWER_DOWN to POWERING_DOWN by itself.
Instead for the DOWN+POWER_DOWN, I think it's OK to consider it as unhealthy.
The combo DOWN+CLOUD+POWER_DOWN+POWERED_DOWN+NOT_RESPONDING will match is_down and considered as unhealthy, in accordance to https://github.com/aws/aws-parallelcluster-node/pull/402

chenwany (approved):
LGTM!

PR #416, opened 2022-06-22
Added support for rebooting compute nodes via Slurm

jdeamicis:
### Description of changes
* Added support for rebooting compute nodes via Slurm without having clustermgtd marking nodes as unhealthy during the reboot.

### Tests
* Manual tests with `scontrol reboot` and `scontrol reboot asap` with both static and dynamic nodes.
* Manual tests with `scontrol update Node=<> State=POWER_DOWN_ASAP`.

### References
* https://github.com/aws/aws-parallelcluster/pull/4133
* https://github.com/aws/aws-parallelcluster-cookbook/pull/1463

### Checklist

### DIFF
--- a/src/slurm_plugin/slurm_resources.py
+++ b/src/slurm_plugin/slurm_resources.py
@@ -121,6 +121,9 @@ class SlurmNode(metaclass=ABCMeta):
     SLURM_SCONTROL_RESUME_FAILED_STATE = {"DOWN", "CLOUD", "POWERED_DOWN", "NOT_RESPONDING"}
     # Due to a bug in Slurm a powered down node can enter IDLE+CLOUD+POWER_DOWN+POWERED_DOWN state
     SLURM_SCONTROL_POWER_STATES = [{"IDLE", "CLOUD", "POWERED_DOWN"}, {"IDLE", "CLOUD", "POWERED_DOWN", "POWER_DOWN"}]
+    SLURM_SCONTROL_REBOOT_REQUESTED_STATE = "REBOOT_REQUESTED"
+    SLURM_SCONTROL_REBOOT_ISSUED_STATE = "REBOOT_ISSUED"
+
 
     EC2_ICE_ERROR_CODES = {
         "InsufficientInstanceCapacity",
@@ -166,11 +169,7 @@ class SlurmNode(metaclass=ABCMeta):
 
         drained(sinfo) is equivalent to IDLE+DRAIN(scontrol) or DOWN+DRAIN(scontrol)
         """
-        return (
-            self._is_drain()
-            and not self.is_power_down()
-            and (self.SLURM_SCONTROL_IDLE_STATE in self.states or self.is_down())
-        )
+        return self._is_drain() and (self.SLURM_SCONTROL_IDLE_STATE in self.states or self.is_down())
 
     def is_power_down(self):
         """Check if slurm node is in power down state."""
@@ -230,6 +229,12 @@ class SlurmNode(metaclass=ABCMeta):
 
     def is_ice(self):
         return self.error_code in self.EC2_ICE_ERROR_CODES
+
+    def is_reboot_requested(self):
+        return self.SLURM_SCONTROL_REBOOT_REQUESTED_STATE in self.states
+
+    def is_reboot_issued(self):
+        return self.SLURM_SCONTROL_REBOOT_ISSUED_STATE in self.states
 
     @abstractmethod
     def is_state_healthy(self, terminate_drain_nodes, terminate_down_nodes, log_warn_if_unhealthy=True):
@@ -320,6 +325,8 @@ class StaticNode(SlurmNode):
         """Check if a slurm node's scheduler state is considered healthy."""
         # Check to see if node is in DRAINED, ignoring any node currently being replaced
         if self.is_drained() and terminate_drain_nodes:
+            if self.is_reboot_issued() or self.is_reboot_requested() or self.is_power_down():
+                return True
             if self.is_being_replaced:
                 logger.debug(
                     "Node state check: node %s in DRAINED but is currently being replaced, ignoring, node state: %s",
@@ -333,6 +340,8 @@ class StaticNode(SlurmNode):
                 return False
         # Check to see if node is in DOWN, ignoring any node currently being replaced
         elif self.is_down() and terminate_down_nodes:
+            if self.is_reboot_issued():
+                return True
             if self.is_being_replaced:
                 logger.debug(
                     "Node state check: node %s in DOWN but is currently being replaced, ignoring. Node state: ",
@@ -399,11 +408,15 @@ class DynamicNode(SlurmNode):
         """Check if a slurm node's scheduler state is considered healthy."""
         # Check to see if node is in DRAINED, ignoring any node currently being replaced
         if self.is_drained() and terminate_drain_nodes:
+            if self.is_reboot_issued() or self.is_reboot_requested() or self.is_power_down():
+                return True
             if log_warn_if_unhealthy:
                 logger.warning("Node state check: node %s in DRAINED, node state: %s", self, self.state_string)
             return False
         # Check to see if node is in DOWN, ignoring any node currently being replaced
         elif self.is_down() and terminate_down_nodes:
+            if self.is_reboot_issued():
+                return True
             if not self.is_nodeaddr_set():
                 # Silently handle failed to launch dynamic node to clean up normal logging
                 logger.debug("Node state check: node %s in DOWN, node state: %s", self, self.state_string)
--- a/src/slurm_plugin/slurm_resources.py
+++ b/src/slurm_plugin/slurm_resources.py
@@ -235,6 +235,27 @@ class SlurmNode(metaclass=ABCMeta):
     def is_reboot_issued(self):
         return self.SLURM_SCONTROL_REBOOT_ISSUED_STATE in self.states
 
+    def is_node_rebooting(self):
+        """Check if the node is in a state consistent with the scontrol reboot request."""
+        cond = False
+        if self.is_drained():
+            if self.is_reboot_issued() or self.is_reboot_requested():
+                logger.debug(
+                    "Node state check: node %s in DRAINED but is currently rebooting, ignoring, node state: %s",
+                    self,
+                    self.state_string,
+                )
+                cond = True
+        elif self.is_down():
+            if self.is_reboot_issued():
+                logger.debug(
+                    "Node state check: node %s in DOWN but is currently rebooting, ignoring, node state: %s",
+                    self,
+                    self.state_string,
+                )
+                cond = True
+        return cond
+
     @abstractmethod
     def is_state_healthy(self, terminate_drain_nodes, terminate_down_nodes, log_warn_if_unhealthy=True):
         """Check if a slurm node's scheduler state is considered healthy."""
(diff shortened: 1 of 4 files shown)


src/slurm_plugin/slurm_resources.py
```diff
@@ -320,6 +324,8 @@ def is_state_healthy(self, terminate_drain_nodes, terminate_down_nodes, log_warn
         """Check if a slurm node's scheduler state is considered healthy."""
         # Check to see if node is in DRAINED, ignoring any node currently being replaced
         if self.is_drained() and terminate_drain_nodes:
+            if self.is_reboot_issued() or self.is_reboot_requested() or self.is_power_down():
```
demartinofra: Can we handle the entire reboot logic (`self.is_reboot_issued() or self.is_reboot_requested()`) as a separate block at the beginning of the is_state_healthy, eventually by moving it to a separate function which also writes a log line when the condition is valid?Otherwise this function is becoming harder and harder to read.

Also, what happens if the node fails to reboot? do we detect this case? is slurm changing the state for the node after a timeout and we eventually mark the node as unhealthy?

jdeamicis: > Can we handle the entire reboot logic (`self.is_reboot_issued() or self.is_reboot_requested()`) as a separate block at the beginning of the is_state_healthy, eventually by moving it to a separate function which also writes a log line when the condition is valid?Otherwise this function is becoming harder and harder to read.

Here are some considerations on my side:
* the way the checks on the reboot flags are nested with the checks on the DRAINED / DOWN statuses reflects Slurm's workflow in setting the reboot flags after a reboot request. Moving those checks out of the `if` statements for DRAINED/DOWN means losing that structure.
* I generally agree with you that the `is_state_healthy` method risks becoming more and more unreadable, but I think it's out of scope: we would something like `handle_reboot_states()`, `handle_drained_state()`, `handle_down_state()`, and this should be done in a homogeneous manner throughout `is_state_healthy()`. And I'm not sure yet this is the best way to deal with this.

> Also, what happens if the node fails to reboot? do we detect this case? is slurm changing the state for the node after a timeout and we eventually mark the node as unhealthy?

Correct. After `ResumeTimeout`, if the node hasn't responded yet it will be marked as DOWN and later `clustermgtd` will treat it as unhealthy.

src/slurm_plugin/slurm_resources.py
```diff
@@ -397,8 +425,13 @@ def __init__(self, name, nodeaddr, nodehostname, state, partitions=None, reason=
...
         """Check if a slurm node's scheduler state is considered healthy."""
-        # Check to see if node is in DRAINED, ignoring any node currently being replaced
-        if self.is_drained() and terminate_drain_nodes:
+        # Check if node is rebooting: if so, the node is healthy
+        if self.is_node_rebooting():
+            return True
+        # Check to see if node is in DRAINED, ignoring any node currently being replaced or in POWER_DOWN
+        if self.is_drained() and not self.is_power_down() and terminate_drain_nodes:
+            if self.is_power_down():
+                return True
```
chenwany: Why are we checking 
```
if self.is_power_down():
  return True
```
here?

src/slurm_plugin/slurm_resources.py
```diff
@@ -318,8 +343,11 @@ def is_healthy(self, terminate_drain_nodes, terminate_down_nodes, log_warn_if_un
 
     def is_state_healthy(self, terminate_drain_nodes, terminate_down_nodes, log_warn_if_unhealthy=True):
         """Check if a slurm node's scheduler state is considered healthy."""
-        # Check to see if node is in DRAINED, ignoring any node currently being replaced
-        if self.is_drained() and terminate_drain_nodes:
+        # Check if node is rebooting: if so, the node is healthy
+        if self.is_node_rebooting():
```
chenwany: Can a node in `is_reboot_requested` unhealthy?

chenwany: If I understand correctly, reboot nodes are always in down or drained state, right?
Shall we add some unit tests cases to check `is_node_rebooting`?

src/slurm_plugin/slurm_resources.py
jdeamicis: Thank you, I missed it in the refactoring (it would have never been evaluated anyway).

src/slurm_plugin/slurm_resources.py
jdeamicis: > Can a node in `is_reboot_requested` unhealthy?

If by "unhealthy" we mean either "drained" or "down" (which are the two conditions we consider unhealthy for compute nodes), then yes, and that's exactly why we need that exclusion.

> If I understand correctly, reboot nodes are always in down or drained state, right?

These are the main cases:
1. `MIXED/ALLOCATED/IDLE+REBOOT_REQUESTED`: this is the result of `scontrol reboot <node>`. If a job is running or the node is selected for a job scheduled for the future, the scheduler won't trigger the reboot yet (it's up to the scheduler to decide when it's a good time to reboot the node).
2. `MIXED/ALLOCATED+DRAIN+REBOOT_REQUESTED`: this is the result of `scontrol reboot asap <node>`. The `asap` flag means that the node is set to `DRAIN`, so that it doesn't accept any more jobs after the one currently running (meaning that the node will be rebooted as soon as the running job is over).
3. `DOWN+REBOOT_ISSUED`: this is the state after the reboot has been issued for case 1.
4. `DOWN+DRAIN+REBOOT_ISSUED`: this is the state after the reboot has been issued for case 2.

The cases we need to consider in `is_state_healthy()` are 2, 3 and 4. Case 1 is not problematic in our current logic.

> Shall we add some unit tests cases to check is_node_rebooting?

We could, but we already have unit tests for `is_state_healthy()`, which always calls `is_node_rebooting()`. But it doesn't cost much, so we can add them.

src/slurm_plugin/slurm_resources.py
jdeamicis: Addressed in the last commit.

src/slurm_plugin/slurm_resources.py
jdeamicis: I added some new unit tests for the `is_rebooting()` method (also, I changed name from `is_node_rebooting()` to `is_rebooting()`.

src/slurm_plugin/slurm_resources.py
```diff
@@ -166,11 +168,7 @@ def is_drained(self):
 
         drained(sinfo) is equivalent to IDLE+DRAIN(scontrol) or DOWN+DRAIN(scontrol)
         """
-        return (
-            self._is_drain()
-            and not self.is_power_down()
-            and (self.SLURM_SCONTROL_IDLE_STATE in self.states or self.is_down())
-        )
+        return self._is_drain() and (self.SLURM_SCONTROL_IDLE_STATE in self.states or self.is_down())
```
chenwany, line 171: is_drained() is not used in the is_rebooting logic now, right? You are using is_drain() in the is_rebootong logic. Shall we recover this function to check also `not self.is_power_down()`?

src/slurm_plugin/slurm_resources.py
```diff
@@ -231,6 +229,38 @@ def is_poweing_up_idle(self):
     def is_ice(self):
         return self.error_code in self.EC2_ICE_ERROR_CODES
 
+    def is_reboot_requested(self):
+        return self.SLURM_SCONTROL_REBOOT_REQUESTED_STATE in self.states
+
+    def is_reboot_issued(self):
+        return self.SLURM_SCONTROL_REBOOT_ISSUED_STATE in self.states
+
+    def is_rebooting(self):
```
chenwany, line 238: The naming of this function is confusing for me.
Reboot nodes with `MIXED/ALLOCATED/IDLE+REBOOT_REQUESTED` will be considered as `is_rebooting` = false, but they are actually rebooting. I understand this doesn't impact the check of `is_state_healthy`. It will return True.

But shall we also include this case in `is_rebooting` function? So in the future even if we change the `is_state_healthy` function, we don't rely on that for whether a node is booting

src/slurm_plugin/slurm_resources.py
```diff
@@ -231,6 +229,38 @@ def is_poweing_up_idle(self):
...
+        cond = False
+        if self._is_drain():
+            if self.is_reboot_issued() or self.is_reboot_requested():
+                logger.debug(
+                    "Node state check: node %s in DRAINED but is currently rebooting, ignoring, node state: %s",
+                    self,
+                    self.state_string,
+                )
+                cond = True
+        elif self.is_down():
```
chenwany: `is_down` here doesn't consider nodes in `DOWN+POWER_DOWN` state. So if nodes is in `REBOOT_ISSUED+DOWN+POWER_DOWN` state, node is not considered as rebooting node. It will be treated as unhealthy node and be terminated.
Does this case happen?

src/slurm_plugin/slurm_resources.py
```diff
@@ -231,6 +229,38 @@ def is_poweing_up_idle(self):
...
+
+    def is_rebooting(self):
+        """
+        Check if the node is rebooting via scontrol reboot.
+
+        Check that the node is in a state consistent with the scontrol reboot request, which could trigger
+        the health checks.
+        """
+        cond = False
+        if self._is_drain():
```
chenwany: We want to consider any nodes with `REBOOT_ISSUED+DRAIN` to be rebooting nodes, right?

src/slurm_plugin/slurm_resources.py
jdeamicis: This case shouldn't happen normally during the `scontrol reboot` workflow, so it should be considered unhealthy.

src/slurm_plugin/slurm_resources.py
jdeamicis: Correct. and in the line below it's what we do (247 in that commit).

src/slurm_plugin/slurm_resources.py
jdeamicis: No, I really want to check just the `DRAIN` flag (so `_is_drain()`) within the logic `is_rebooting()`. The case for `is_down()` is treated in the `elif` below, and has a slightly different logic.

The previous usage of `is_drained()` was coming from the structure of `is_healthy_state()`, but it's better to use `_is_drain()` here.

src/slurm_plugin/slurm_resources.py
jdeamicis: I agree the name is confusing. A clearer (although uglier) name would be `is_rebooting_healthily()`, because in this function we should check if the node is in any rebooting state which might trigger our health checks, and if so return `True` to prevent the node from being replaced.

I will discuss with @demartinofra about the rest of your comments.

src/slurm_plugin/slurm_resources.py
```diff
@@ -231,6 +229,28 @@ def is_poweing_up_idle(self):
...
+        Check that the node is in a state consistent with the scontrol reboot request.
+        """
+        cond = False
+        if self.is_reboot_issued() or self.is_reboot_requested():
+            logger.debug(
+                "Node state check: node %s is currently rebooting, ignoring, node state: %s",
+                self,
+                self.state_string,
+            )
+            cond = True
```
demartinofra, line 251: nit: now you could even drop the cond variable and replace with a return

demartinofra (approved):
LGTM! Well done and thanks for the extra testing! 👍🏻

PR #424, opened 2022-07-04
[Dependencies] Set requirement retrying~=1.3.3

gmarciani:
### Description of changes
Set requirement retrying~=1.3

### Tests
1. `pip install -r requirements.txt` installs retrying 1.3.3, like it was before the change.

Signed-off-by: Giacomo Marciani <mgiacomo@amazon.com>

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
 boto3>=1.7.55
-retrying>=1.3.3
+retrying~=1.3


requirements.txt
```diff
@@ -1,2 +1,2 @@
 boto3>=1.7.55
-retrying>=1.3.3
+retrying~=1.3.3
```
demartinofra: I'd pin only the major version with something like: retrying~=1.3 which in the end translates to retrying>=1.3.3,retrying<2

demartinofra (approved):

PR #429, opened 2022-08-04
Use CreateFleet API when multiple instance types are specified in the compute resource configuration

enrico-usai:
## Description of changes

This patch permits to call boto3 `create-fleet` API when `InstanceTypeList` is specified in the cluster configuration.
The create-fleet response just contain the instance-id list, so we need to execute an additional describe-instances call
to retrieve other information like ip address to be saved in DynamoDB.

To pass the queue/compute-resource configuration parameters to the create-fleet api I'm using the `cluster-config.yaml` file.
This file contains the settings from the user and default values (e.g. `AllocationStrategy`).

The create-fleet api can return a list of Errors,
they will be logged as errors if there are no returned instances, as warning otherwise.

### New class hierarchy: `*FleetManager`
Within this patch we're moving responsibility to launch the EC2 instances from `InstanceManager` to `*FleetManager`.
The fleet manager will be selected (`Ec2InstanceManager` - `run-instances` API vs `Ec2FleetManager` - `create-fleet` API)
according to the configuration of the given compute resource.

### Run-instances-overrides
This patch renames internal variables from `run_instances_overrides` to `launch_overrides` to clarify
the code is able to pass the overridden parameters to both the APIs.
The configuration parameter and the file has not been renamed to preserve back-word compatibility.

### All or nothing
I'm adding a basic support for All-or-nothing feature, setting `MinTargetCapacity` to 1
when all_or_nothing is not configured.
If the minimum target capacity is not reached, the fleet launches no instances,
this might differ from run-instances approach where the run-instances call fails, to be verified.

### Multi-AZ support
In the create-fleet API we're explicitly setting `SingleAvailabilityZone` to `True`, the multi-AZ support is not included.

## Tests
Manually tested by updating the node package in a running cluster and reloading clustermgtd daemon
(slurm_resume changes don't require any reload).
To test this package I had to manually modify the `cluster-config.yaml` by adding `AllocationStrategy` and `InstanceTypeList` params
and adding some required policies to the Role attached to the head node.
Required policies are: `ec2:CreateFleet`, `ec2:RunInstances`, `ec2:DescribeCapacityReservations` and `ec2:CreateTags`.
* Manually tested config file with `SpotPrice: null` and with a value of 10. In the log I see `'Overrides': [{'MaxPrice': '10', 'InstanceType': 't2.medium'}]}],` if value is set, `'Overrides': [{'InstanceType': 't2.medium'}]}],` otherwise
* TODO: integration tests

## References
* https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.create_fleet

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ aws-parallelcluster-node CHANGELOG
 
 This file is used to list changes made in each version of the aws-parallelcluster-node package.
 
+3.x.x
+------
+
+**ENHANCEMENTS**
+- Add support for EC2 Fleet as an alternative instance provisioning mechanism that allows greater flexibility in terms of instance diversification and launch strategy.
+
 3.2.0
 ------
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 boto3>=1.7.55
+PyYAML~=5.3  # TODO remove
 retrying~=1.3
--- a/src/slurm_plugin/clustermgtd.py
+++ b/src/slurm_plugin/clustermgtd.py
@@ -122,11 +122,12 @@ class ClustermgtdConfig:
         "logging_config": os.path.join(
             os.path.dirname(__file__), "logging", "parallelcluster_clustermgtd_logging.conf"
         ),
-        "instance_type_mapping": "/opt/slurm/etc/pcluster/instance_name_type_mappings.json",
-        "run_instances_overrides": "/opt/slurm/etc/pcluster/run_instances_overrides.json",
+        "instance_type_mapping": "/opt/slurm/etc/pcluster/instance_name_type_mappings.json",  # TODO remove this file
         # Launch configs
         "launch_max_batch_size": 500,
         "update_node_address": True,
+        "run_instances_overrides": "/opt/slurm/etc/pcluster/run_instances_overrides.json",
+        "cluster_config_file": "/opt/parallelcluster/shared/cluster-config.yaml",
         # Terminate configs
         "terminate_max_batch_size": 1000,
         # Timeout to wait for node initialization, should be the same as ResumeTimeout
@@ -159,7 +160,7 @@ class ClustermgtdConfig:
             return (
                 self._config == other._config
                 and self.instance_name_type_mapping == other.instance_name_type_mapping
-                and self.run_instances_overrides == other.run_instances_overrides
+                and self.launch_overrides == other.launch_overrides
             )
         return False
 
@@ -177,21 +178,6 @@ class ClustermgtdConfig:
             "clustermgtd", "instance_type_mapping", fallback=self.DEFAULTS.get("instance_type_mapping")
         )
         self.instance_name_type_mapping = read_json(instance_name_type_mapping_file)
-
-        # run_instances_overrides_file contains a json with the following format:
-        # {
-        #     "queue_name": {
-        #         "compute_resource_name": {
-        #             "RunInstancesCallParam": "Value"
-        #         },
-        #         ...
-        #     },
-        #     ...
-        # }
-        run_instances_overrides_file = config.get(
-            "clustermgtd", "run_instances_overrides", fallback=self.DEFAULTS.get("run_instances_overrides")
-        )
-        self.run_instances_overrides = read_json(run_instances_overrides_file, default={})
 
         # Configure boto3 to retry 1 times by default
         self._boto3_retry = config.getint("clustermgtd", "boto3_retry", fallback=self.DEFAULTS.get("max_retry"))
@@ -218,6 +204,25 @@ class ClustermgtdConfig:
         self.update_node_address = config.getboolean(
             "clustermgtd", "update_node_address", fallback=self.DEFAULTS.get("update_node_address")
         )
+        self.cluster_config_file = config.get(
+            "clustermgtd", "cluster_config_file", fallback=self.DEFAULTS.get("cluster_config_file")
+        )
+
+        # run_instances_overrides_file contains a json with the following format:
+        # TODO generalize file name to add create-fleet support too.
+        # {
+        #     "queue_name": {
+        #         "compute_resource_name": {
+        #             "RunInstancesCallParam": "Value"
+        #         },
+        #         ...
+        #     },
+        #     ...
+        # }
+        run_instances_overrides_file = config.get(
+            "clustermgtd", "run_instances_overrides", fallback=self.DEFAULTS.get("run_instances_overrides")
+        )
+        self.launch_overrides = read_json(run_instances_overrides_file, default={})
 
     def _get_health_check_config(self, config):
         self.disable_ec2_health_check = config.getboolean(
@@ -340,7 +345,8 @@ class ClusterManager:
             head_node_private_ip=config.head_node_private_ip,
             head_node_hostname=config.head_node_hostname,
             instance_name_type_mapping=config.instance_name_type_mapping,
-            run_instances_overrides=config.run_instances_overrides,
+            launch_overrides=config.launch_overrides,
+            cluster_config_file=config.cluster_config_file,
         )
 
     def _update_compute_fleet_status(self, status):
--- a/src/slurm_plugin/fleet_manager.py
+++ b/src/slurm_plugin/fleet_manager.py
@@ -16,9 +16,47 @@ from abc import ABC, abstractmethod
 import boto3
 import yaml
 from botocore.exceptions import ClientError
-from slurm_plugin.slurm_resources import EC2Instance
 
 logger = logging.getLogger(__name__)
+
+
+class EC2Instance:
+    def __init__(self, id, private_ip, hostname, launch_time):
+        """Initialize slurm node with attributes."""
+        self.id = id
+        self.private_ip = private_ip
+        self.hostname = hostname
+        self.launch_time = launch_time
+        self.slurm_node = None
+
+    def __eq__(self, other):
+        """Compare 2 SlurmNode objects."""
+        if isinstance(other, EC2Instance):
+            return self.__dict__ == other.__dict__
+        return False
+
+    def __repr__(self):
+        attrs = ", ".join(["{key}={value}".format(key=key, value=repr(value)) for key, value in self.__dict__.items()])
+        return "{class_name}({attrs})".format(class_name=self.__class__.__name__, attrs=attrs)
+
+    def __str__(self):
+        return f"{self.id}"
+
+    def __hash__(self):
+        return hash(self.id)
+
+    @staticmethod
+    def from_describe_instance_data(instance_info):
+        try:
+            return EC2Instance(
+                instance_info["InstanceId"],
+                instance_info["PrivateIpAddress"],
+                instance_info["PrivateDnsName"].split(".")[0],
+                instance_info["LaunchTime"],
+            )
+        except KeyError as e:
+            logger.error("Unable to retrieve EC2 instance info: %s", e)
+            raise e
 
 
 class FleetManagerException(Exception):
--- a/src/slurm_plugin/fleet_manager.py
+++ b/src/slurm_plugin/fleet_manager.py
@@ -1,0 +1,361 @@
+# Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with
+# the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+import copy
+import logging
+import time
+from abc import ABC, abstractmethod
+
+import boto3
+import yaml
+from botocore.exceptions import ClientError
+from slurm_plugin.slurm_resources import EC2Instance
+
+logger = logging.getLogger(__name__)
+
+
+class FleetManagerException(Exception):
+    """Represent an error during the execution of an action with the FleetManager or FleetManagerFactory."""
+
+    def __init__(self, message: str):
+        super().__init__(message)
+
+
+class FleetManagerFactory:
+    @staticmethod
+    def get_manager(cluster_name, region, boto3_config, cluster_config_file, queue, compute_resource, all_or_nothing):
+        try:
+            cluster_config = FleetManagerFactory._load_cluster_config(cluster_config_file)
+
+            queue_config = next(
+                q_config for q_config in cluster_config["Scheduling"]["SlurmQueues"] if q_config["Name"] == queue
+            )
+            compute_resource_config = next(
+                cr_config for cr_config in queue_config["ComputeResources"] if cr_config["Name"] == compute_resource
+            )
+        except FileNotFoundError as e:
+            message = f"Unable to read {cluster_config_file}, {e}"
+            logger.error(message)
+            raise FleetManagerException(message)
+        except StopIteration:
+            message = (
+                f"Unable to found queue {queue} or compute resource {compute_resource} in the {cluster_config_file}"
+            )
+            logger.error(message)
+            raise FleetManagerException(message)
+
+        if compute_resource_config.get("InstanceTypeList"):
+            return Ec2CreateFleetManager(
+                cluster_name,
+                region,
+                boto3_config,
+                queue_config,
+                compute_resource_config,
+                all_or_nothing,
+            )
+        elif compute_resource_config.get("InstanceType"):
+            return Ec2RunInstancesManager(
+                cluster_name,
+                region,
+                boto3_config,
+                queue_config,
+                compute_resource_config,
+                all_or_nothing,
+            )
+        else:
+            raise Exception(
+                "InstanceTypeList or InstanceType field not found "
+                f"in queue: {queue}, compute resource: {compute_resource}"
+            )
+
+    @staticmethod
+    def _load_cluster_config(config_file_path):
+        """
+        Load cluster configuration.
+
+        Cluster config is required to retrieve queue/compute-resource settings to be used in create-fleet call.
+        """
+        with open(config_file_path) as config_file:
+            return yaml.load(config_file, Loader=yaml.SafeLoader)
+
+
+class FleetManager(ABC):
+    """Abstract Fleet Manager."""
+
+    @abstractmethod
+    def __init__(
+        self,
+        cluster_name,
+        region,
+        boto3_config,
+        queue_config,
+        compute_resource_config,
+        all_or_nothing=False,
+    ):
+        self._cluster_name = cluster_name
+        self._region = region
+        self._boto3_config = boto3_config
+        self._queue_config = queue_config
+        self._queue = queue_config["Name"]
+        self._compute_resource_config = compute_resource_config
+        self._compute_resource = compute_resource_config["Name"]
+        self._all_or_nothing = all_or_nothing
+
+    @abstractmethod
+    def _evaluate_launch_params(self, count, launch_overrides):
+        pass
+
+    @abstractmethod
+    def _launch_instances(self, launch_params):
+        pass
+
+    def launch_ec2_instances(self, count, launch_overrides):
+        """
+        Launch EC2 instances.
+
+        :raises ClientError in case of failures with Boto3 calls (run_instances, create_fleet, describe_instances)
+        :raises FleetManagerException in case of missing required instance type info (e.g. private-ip) after 3 retries.
+        """
+        launch_params = self._evaluate_launch_params(count, launch_overrides)
+        assigned_nodes = self._launch_instances(launch_params)
+        logger.info(f"Launched the following instances: {assigned_nodes.get('Instances')}")
+        return [EC2Instance.from_describe_instance_data(instance_info) for instance_info in assigned_nodes["Instances"]]
+
+
+class Ec2RunInstancesManager(FleetManager):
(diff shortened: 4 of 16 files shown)


EddyMM (dismissed):

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ aws-parallelcluster-node CHANGELOG
 
 This file is used to list changes made in each version of the aws-parallelcluster-node package.
 
+3.x.x
+------
+
+**ENHANCEMENTS**
+- Add support for flexible instance types, by calling create-fleet API when multiple instance types are specified in the cluster configuration.
```
demartinofra: nit: I would slightly reword this to capture the changes specific to this package. e.g. `Add support for EC2 Fleet as an alternative instance provisioning mechanism that allows greater flexibility in terms of instance diversifications and launch strategy`?

requirements.txt
```diff
@@ -1,2 +1,3 @@
 boto3>=1.7.55
+PyYAML~=5.3
```
demartinofra: Shall we add a TODO to remove this once we change the config format or you'll remember to do so?

src/slurm_plugin/clustermgtd.py
```diff
@@ -218,6 +205,25 @@ def _get_launch_config(self, config):
         self.update_node_address = config.getboolean(
             "clustermgtd", "update_node_address", fallback=self.DEFAULTS.get("update_node_address")
         )
+        self.cluster_config_file = config.get(
+            "clustermgtd", "cluster_config_file", fallback=self.DEFAULTS.get("cluster_config_file")
+        )
```
demartinofra, line 209: reminder: add this to the __eq__ function when using the final config file.

src/slurm_plugin/fleet_manager.py
```diff
@@ -0,0 +1,300 @@
...
+        Cluster config is required to retrieve queue/compute-resource settings to be used in create-fleet call.
+        """
+        with open(config_file_path) as config_file:
+            return yaml.load(config_file, Loader=yaml.SafeLoader)
+
+
+class FleetManager(ABC):
+    """Abstract Fleet Manager."""
+
+    def __init__(
```
demartinofra, line 131: isn't this also supposed to be abstract so that FleetManager can't be instantiated?

optional: we might start adding typing info for the new methods we add.

src/slurm_plugin/fleet_manager.py
```diff
@@ -0,0 +1,300 @@
...
+    @abstractmethod
+    def _launch_instances(self, launch_params):
+        pass
+
+    def launch_ec2_instances(self, count, launch_overrides):
+        """
+        Launch EC2 instances.
+
+        :raises ClientError in case of failures with Boto3 calls (run_instances, create_fleet, describe_instances)
+        :raises FleetManagerException in case of consecutive missing info
```
demartinofra: what do you mean by consecutive missing info? might be worth clarifying since it's not obvious from the implementation of launch_ec2_instances but requires digging into other calls

src/slurm_plugin/slurm_resources.py
```diff
@@ -63,6 +63,19 @@ def __str__(self):
     def __hash__(self):
         return hash(self.id)
 
+    @staticmethod
+    def from_info(instance_info):
```
demartinofra: nit: from_describe_instance_data/info?

src/slurm_plugin/fleet_manager.py
```diff
@@ -0,0 +1,300 @@
...
+        """
+        launch_params = self._evaluate_launch_params(count, launch_overrides)
+        assigned_nodes = self._launch_instances(launch_params)
+        return [EC2Instance.from_info(instance_info) for instance_info in assigned_nodes["Instances"]]
+
+
+class Ec2InstancesManager(FleetManager):
```
demartinofra: nit: would it make sense to keep the full EC2 api in the name: Ec2RunInstancesManager and Ec2CreateFleetManager?

src/slurm_plugin/fleet_manager.py
```diff
@@ -0,0 +1,300 @@
...
+        if launch_overrides:
+            logger.info("Found CreateFleet parameters override. Launching instances with: %s", launch_params)
+        return launch_params
+
+    def _launch_instances(self, launch_params):
+        """Launch a batch of ec2 instances."""
+        try:
+            response = create_fleet(self._region, self._boto3_config, launch_params)
+            for error in response.get("Errors", []):
+                logger.error("Error in CreateFleet request: %s - %s", error.get("ErrorCode"), error.get("ErrorMessage"))
```
demartinofra: is an error in the fleet response also always an error for our process? or is an error returned also when a subset of the instance types is not available and we don't want to mark it as an error?

src/slurm_plugin/fleet_manager.py
```diff
@@ -0,0 +1,300 @@
...
+
+            logger.debug("CreateFleet response: %s", response)
+            instance_ids = [inst_id for instances in response["Instances"] for inst_id in instances["InstanceIds"]]
+            return {"Instances": self._get_instances_info(instance_ids)}
```
demartinofra: in case we can't get the info for a single instance out of the entire batch it would be still ideal to avoid aborting the entire instance launch. Can we make it so that we return the instances we could describe and log an error for the missing ones that will be eventually terminated at the next iteration?

src/slurm_plugin/fleet_manager.py
```diff
@@ -0,0 +1,300 @@
...
+        :raises EC2InstancesInfoUnavailable if after N retries it's not possible to retrieve info for all the instances
+        :raises ClientError in case of boto3 failure
+        """
+        # Wait for instances to be available in EC2
+        time.sleep(5)
+
+        # Retrieve instance info from Instance Ids
+        instances = []
+        if instance_ids:
+            logger.info(f"Retrieving instance info for instances: {instance_ids}")
```
demartinofra: Shall we move this log to the _launch_instances function rephrasing to something like: launched the following EC2 instances:?

src/slurm_plugin/fleet_manager.py
```diff
@@ -0,0 +1,300 @@
...
+
+            partial_info_instances = []
+            for instance_info in instances:
+                try:
+                    # Try to build EC2Instance objects retrieving all the required fields
+                    EC2Instance.from_info(instance_info)
+                except KeyError:
+                    failed_instance_id = instance_info["InstanceId"]
+                    partial_info_instances.append(failed_instance_id)
+                    instances.remove(instance_info)
```
demartinofra: you don't want to modify the collection while iterating over it, make a copy if you need to do so. Also a set is more efficient than a list in this case

src/slurm_plugin/fleet_manager.py
```diff
@@ -0,0 +1,300 @@
...
+                    logging.warning(f"Unable to retrieve all the required fields for instance {failed_instance_id}")
```
demartinofra: Can we log this only once for all failed instances?

src/slurm_plugin/fleet_manager.py
```diff
@@ -0,0 +1,300 @@
...
+
+            if partial_info_instances:
+                # Try to find information for incomplete instances up to 3 times
+                if retry != 0:
+                    instances.extend(self._get_instances_info(partial_info_instances, retry - 1))
```
demartinofra: rather than calling this function recursively why don't we separate the logic in two functions, one that tries to retrieve the info and another one that manages the retry? it should make the entire code more readable by separating the responsibilities

src/slurm_plugin/fleet_manager.py
```diff
@@ -0,0 +1,300 @@
...
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+import logging
+import time
+from abc import ABC, abstractmethod
+
+import boto3
+import yaml
+from botocore.exceptions import ClientError
+from slurm_plugin.slurm_resources import EC2Instance
```
demartinofra: optional: it would be nice to drop the dependency of fleet manager on something which is defined in slurm_plugin. I'd rather expect the other way around. Would it make sense to move EC2Instance out of the slurm modules?

demartinofra (commented):
Nice work!

src/slurm_plugin/fleet_manager.py
enrico-usai: This is something still to be verified/tested, from boto3 create_fleet documentation it is not clear.
This is a behaviour different from the standard boto3 calls on which a ClientError exception is raised.

src/slurm_plugin/fleet_manager.py
enrico-usai: You're right for the modification of the collection - fixed.
For the set 'm using a list because describe_instance_types requires a list at the end.

src/slurm_plugin/fleet_manager.py
enrico-usai: Done

CHANGELOG.md
enrico-usai: done

src/slurm_plugin/fleet_manager.py
enrico-usai: done. Typing not completed

src/slurm_plugin/fleet_manager.py
enrico-usai: improved description.

src/slurm_plugin/slurm_resources.py
enrico-usai: changed

src/slurm_plugin/fleet_manager.py
enrico-usai: done

src/slurm_plugin/fleet_manager.py
enrico-usai: done

src/slurm_plugin/fleet_manager.py
enrico-usai: Done, moved to fleet_manager.py module.

tests/slurm_plugin/test_clustermgtd.py
```diff
@@ -34,7 +34,7 @@
     StaticNode,
 )
 
-from tests.common import MockedBoto3Request
+from tests.common import MockedBoto3Request, client_error, fleet_manager_factory
```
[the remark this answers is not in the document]
enrico-usai: resolved.

tests/slurm_plugin/test_fleet_manager.py
```diff
@@ -0,0 +1,234 @@
...
+import logging
+import os
+from datetime import datetime, timezone
+
+import pytest
+import slurm_plugin
+from assertpy import assert_that
+from slurm_plugin.fleet_manager import Ec2FleetManager, Ec2InstancesManager, FleetManagerFactory
+
+from tests.common import MockedBoto3Request, fleet_manager_factory
```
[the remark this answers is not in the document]
enrico-usai: resolved

tests/slurm_plugin/slurm_resources/test_instance_manager.py
```diff
@@ -25,14 +26,14 @@
     EC2InstanceHealthState,
 )
 
-from tests.common import MockedBoto3Request
+from tests.common import MockedBoto3Request, client_error, fleet_manager_factory
```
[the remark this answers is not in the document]
enrico-usai: resolved.

tests/slurm_plugin/test_resume.py
```diff
@@ -416,10 +345,26 @@
     mock_update_dns = mocker.patch.object(
         slurm_plugin.instance_manager.InstanceManager, "_update_dns_hostnames", auto_spec=True
     )
-    # Only mock boto3 if testing case of valid clustermgtd heartbeat
+    # patch fleet manager
+    mock_fleet_manager = mocker.patch.object(
```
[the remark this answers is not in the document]
enrico-usai: resolved.

requirements.txt
enrico-usai: done. I'll left this comment open to be addressed when changing the config format.

src/slurm_plugin/fleet_manager.py
enrico-usai: Done, removed the raising of the Exception.

src/slurm_plugin/fleet_manager.py
enrico-usai: Partially done. Let's talk about this.

enrico-usai: Done by creating a private recursive function.

src/slurm_plugin/instance_manager.py
```diff
@@ -86,16 +88,23 @@ def add_instances_for_nodes(
...
+                    self._region,
+                    self._boto3_config,
+                    self._cluster_config_file,
+                    queue,
+                    compute_resource,
+                    all_or_nothing_batch,
+                )
+                launch_overrides = self._launch_overrides.get(queue, {}).get(compute_resource, {})
+
                 for batch_nodes in grouper(slurm_node_list, launch_batch_size):
```
lukeseawalker, line 104: does the `launch_batch_size`  has to be the same for both launch instance APIs? Can it be tuned differently or even removed?

enrico-usai: It cannot be removed because we need to use a batch size when adding the items in route53. 
Yes in future we could have different size for the different apis.

src/slurm_plugin/fleet_manager.py
```diff
@@ -0,0 +1,386 @@
...
+                cluster_name,
+                region,
+                boto3_config,
+                queue_config,
+                compute_resource_config,
+                all_or_nothing,
+            )
+        else:
+            raise Exception(
+                f"InstanceTypes or InstanceType field not found in compute resource config: {compute_resource_config}"
```
lukeseawalker: Do you need to print also queue name to be able to identify the compute resource uniquely?
I may be wrong but you can give two compute resources the same name under different queues

src/slurm_plugin/fleet_manager.py
```diff
@@ -0,0 +1,386 @@
...
+                }
+            ],
+            "TargetCapacitySpecification": {
+                "TotalTargetCapacity": count,
+                "DefaultTargetCapacityType": "on-demand"
+                if self._queue_config["CapacityType"] == "ONDEMAND"
+                else "spot",
+            },
+            "Type": "instant",
+            "SpotOptions": {
```
lukeseawalker: can be this set when launching only on-demand?

src/slurm_plugin/fleet_manager.py
```diff
@@ -0,0 +1,386 @@
...
+                "AllocationStrategy": self._queue_config["AllocationStrategy"],
+                "SingleInstanceType": False,
+                "SingleAvailabilityZone": True,  # Set to False for Multi-AZ support
+                # If the minimum target capacity is not reached, the fleet launches no instances
+                "MinTargetCapacity": 1 if not self._all_or_nothing else count,
+            },
+            "OnDemandOptions": {
```
lukeseawalker: same Q here, can be this set when launching only spot?

src/slurm_plugin/fleet_manager.py
enrico-usai: yes you can fill both of them and then select the one to use with `DefaultTargetCapacityType`

src/slurm_plugin/fleet_manager.py
enrico-usai: same here.

src/slurm_plugin/fleet_manager.py
```diff
@@ -0,0 +1,386 @@
...
+                    EC2Instance.from_describe_instance_data(instance_info)
+                    complete_instances.append(instance_info)
+                except KeyError:
+                    partial_instance_ids.append(instance_info["InstanceId"])
+
+            if partial_instance_ids:
+                # Try to find information for incomplete instances up to 3 times
+                message = f"Unable to retrieve instance info for instances: {partial_instance_ids}"
+                if retry != 0:
+                    logging.warning(message + ". Retrying...")
```
demartinofra: Use logger (same in other places)

src/slurm_plugin/fleet_manager.py
```diff
@@ -0,0 +1,386 @@
...
+            raise e
+
+    def _get_instances_info(self, instance_ids: list):
+        """
+        Describe instances to retrieve info not available from create-fleet response.
+
+        :raises ClientError in case of boto3 failure
+        """
+
+        def _get_instances_info_or_retry(instance_ids_to_retry: list, retry=3):
```
demartinofra: in the end this private function includes everything so at this point you could even get rid of it and assign retry to a local variable.

What I initially had in mind was something like this:
```
    # in _launch_instances
    instance_ids = [inst_id for instance in instances for inst_id in instance["InstanceIds"]]
    complete_instances, partial_instance_ids = self._get_instances_info(instance_ids)
    if partial_instance_ids:
        message = f"Unable to retrieve instance info for instances: {partial_instance_ids}"
        logging.error(message)
    return {"Instances": complete_instances}


    def _get_instances_info(self, instance_ids: list):
        """
        Describe instances to retrieve info not available from create-fleet response.
        :raises ClientError in case of boto3 failure
        """
        complete_instances = []
        partial_instance_ids = list(instance_ids)
        
        
        retry = 3

        while retry > 0 and partial_instance_ids:
            # Wait for instances to be available in EC2
            time.sleep(5)
            result, partial_instance_ids = _retrieve_instances_info_from_ec2(partial_instance_ids)
            complete_instances.append(result)
            retry = retry-1

        return complete_instances, partial_instance_ids
        

    def _retrieve_instances_info_from_ec2(self, instance_ids: list):
        """Retrieve instance info from EC2 by Instance Ids."""
        complete_instances = []
        partial_info_instances = []
        if instance_ids:
            ec2_client = boto3.client("ec2", region_name=self._region, config=self._boto3_config)
            paginator = ec2_client.get_paginator("describe_instances")
            response_iterator = paginator.paginate(InstanceIds=instance_ids)
            filtered_iterator = response_iterator.search("Reservations[].Instances[]")
            instances = [instance_info for instance_info in filtered_iterator]
            logging.debug("Launched instances: %s", instances)

            for instance_info in filtered_iterator
                try:
                    # Try to build EC2Instance objects using all the required fields
                    EC2Instance.from_describe_instance_data(instance_info)
                    complete_instances.append(instance_info)
                except KeyError:
                    partial_info_instances.append(instance_info["InstanceId"])

        return complete_instances, partial_info_instances
```

src/slurm_plugin/fleet_manager.py
enrico-usai: Changed. InstanceTypes vs InstanceTypeList too.

src/slurm_plugin/fleet_manager.py
enrico-usai: done.

src/slurm_plugin/fleet_manager.py
enrico-usai: Added your proposal, except for `extend` vs `append`.

tests/slurm_plugin/test_clustermgtd.py
```diff
@@ -912,12 +917,16 @@ def test_handle_unhealthy_static_nodes(
...
+                "c5xlarge": [
+                    "queue1-st-c5xlarge-1",
+                    "queue1-st-c5xlarge-2",
+                    "queue1-st-c5xlarge-3",
+                ]
             }
         }
     )
-    cluster_manager._instance_manager._launch_ec2_instances = mocker.MagicMock(return_value=launched_instances)
+    # cluster_manager._instance_manager._launch_ec2_instances = mocker.MagicMock(return_value=launched_instances)
```
lukeseawalker: remove if not needed

src/slurm_plugin/fleet_manager.py
```diff
@@ -0,0 +1,383 @@
...
+
+        launch_params = {
+            "LaunchTemplateConfigs": [
+                {
+                    "LaunchTemplateSpecification": {
+                        # LaunchTemplate is different for every compute resources in every queue
+                        "LaunchTemplateName": f"{self._cluster_name}-{self._queue}-{self._compute_resource}",
+                        "Version": "$Latest",
+                    },
+                    "Overrides": overrides,
```
lukeseawalker: missing the `LaunchTemplateConfigs/Overrides/MaxPrice` for the spot case. Correspond to the ParallelCluster configuration option [SpotPrice](https://docs.aws.amazon.com/parallelcluster/latest/ug/Scheduling-v3.html#yaml-Scheduling-SlurmQueues-ComputeResources-SpotPrice)

src/slurm_plugin/fleet_manager.py
```diff
@@ -0,0 +1,383 @@
...
+                }
+            ],
+            "TargetCapacitySpecification": {
+                "TotalTargetCapacity": count,
+                "DefaultTargetCapacityType": "on-demand"
+                if self._queue_config["CapacityType"] == "ONDEMAND"
+                else "spot",
+            },
+            "Type": "instant",
```
lukeseawalker: missing `TagSpecifications`

src/slurm_plugin/fleet_manager.py
enrico-usai: Added. 

Manually tested config file with `SpotPrice: null` and with a value of 10. In the log I see `'Overrides': [{'MaxPrice': '10', 'InstanceType': 't2.medium'}]}],` if value is set, `'Overrides': [{'InstanceType': 't2.medium'}]}],` otherwise

tests/slurm_plugin/test_clustermgtd.py
enrico-usai: done

demartinofra (approved):
🚀🚀🚀🚀🚀

EddyMM (approved):

PR #442, opened 2022-09-15
[FIT] Differentiate run-instances and create-fleet overrides json files

enrico-usai:
### Description of changes
The two files, expected by both clustermgtd and slurm_resume daemons are:
* `/opt/slurm/etc/pcluster/run_instances_overrides.json`
* `/opt/slurm/etc/pcluster/create_fleet_overrides.json`

They will be passed to the fleet manager instance according to the API
to use for the given compute resource.

I modified the `read_json` file to avoid printing an INFO message with "exception" word
into it, when the file doesn't exist, to avoid to confuse users with this Errno/exception word.
Before of this patch we had:
```
2022-09-14 09:42:24,675 - [common.utils:read_json] - INFO - Unable to read file
'/opt/slurm/etc/pcluster/run_instances_overrides.json'
due to an exception: [Errno 2] No such file or directory:
'/opt/slurm/etc/pcluster/run_instances_overrides.json'. Using default: {}
````
Now we'll have this INFO message only when we're not able to parse the file (e.g. `JSONDecodeError`).`

Within this patch I also removed the useless cluster_config.yaml file from slurm_resume and clustermgtd config files.
This was a leftover of a previous commit on which we were using the yaml file to identify the API to use for each compute resource. We're using fleet_config_file now.

### Unit tests changes

I extended and fixed `read_json` unit tests that weren't using the files in the right folder
and they were always failing because of `FileNotFoundError`.

I added an empty override dict {} to all the read_json side effect,
because right now we are parsing 3 files rather than 2.
I also removed useless overrides content when not needed, replacing the values with {}

### Checklist

### DIFF
--- a/src/common/utils.py
+++ b/src/common/utils.py
@@ -263,5 +263,6 @@ def read_json(file_path, default=None):
             log.error("Unable to read file from '%s'. Failed with exception: %s", file_path, e)
             raise
         else:
-            log.info("Unable to read file '%s' due to an exception: %s. Using default: %s", file_path, e, default)
+            if not isinstance(e, FileNotFoundError):
+                log.info("Unable to read file '%s' due to an exception: %s. Using default: %s", file_path, e, default)
             return default
--- a/tests/slurm_plugin/test_clustermgtd.py
+++ b/tests/slurm_plugin/test_clustermgtd.py
@@ -914,7 +914,6 @@ def test_handle_unhealthy_static_nodes(
         protected_failure_count=10,
         insufficient_capacity_timeout=600,
         launch_overrides={"dynamic": {"c5.xlarge": {"InstanceType": "t2.micro"}}},
-        cluster_config_file="/path/of/config-file.yaml",
     )
     for node, instance in zip(unhealthy_static_nodes, instances):
         node.instance = instance
--- a/tests/slurm_plugin/test_clustermgtd.py
+++ b/tests/slurm_plugin/test_clustermgtd.py
@@ -150,7 +150,9 @@ class TestClustermgtdConfig:
         ids=["default", "all_options", "health_check"],
     )
     def test_config_parsing(self, config_file, expected_attributes, test_datadir, mocker):
-        mocker.patch("slurm_plugin.clustermgtd.read_json", side_effect=[FLEET_CONFIG, LAUNCH_OVERRIDES])
+        mocker.patch(
+            "slurm_plugin.clustermgtd.read_json", side_effect=[FLEET_CONFIG, LAUNCH_OVERRIDES, LAUNCH_OVERRIDES]
+        )
         sync_config = ClustermgtdConfig(test_datadir / config_file)
         for key in expected_attributes:
             assert_that(sync_config.__dict__.get(key)).is_equal_to(expected_attributes.get(key))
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

File: src/common/utils.py [hunk 1]
```diff
@@ -263,5 +263,6 @@ def read_json(file_path, default=None):
             log.error("Unable to read file from '%s'. Failed with exception: %s", file_path, e)
             raise
         else:
-            log.info("Unable to read file '%s' due to an exception: %s. Using default: %s", file_path, e, default)
+            if not isinstance(e, FileNotFoundError):
```
EddyMM, line 266: This will limit the logs available to detect scenarios where the file can't be found but a default value is provided.
Is there a reason for avoiding logging of `FileNotFoundError` errors?

tests/slurm_plugin/test_clustermgtd.py
```diff
@@ -1840,7 +1846,7 @@ def test_manage_cluster_boto3(
     # patch boto3 call
     boto3_stubber("ec2", mocked_boto3_request)
     mocker.patch("slurm_plugin.clustermgtd.datetime").now.return_value = datetime(2020, 1, 2, 0, 0, 0)
-    mocker.patch("slurm_plugin.clustermgtd.read_json", side_effect=[FLEET_CONFIG, LAUNCH_OVERRIDES])
+    mocker.patch("slurm_plugin.clustermgtd.read_json", side_effect=[FLEET_CONFIG, LAUNCH_OVERRIDES, LAUNCH_OVERRIDES])
```
EddyMM, line 1849: Since we'll be having two `overrides` files, can we make the tests explicitly check that the correct overrides type is being applied. At the moment `LAUNCH_OVERRIDES` is an empty JSON object`{}` so we can't be 100% sure if the correct content is being read.

It would be great to have a `RUN_INSTANCE_LAUNCH_OVERRIDES` and `CREATE_FLEET_LAUNCH_OVERRIDES` with corresponding test content for proper comparison.

src/common/utils.py
enrico-usai: I did it by purpose, see PR description.
Now we have two files and it will become a common case to not have the run-instances-override.json file or anyway miss one of the two files and this `due to an exception: [Errno 2] No such file or directory` is a confusing message. 

I don't like printing the word "exception" and "Errno" if the file is missing, but there are no real errors, the file simply doesn't exist and this is the default and the most common case. 
Let's discuss about this change.

tests/slurm_plugin/test_clustermgtd.py
enrico-usai: I left an empty json object for the other test cases on which we're not really using these parameters.

I have this kind of tests, with different content for the two files, in the `TestCreateFleetManager.test_evaluate_launch_params` and `TestRunInstancesManager.test_evaluate_launch_params`, in the `test_fleet_manager.py` module. 

Could you double check if this is what you're searching for?

src/common/utils.py
enrico-usai: Additional details:
* for files without a default we're still printing the error and failing.
* for files with a default (for which we were printing just an info message) we're now differentiating two cases:
  * if the file doesn't exist --> this is not a real error and this will be the default for most of the users --> no info message
  * if the file exists and there are parsing error --> show the info message but continue to use the default as before.

EddyMM: Thanks for the explanation. Makes sense.

tests/slurm_plugin/test_clustermgtd.py
EddyMM: Thanks for referencing those tests.
- [TestEc2RunInstancesManager](https://github.com/aws/aws-parallelcluster-node/blob/cd6781f2f5fc167052a61151d4c9611e67a5ef33/tests/slurm_plugin/test_fleet_manager.py#L130-L130)
- [TestCreateFleetManager](https://github.com/aws/aws-parallelcluster-node/blob/cd6781f2f5fc167052a61151d4c9611e67a5ef33/tests/slurm_plugin/test_fleet_manager.py#L326-L326)

EddyMM (approved):

PR #456, opened 2022-11-02
Support compute node launches on multiple subnets across availability zones

EddyMM:
### Description of changes
* Each queue is currently associated with one subnet (therefore one Availability Zone)
* To increase the capacity pool from which instances can be launched, these changes add support to read a list of Subnet IDs from the `fleet-config` file
* The `fleet-config` file has this format
```
{
    "queue": {
        "compute-resource": {
            "Api": "create-fleet",
            "Instances": [
                {
                    "InstanceType": "..."
                },
                ...
            ]
            "Networking": {
                "SubnetIds": [...]
            }
        }
    }
}
```
* A combination of Subnet IDs and InstanceTypes is then used in the [Overrides](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_FleetLaunchTemplateOverridesRequest.html) parameter of the [CreateFleet](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateFleet.html) Request
* The minimum target capacity is only available with single instance fleets or single AZ fleets

### Tests
* Unit test that includes both a single and multiple subnets in a queue
    * Refactored tests to use explicit launch parameters instead of generating them in code (this is to simplify verification fo input & output)

### References
* [CreateFleet](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateFleet.html)
* [FleetLaunchTemplateOverridesRequest](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_FleetLaunchTemplateOverridesRequest.html)

### Checklist


src/slurm_plugin/fleet_manager.py
```diff
@@ -243,24 +243,50 @@ def __init__(
...
+        """Build and return the list of Launch Template Overrides to be applied in the CreateFleet request.
+
+        (https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_FleetLaunchTemplateOverridesRequest.html)
+        """
+        template_overrides = []
+        overrides = {}
+
+        if self._compute_resource_config["CapacityType"] == "spot":
+            if self._compute_resource_config.get("MaxPrice"):
+                overrides = {"MaxPrice": str(self._compute_resource_config["MaxPrice"])}
```
lukeseawalker: minor - if this code block is moved, than the assignment will break the logic. What about using ".update" also here?

lukeseawalker (dismissed):

lukeseawalker (approved):

PR #461, opened 2022-11-19
3.x Capture the console output of static compute nodes that fail to bootstrap

davprat:
### Description of changes
* This changes will cause the head node to retrieve the console output of a static compute node that fails to bootstrap.
* The console output will be logged to CloudWatch from the head node.
* This change breaks compatibility with Python 3.7 and 3.8 due to using features not available before 3.9.

### Tests
* Fixed unit tests broken by this change.
* Added unit tests to cover new functionality.
* Manually tested by deploying static compute nodes into a subnet with no NAT or public IP's.
  * Replaced node package after head node booted up
  * Waited for compute nodes to timeout and terminate
  * Received console output events in the proper CloudWatch stream.

### Checklist


.flake8
```diff
@@ -9,6 +9,7 @@ ignore =
     D107, # Missing docstring in __init__
     W503, # line break before binary operator => Conflicts with black style.
     N818, # exception name should be named with an Error suffix
+    E203, # No space around ':' in a slice => conflicts with black style.
```
demartinofra: can we suppress locally with `# noqa: E203` otherwise we also suppress other checks that are compatible with our style?

src/slurm_plugin/clustermgtd.py
```diff
@@ -350,6 +396,32 @@ def _initialize_instance_manager(config):
...
+            semaphore.release()
+
+        semaphore = self._executor_limit
+        if task:
+            if semaphore.acquire(blocking=False):
+                future = self._executor_pool.submit(task)
+                future.add_done_callback(queue_executor_task_callback)
+                return future
+            else:
+                log.warning("Unable to queue task due to exceeding backlog limit")
```
demartinofra: nit: print current limit

src/slurm_plugin/clustermgtd.py
```diff
@@ -350,6 +396,32 @@ def _initialize_instance_manager(config):
...
+        return None
```
demartinofra: Shall we extract the task executor logic in its own class? especially since clustermgtd is getting pretty overloaded of responsibilities

src/slurm_plugin/clustermgtd.py
```diff
@@ -679,13 +751,42 @@ def _handle_powering_down_nodes(self, slurm_nodes):
...
+                report_nodes = compute_nodes
+
+            self._queue_executor_task(
+                self._instance_manager.get_console_output_task(
+                    report_nodes,
+                    ClusterManager._console_output_callback,
+                    (0, self._config.compute_console_wait_time),
+                )
+            )
+
```
demartinofra: optional: consider extracting this logic too in a ConsoleLogger class. at that point also the get_console_output_task could go in there and not in instance_manager

src/slurm_plugin/clustermgtd.py
```diff
@@ -679,13 +751,42 @@ def _handle_powering_down_nodes(self, slurm_nodes):
...
+    def _report_console_output_from_nodes(self, compute_nodes):
+        if self._config.compute_console_logging_enabled:
+            if self._config.compute_console_logging_max_sample_size > 0:
+                report_nodes = compute_nodes[: self._config.compute_console_logging_max_sample_size]
+            else:
+                report_nodes = compute_nodes
+
+            self._queue_executor_task(
+                self._instance_manager.get_console_output_task(
```
demartinofra: do we get any error logged in case a task fails to be executed?

src/slurm_plugin/instance_manager.py
```diff
@@ -358,3 +361,90 @@ def terminate_all_compute_nodes(self, terminate_batch_size):
...
+                    output_callback(output)
+                    output_hashes.update({instance_id: current_hash})
+
+        def console_collector():
+            try:
+                ec2 = boto3.session.Session().client("ec2", region_name=region)
+
+                output_hashes = {}
+
+                for next_wait in wait_times:
```
demartinofra: why do we do this twice and do the hash comparison? can't we just fetch the output once after the sleep time?

src/slurm_plugin/instance_manager.py
```diff
@@ -358,3 +361,90 @@ def terminate_all_compute_nodes(self, terminate_batch_size):
     def _update_failed_nodes(self, nodeset, error_code="Exception"):
         """Update failed nodes dict with error code as key and nodeset value."""
         self.failed_nodes[error_code] = self.failed_nodes.get(error_code, set()).union(nodeset)
+
+    def get_console_output_task(self, nodes, output_callback, wait_times):
```
demartinofra: it would be nice to move all this code outside of instance_manager

demartinofra (commented):
The approach LGTM overall. I'd however revise the class structure and move the new code outside of clusermgtd and InstanceManager, by creating additional classes such as TaskExecutor and ConsoleLogger so that we better decouple responsibilities and we can more easily understand what is being done without having to jump from one class to the other. What do you think?

src/slurm_plugin/instance_manager.py
davprat: I was thinking we might want to grab the log at the time we detect the node is down and then collect it again after the wait time in an attempt to get extra info in case the final output went over the 64K mark.

Removed.

.flake8
davprat: Moved.

src/slurm_plugin/clustermgtd.py
davprat: Done

src/slurm_plugin/clustermgtd.py
davprat: Done

src/slurm_plugin/clustermgtd.py
davprat: Done. 
Note: The node to instance mapping is still done by the instance_manager since the instance_manager is responsible for the ddb schema.

src/slurm_plugin/clustermgtd.py
davprat: Changed the queue function to raise an exception if the backlog is exceeded.
Also, the task themselves should log errors if they encounter an exception.
And finally, I change the task exception handlers to re-raise any encountered exception so that the exception my be retrieved from the returned future object.

src/slurm_plugin/instance_manager.py
davprat: The code that retrieves the console output has been moved. However, I think it makes sense to leave the code that does the node -> instance id mapping in instance manager since instance manager is responsible for that relationship. Wdyt?

src/slurm_plugin/task_executor.py
```diff
@@ -0,0 +1,51 @@
...
+
+class TaskExecutor:
+    class MaximumBacklogExceededException(RuntimeError):
+        """Exception raised when a task can't be queued due to backlog."""
+
+        def __init__(self, task, maximum_backlog):
+            self.failed_task = task
+            self.maximum_backlog = maximum_backlog
+
+    """Class for managing execution of asynchronous tasks."""
```
demartinofra: does this need to go right under the class declaration?

src/slurm_plugin/task_executor.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def __init__(self, worker_pool_size, max_backlog):
+        self._max_backlog = max_backlog
+        self._executor_limit = Semaphore(max_backlog)
+        self._executor_pool = ThreadPoolExecutor(max_workers=worker_pool_size)
```
demartinofra, line 27: According to the [docs](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor):
> All threads enqueued to ThreadPoolExecutor will be joined before the interpreter can exit.

What happens if we request a stop or a restart of the clustermgtd, will it hand until all threads are done? If so is there anything we can do to avoid the hanging?

Another question I have is: do we need to do anything in the destuctor (`__del__`) of TaskExecutor? What happens when the object is cleaned-up and the ThreadPoolExecutor destroyed?

src/slurm_plugin/task_executor.py
```diff
@@ -0,0 +1,51 @@
+import inspect
+import logging
+from concurrent.futures import ThreadPoolExecutor
+from threading import Semaphore
+
+logger = logging.getLogger(__name__)
+
+
+class TaskExecutor:
```
demartinofra, line 13: [optional] can you add typing information to the new functions?

src/slurm_plugin/clustermgtd.py
```diff
@@ -323,14 +359,24 @@ def __init__(self, config):
...
         self._instance_manager = None
+        self._task_executor = None
+        self._console_logger = None
         self.set_config(config)
 
     def set_config(self, config):
         if self._config != config:
             log.info("Applying new clustermgtd config: %s", config)
+
+            self._task_executor = self._initialize_executor(config)
```
demartinofra, line 385: here I believe that when the config is changed and the executor reinitialized we call shutdown with wait set to False which lets running threads complete. Do we want to add a comment on this behavior? also any side effect this could cause?

src/slurm_plugin/clustermgtd.py
```diff
@@ -686,6 +758,15 @@ def _handle_unhealthy_static_nodes(self, unhealthy_static_nodes):
 
         Set node to down, terminate backing instance, and launch new instance for static node.
         """
+        try:
+            self._console_logger.report_console_output_from_nodes(
+                compute_nodes=unhealthy_static_nodes,
+                instance_supplier=self._instance_manager.get_node_instance_mapper(),
+                task_executor=lambda task: self._task_executor.queue_executor_task(task),
+            )
```
demartinofra: nit: is there a reason we are passing the task executor to the `report_console_output_from_nodes` rather than submitting `report_console_output_from_nodes` to the queue from here?

src/slurm_plugin/console_logger.py
```diff
@@ -0,0 +1,82 @@
...
+                output = response.get("Output")
+                yield {
+                    "Name": instance_name,
+                    "InstanceId": instance_id,
+                    "ConsoleOutput": pattern.sub("\r", output) if output else None,
+                }
+
+        def console_collector():
+            try:
+                sleep_remaining_loop_time(wait_time, start_time)
```
demartinofra: nit: please leave a comment on why we sleep here

src/slurm_plugin/instance_manager.py
```diff
@@ -33,6 +33,28 @@
...
+        str(table_name): {
+            "Keys": [
+                {
+                    "Id": str(node_name),
+                }
+                for node_name in node_names
+            ],
+            "ProjectionExpression": "Id, InstanceId",
+        }
+    }
```
demartinofra: nit: any reason why we want to keep this at the module level?

src/slurm_plugin/instance_manager.py
```diff
@@ -358,3 +383,43 @@ def terminate_all_compute_nodes(self, terminate_batch_size):
...
+            # Return instance ids that don't require a DDB lookup first.
+            for node in nodes_with_instance_id:
+                yield node
+
+            # Only get a ddb resource if we have something to do.
+            if len(nodes_without_instance_id) < 1:
+                return
+
+            # Lookup instance ids in DDB for nodes that we don't already have the instance id for.
+            ddb_resource = resource_factory("dynamodb")
```
demartinofra: resource_factory is in a way a required dependency of `get_instances_for_nodes` but it does not appear in its signature - which means the function expects `resource_factory` to be available in the scope in order to work. I'm wondering if it makes sense to make this requirement explicit.

Same happens for table_name.

One possibility could be to use [partial](https://docs.python.org/3/library/functools.html#functools.partial)

src/slurm_plugin/console_logger.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        if not self._console_logging_enabled:
+            return None
+
+        start_time = datetime.now(tz=timezone.utc)
+
+        instances = tuple(instance for instance in instance_supplier(compute_nodes) if instance.get("InstanceId"))
+
+        if self._max_sample_size > 0:
+            instances = instances[: self._max_sample_size]
```
demartinofra: nit: being very picky here but in case we want to spare the extra memory (and eventually some extra calls) we could just read max_sample_size instances from the supplier rather than reading all of them and then grabbing only the first max_sample_size.

tests/slurm_plugin/test_clustermgtd/TestClustermgtdConfig/test_config_parsing/all_options.conf
```diff
@@ -25,4 +25,6 @@ hosted_zone = hosted-zone
 dns_domain = dns.domain
 use_private_hostname = false
 protected_failure_count = 5
-insufficient_capacity_timeout = 50.5
\ No newline at end of file
+insufficient_capacity_timeout = 50.5
+compute_console_logging_enabled = False
+compute_console_logging_max_sample_size = 50
```
demartinofra, line 30: do we need to add all supported options since this test apparently is supposed to test all of them?

src/slurm_plugin/instance_manager.py
davprat: Not particularly - it was just what IntelliJ recommended - I can put them back in the class and mark them static.

src/slurm_plugin/clustermgtd.py
davprat: Not really - I think this was reduced from an earlier version, but I failed to finally reduce it to its minimal implementation.

src/slurm_plugin/clustermgtd.py
davprat: What kind of side effects are you thinking might happen? I suppose the biggest one is that any Futures that are in progress will still complete and so any callbacks on those Futures will be executed after shutdown. Anything else you can recommend looking into?

src/slurm_plugin/task_executor.py
davprat: Yes, it will hang until all busy worker threads complete - for Python versions >= 3.9, we pass a flag to cancel any pending tasks, but versions < 3.9 do not support that option and so the pool will still complete all pending tasks in the queue before letting the process exit.

src/slurm_plugin/clustermgtd.py
davprat: Fixed.

src/slurm_plugin/console_logger.py
davprat: Refactored to do this.

src/slurm_plugin/console_logger.py
davprat: Added comment.

src/slurm_plugin/instance_manager.py
davprat: Added back to the InstanceManager as static methods.

src/slurm_plugin/instance_manager.py
davprat: Interesting - are partials preferable to using closures in Python?

This code was refactored so that the factory and table name are passed in as parameters to the enclosing method. lmkwyt...

src/slurm_plugin/task_executor.py
davprat: Yes 🤦

tests/slurm_plugin/test_clustermgtd/TestClustermgtdConfig/test_config_parsing/all_options.conf
davprat: Yes 🤦

src/slurm_plugin/task_executor.py
davprat: I took a shot at it. lmkwyt...

src/slurm_plugin/task_executor.py
demartinofra: having the clusetermgtd hang for 3 minutes in the worst case scenario is not a good idea. I'm wondering if we should and could just forcefully kill all threads, or eventually give them a very very small time to complete (e.g. 2secs).

src/slurm_plugin/clustermgtd.py
demartinofra: nope nothing specific on top of what you mentioned. When I first read this line of code without looking at what `_initialize_executor` does I did not think about having threads form the old executor still running. I was just wondering if this implicit behavior could lead to any side affect now (apparently not) or bugs in the future (which is why a comment might slightly help)

src/slurm_plugin/task_executor.py
davprat: I've refactored a bit - so now the tasks will wait on what is ultimately an event. When a shutdown is requested, the event will be set and any "sleeping" tasks will raise an exception. If someone cares about the success or failure of the task, the returned Future can be checked for task status - either cancelled or that an exception was raised. What do you think?

src/slurm_plugin/clustermgtd.py
davprat: As I commented above, by rewriting the method of sleeping, any sleeping tasks can be woken up and terminated when a shutdown is requested.

demartinofra (dismissed):
🚀 nicely done! thanks!

demartinofra (approved):

PR #477, opened 2022-12-16
3.x Remove checks against Python 3.7 and 3.8

davprat:
### Description of changes
* This change removes testing python components against 3.7 and 3.8
*  It should no longer be necessary to test against these versions and the node components are installed and run in a Python 3.9 virtual environment.
* Testing on 3.7 and 3.8 keeps us from using features that are only supported in 3.9 and above.
* Added testing for v3.10.
* Fixed unit tests that failed when run against 3.10.

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,18 +26,10 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         name:
-          - Python 3.7 Tests
-          - Python 3.8 Tests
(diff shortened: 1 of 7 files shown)


demartinofra (commented):
we also need to update setup.py https://github.com/aws/aws-parallelcluster-node/blob/develop/setup.py#L45, and optionally enable tests for Python 3.10.

demartinofra (approved):
Thanks for the extra fixes!

PR #481, opened 2023-01-04
Sort nodename, nodeaddrs, nodehostnames when doing scontrol update

hanwen-pcluste:
### Description of changes
* `update_nodes` function is called from multiple places. What we care about is the callings with nodeaddrs or nodehostnames. Moreover, we only apply the sort when there is nodeaddrs or nodehostnames
* There are two calling places with nodeaddrs or nodehostnames: `reset_nodes` and `_update_slurm_node_addrs`
* In `reset_nodes`, the nodename, nodeaddrs, and nodehostnames are identical. We use logic sort  the nodename, and assign the value to nodeaddrs and nodehostnames
* In `_update_slurm_node_addrs`, we know that all the nodes and addresses are already expanded (i.e. there is not Slurm compressed list expression). We use `zip` to sort the nodeaddrs and nodehostnames according to nodename
* The sort is put after the nodes are batched because `scontrol` accepts 5000 nodes at most. Putting the sort after the nodes are batches will make sure not hitting the limit

### Tests
* Unit tests are changed accordingly.
* Manual testing
We tested 30000 nodenames, addresses, hostnames mismatch can be recovered by clustermgtd.
* Whole integration tests (without AD integration and multiple FSx integration) look ok.

### Checklist

### DIFF
--- a/src/common/schedulers/slurm_commands.py
+++ b/src/common/schedulers/slurm_commands.py
@@ -104,15 +104,41 @@ def update_nodes(
         update_cmd += f" state={state}"
     if reason:
         update_cmd += f' reason="{reason}"'
-    for nodenames, addrs, hostnames in batched_node_info:
-        node_info = f"nodename={nodenames}"
+    for nodenames_, addrs_, hostnames_ in batched_node_info:
+        if nodeaddrs or nodehostnames:
+            # Sorting is only necessary if we set nodeaddrs or nodehostnames
+            nodenames, addrs, hostnames = _sort_nodes_attributes(nodenames_, addrs_, hostnames_)
+        node_info = f"nodename={','.join(nodenames)}"
         if addrs:
-            node_info += f" nodeaddr={addrs}"
+            node_info += f" nodeaddr={','.join(addrs)}"
         if hostnames:
-            node_info += f" nodehostname={hostnames}"
+            node_info += f" nodehostname={','.join(hostnames)}"
         run_command(  # nosec
             f"{update_cmd} {node_info}", raise_on_error=raise_on_error, timeout=command_timeout, shell=True
         )
+
+
+def _sort_nodes_attributes(nodes, nodeaddrs, nodehostnames):
+    nodes_str = ",".join(nodes) if type(nodes) is tuple else nodes
+    sorted_node_names_str = check_command_output(
+        f"{SCONTROL} show hostlistsorted {nodes_str} | xargs {SCONTROL} show hostnames", shell=True
+    )
+    sorted_node_names = sorted_node_names_str.strip().split("\n")
+    if nodes == nodeaddrs and nodeaddrs == nodehostnames:
+        # Path from reset_nodes
+        nodes = nodeaddrs = nodehostnames = sorted_node_names
+    else:
+        # Path from _update_slurm_node_addrs
+        order = {k: i for i, k in enumerate(sorted_node_names)}
+        sorting_key = lambda x: order[x[0]]
+        if nodeaddrs and nodehostnames:
+            _, nodeaddrs, nodehostnames = zip(*sorted(zip(nodes, nodeaddrs, nodehostnames), key=sorting_key))
+        elif nodeaddrs:
+            _, nodeaddrs = zip(*sorted(zip(nodes, nodeaddrs), key=sorting_key))
+        elif nodehostnames:
+            _, nodehostnames = zip(*sorted(zip(nodes, nodehostnames), key=sorting_key))
+        nodes = sorted_node_names
+    return nodes, nodeaddrs, nodehostnames
 
 
 def update_partitions(partitions, state):
@@ -156,7 +182,7 @@ def _batch_attribute(attribute, batch_size, expected_length=None):
     if expected_length and len(attribute) != expected_length:
         raise ValueError
 
-    return [",".join(batch) for batch in grouper(attribute, batch_size)]
+    return [batch for batch in grouper(attribute, batch_size)]
 
 
 def _batch_node_info(nodenames, nodeaddrs, nodehostnames, batch_size):
--- a/src/common/schedulers/slurm_commands.py
+++ b/src/common/schedulers/slurm_commands.py
@@ -122,7 +122,7 @@ def update_nodes(
 
 
 def _sort_nodes_attributes(nodes, nodeaddrs, nodehostnames):
-    nodes_str = ",".join(nodes) if type(nodes) is tuple else nodes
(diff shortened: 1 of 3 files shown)


chenwany (commented):
Could you add more test cases to `test_batch_node_info` with nodenames not sorted?

src/common/schedulers/slurm_commands.py
```diff
@@ -165,6 +165,14 @@ def _batch_node_info(nodenames, nodeaddrs, nodehostnames, batch_size):
         # Only split on , if there is ] before
         # For ex. "node-[1,3,4-5],node-[20,30]" should split into ["node-[1,3,4-5]","node-[20,30]"]
         nodenames = re.split("(?<=]),", nodenames)
+    if nodeaddrs and nodehostnames:
```
chenwany: The input of nodeaddrs here can be a string

src/common/schedulers/slurm_commands.py
```diff
@@ -97,6 +97,25 @@ def update_nodes(
     if we run scontrol update state=fail_state nodeaddr=good_addr nodename=name,
     the scontrol command will fail but nodeaddr will be updated to good_addr.
     """
+    # Only split on , if there is ] before
+    # For ex. "node-[1,3,4-5],node-[20,30]" should split into ["node-[1,3,4-5]","node-[20,30]"]
+    nodes = re.split("(?<=]),", nodes) if type(nodes) is str else nodes
+    nodeaddrs = re.split("(?<=]),", nodeaddrs) if type(nodeaddrs) is str else nodeaddrs
+    nodehostnames = re.split("(?<=]),", nodehostnames) if type(nodehostnames) is str else nodehostnames
```
hanwen-pcluste: We need to double check if these are needed

src/common/schedulers/slurm_commands.py
```diff
@@ -202,8 +221,15 @@ def set_nodes_power_down(nodes, reason=None):
 
 def reset_nodes(nodes, state=None, reason=None, raise_on_error=False):
     """Reset nodeaddr and nodehostname to be equal to nodename."""
+    nodes_str = ",".join(nodes) if type(nodes) is list else nodes
```
hanwen-pcluste: We need to double check if these are needed

src/common/schedulers/slurm_commands.py
```diff
@@ -97,6 +97,25 @@ def update_nodes(
...
+    # Only split on , if there is ] before
+    # For ex. "node-[1,3,4-5],node-[20,30]" should split into ["node-[1,3,4-5]","node-[20,30]"]
+    nodes = re.split("(?<=]),", nodes) if type(nodes) is str else nodes
+    nodeaddrs = re.split("(?<=]),", nodeaddrs) if type(nodeaddrs) is str else nodeaddrs
+    nodehostnames = re.split("(?<=]),", nodehostnames) if type(nodehostnames) is str else nodehostnames
+    if nodeaddrs or nodehostnames:
+        # Sorting is only necessary if we set nodeaddrs or nodehostnames
+        nodes_str = ",".join(nodes)
+        sorted_node_names = (
+            check_command_output(f"scontrol show hostlistsorted {nodes_str} | xargs scontrol show hostname", shell=True)
```
charlesg3: Let's just put the command on a separate line which will make this easier to read

demartinofra: what if this command fails? is the error handled already and the failure reason clear enough?
is there a limit to the number of nodes that these commands can handle?

src/common/schedulers/slurm_commands.py
```diff
@@ -97,6 +97,25 @@ def update_nodes(
     if we run scontrol update state=fail_state nodeaddr=good_addr nodename=name,
     the scontrol command will fail but nodeaddr will be updated to good_addr.
     """
+    # Only split on , if there is ] before
```
demartinofra: please move the new code to a separate function

src/common/schedulers/slurm_commands.py
demartinofra: probably you don't need a regex split for nodeaddrs and nodehostnames but simply need to split on ","

src/common/schedulers/slurm_commands.py
jdeamicis: (minor) the official command is `scontrol show hostnames`, but `hostname` works too.

hanwen-pcluste: Addressed [charlesg3](https://github.com/charlesg3) and [jdeamicis](https://github.com/jdeamicis)'s comments

> what if this command fails? is the error handled already and the failure reason clear enough?
is there a limit to the number of nodes that these commands can handle?

check_command_output has try, catch and logging

src/common/schedulers/slurm_commands.py
hanwen-pcluste: Done

src/common/schedulers/slurm_commands.py
hanwen-pcluste: I removed the code. My code is only executed if if nodeaddrs or hostnames are are/is being changed. The input in this case should be list already

src/common/schedulers/slurm_commands.py
hanwen-pcluste: I refactored the code

src/common/schedulers/slurm_commands.py
hanwen-pcluste: We do not need it

src/common/schedulers/slurm_commands.py
hanwen-pcluste: We do not need it

src/common/schedulers/slurm_commands.py
```diff
@@ -104,17 +104,46 @@ def update_nodes(
...
         if hostnames:
-            node_info += f" nodehostname={hostnames}"
+            node_info += f" nodehostname={','.join(hostnames)}"
         run_command(  # nosec
             f"{update_cmd} {node_info}", raise_on_error=raise_on_error, timeout=command_timeout, shell=True
         )
 
 
+def _sort_nodes_attributes(nodes, nodeaddrs, nodehostnames):
+    nodes_str = ",".join(nodes) if type(nodes) is tuple else nodes
```
chenwany: Is there any case for nodes type to be not tuple?

hanwen-pcluste: It should be always tuple because it is the output of `def grouper(iterable, n):`

charlesg3 (approved):

PR #489, opened 2023-01-25
[3.x] Reducing risks associated to Bandit findings

maurizp:
### Description of changes
* Restrict the scope of the overrides related to Bandit findings
* Add a set of string validators
* Use the string validators to validate input arguments used in subprocess calls
* Fix emergent issue related to B017 findings in the flake8-bugbear module

### Tests
* manual tested the stop/start/stop fleet commands
* manual tested the job submissions on multiple partitions, verified logs and outputs
* launched integration tests ("multiple_job_submission" and "test_slurm")
* added unit tests for the utils.py class

### References
* https://bandit.readthedocs.io/en/latest/config.html#exclusions

### Checklist

### DIFF
--- a/src/common/schedulers/slurm_commands.py
+++ b/src/common/schedulers/slurm_commands.py
@@ -110,7 +110,8 @@ def update_nodes(
             node_info += f" nodeaddr={addrs}"
         if hostnames:
             node_info += f" nodehostname={hostnames}"
-        run_command(  # nosec
+        # It's safe to use the function affected by B604 since the command is fully built in this code
+        run_command(  # nosec B604
             f"{update_cmd} {node_info}", raise_on_error=raise_on_error, timeout=command_timeout, shell=True
         )
 
@@ -119,7 +120,7 @@ def update_partitions(partitions, state):
     succeeded_partitions = []
     for partition in partitions:
         try:
-            run_command(  # nosec
+            run_command(  # nosec B604
                 f"{SCONTROL} update partitionname={partition} state={state}", raise_on_error=True, shell=True
             )
             succeeded_partitions.append(partition)
@@ -238,7 +239,7 @@ def get_nodes_info(nodes="", command_timeout=DEFAULT_GET_INFO_COMMAND_TIMEOUT):
         'grep -oP "^(NodeName=\\S+)|(NodeAddr=\\S+)|(NodeHostName=\\S+)|(State=\\S+)|'
         '(Partitions=\\S+)|(Reason=.+) |(######)"'
     )
-    nodeinfo_str = check_command_output(show_node_info_command, timeout=command_timeout, shell=True)  # nosec
+    nodeinfo_str = check_command_output(show_node_info_command, timeout=command_timeout, shell=True)  # nosec B604
 
     return _parse_nodes_info(nodeinfo_str)
 
@@ -246,7 +247,8 @@ def get_nodes_info(nodes="", command_timeout=DEFAULT_GET_INFO_COMMAND_TIMEOUT):
 def get_partition_info(command_timeout=DEFAULT_GET_INFO_COMMAND_TIMEOUT, get_all_nodes=True):
     """Retrieve slurm partition info from scontrol."""
     show_partition_info_command = f'{SCONTROL} show partitions | grep -oP "^PartitionName=\\K(\\S+)| State=\\K(\\S+)"'
-    partition_info_str = check_command_output(show_partition_info_command, timeout=command_timeout, shell=True)  # nosec
+    # It's safe to use the function affected by B604 since the command is fully built in this code
+    partition_info_str = check_command_output(show_partition_info_command, timeout=command_timeout, shell=True)  # nosec B604
     partitions_info = _parse_partition_name_and_state(partition_info_str)
     return [
         SlurmPartition(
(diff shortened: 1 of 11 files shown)


src/common/utils.py
```diff
@@ -274,3 +274,23 @@ def read_json(file_path, default=None):
             if not isinstance(e, FileNotFoundError):
                 log.info("Unable to read file '%s' due to an exception: %s. Using default: %s", file_path, e, default)
             return default
+
+
+def validate_subprocess_argument(argument):
```
EddyMM, line 287: Future usage of `check_command_output` and `run_command` will result in the `B604`  bandit violation. We have a Github check as well to flag `nosec` and `nosemgrep` comments.

It might be worth mentioning in the docstring of those functions how to handle the violation i.e. use `validate_subprocess_argument` to validate any arguments.

tests/common/test_utils.py
```diff
@@ -91,3 +91,43 @@ def test_sleep_remaining_loop_time(mocker, loop_start_time, loop_end_time, loop_
...
+    [
+        ("standard parameter name", does_not_raise),
+        ("my/parameter", does_not_raise),
+        ("execute this & then this", pytest.raises(Exception)),
+        ("redirect | my output", pytest.raises(Exception)),
+        ("execute\nmultiline", pytest.raises(Exception)),
+    ],
+)
+def test_validate_subprocess_argument(argument, expectation):
+    with expectation:
```
EddyMM: We can use [assertpy](https://github.com/assertpy/assertpy) which is already included in the package to avoid using a NullContext class.


```
@pytest.mark.parametrize(
    "argument,raises_exception",
    [
        ("standard parameter name", False),
        ("my/parameter", False),
        ("execute this & then this", True),
        ("redirect | my output", True),
        ("execute\nmultiline", True),
    ],
)
if raises_exception:
    assert_that(utils.validate_subprocess_argument)
        .raises(Exception)
        .when_called_with(argument)
else:
    assert_that(utils.validate_subprocess_argument(argument)).is_true()
```
```

File: tests/common/test_utils.py [hunk 3]
```diff
@@ -91,3 +91,43 @@ def test_sleep_remaining_loop_time(mocker, loop_start_time, loop_end_time, loop_
...
+@pytest.mark.parametrize(
+    "argument,expectation",
+    [
+        ("/usr/my_path", does_not_raise),
+        ("./my_path", pytest.raises(Exception)),
+        ("my_path", pytest.raises(Exception)),
+        (".my_path", pytest.raises(Exception)),
+    ],
+)
+def test_validate_absolute_path(argument, expectation):
```
EddyMM: Same here.

src/common/utils.py
```diff
@@ -274,3 +274,23 @@ def read_json(file_path, default=None):
...
+    # As for https://www.gnu.org/software/bash/manual/bash.html#Lists
+    forbidden_patterns = ["&", "|",  ";",  "$",  ">", "<", "`", "\\", "!", "#", "\n"]
+
+    # Forcing the encoding to be the standard Python Unicode / UTF-8
+    # https://docs.python.org/3/howto/unicode.html
+    # https://docs.python.org/3/library/codecs.html#standard-encodings
+    _argument = (str(argument).encode("utf-8", "ignore")).decode()
+
+    if any(map(_argument.__contains__, forbidden_patterns)):
```
EddyMM: I'm curious, if `_argument` is a string at this point, why use `__contains__` instead of something like this?

```
any(pattern in _argument for pattern in forbidden_patterns)
```

Dunder methods are python's internal implementation. Should we stick to using exposed APIs  to avoid any surprises?

tests/common/test_utils.py
```diff
@@ -91,3 +91,49 @@ def test_sleep_remaining_loop_time(mocker, loop_start_time, loop_end_time, loop_
     elif expected_sleep_time == 0:
         sleep_mock.assert_not_called()
     datetime_now_mock.now.assert_called_with(tz=timezone.utc)
+
+
+class NullContext:
```
EddyMM: This can now be removed I think.

src/common/schedulers/slurm_commands.py
```diff
@@ -283,7 +297,8 @@ def _get_slurm_nodes(states=None, partition_name=None, command_timeout=DEFAULT_G
     if states:
         sinfo_command += f" -t {states}"
```
EddyMM, line 305: Have we validated the `partition_name` and `states` arguments here?

src/common/schedulers/slurm_commands.py
```diff
@@ -110,16 +110,21 @@ def update_nodes(
             node_info += f" nodeaddr={addrs}"
         if hostnames:
             node_info += f" nodehostname={hostnames}"
-        run_command(  # nosec
```
EddyMM, line 113: Have we validated the arguments used when building the command? E.g. `state`, `reason` etc?

EddyMM (approved):

src/common/schedulers/slurm_commands.py
```diff
@@ -101,25 +101,35 @@ def update_nodes(
...
 def update_partitions(partitions, state):
     succeeded_partitions = []
+    # Validation to sanitize the input argument and make it safe to use the function affected by B604
+    validate_subprocess_argument(state)
     for partition in partitions:
         try:
-            run_command(  # nosec
+            # Validation to sanitize the input argument and make it safe to use the function affected by B604
+            validate_subprocess_argument(partition)
+            run_command(  # nosec B604
```
gmarciani, line 132: Minor observation
We are getting a [B604](https://docs.openstack.org/bandit/1.4.0/plugins/any_other_function_with_shell_equals_true.html) violation because the function `run_command` has an argument named `shell`, but the actual violation is not here but in the `subprocess.run` call [here](https://github.com/aws/aws-parallelcluster-node/blob/develop/src/common/utils.py#L127-L127). 

Trivial solution as a further improvement not to address in this PR: renaming the argument `shell` in `run_commnad()` function so that the B604 violation will be in the strictly meaningful line, that is the `subprocess.run` call.

gmarciani (approved):

PR #490, opened 2023-01-30
Allow clustermgtd to handle compute nodes rebooted via `scontrol reboot`

jdeamicis:
### Description of changes
* Fix logic of clustermgtd to deal with EC2 status checks failures when rebooting compute nodes with `scontrol reboot`:
  * Do not consider nodes failing EC2 health checks as unhealthy while they are in Slurm state rebooting (`REBOOT_ISSUED`).
  * Do not consider nodes failing EC2 health checks as unhealthy for a certain time after Slurmd has started (EC2 may be late in updating the instance status checks).
  * Relates to https://github.com/aws/aws-parallelcluster/issues/4751

### Tests
* Added unit test scenarios to validate new logic.
* Added integration test to validate new logic:

### References
* https://github.com/aws/aws-parallelcluster/pull/4838

### Checklist


src/common/schedulers/slurm_commands.py
```diff
@@ -324,7 +329,7 @@ def _get_partition_nodes(partition_name, command_timeout=DEFAULT_GET_INFO_COMMAN
     return ",".join(nodes)
 
 
-def _parse_nodes_info(slurm_node_info):
+def _parse_nodes_info(slurm_node_info: str) -> List[SlurmNode]:
```
chenwany, line 332: Could you update the example below with `SlurmdStartTime` and `NextState`?

src/slurm_plugin/slurm_resources.py
```diff
@@ -82,6 +82,7 @@ def __eq__(self, other):
 
 
 class SlurmNode(metaclass=ABCMeta):
+    slurmdstarttime: datetime
```
chenwany: Instead of leaving it here, what about we change line 113 to be `def __init__(...., slurmdstarttime: datetime = None)` ?

tests/common/schedulers/test_slurm_commands.py
```diff
@@ -83,82 +86,122 @@ def test_is_static_node(nodename, expected_is_static):
...
+            ],
+            False,
+        ),
+        (
+            "NodeName=multiple-dy-c5xlarge-4\n"
+            "NodeAddr=multiple-dy-c5xlarge-4\n"
+            "NodeHostName=multiple-dy-c5xlarge-4\n"
+            "State=IDLE+CLOUD+POWER\n"
+            "Partitions=multiple,multiple2\n"
+            "Reason=(Code:InsufficientInstanceCapacity)Failure when resuming nodes \n"
```
chenwany: With the changes in code: `(Reason=.+) |(######)` to `(Reason=.*)|(######)` Looks like we are passing more information to Reasons, could you change this test case accordingly?

src/common/schedulers/slurm_commands.py
```diff
@@ -362,15 +367,21 @@ def _parse_nodes_info(slurm_node_info):
...
+    node_info = slurm_node_info.split("######\n")
     slurm_nodes = []
     for node in node_info:
-        lines = node.strip().splitlines()
+        lines = node.splitlines()
         kwargs = {}
         for line in lines:
             key, value = line.split("=")
+            if key == "SlurmdStartTime":
+                if value != "None":
```
chenwany, line 385: minor: line 381- 384 can be simplify into one line: `value = datetime.strptime(value, "%Y-%m-%dT%H:%M:%S").astimezone(tz=timezone.utc) if value != "None" else None`

src/common/utils.py
```diff
@@ -250,11 +250,13 @@ def convert_range_to_list(node_range):
     )
 
 
-def time_is_up(initial_time, current_time, grace_time):
+def time_is_up(initial_time: datetime, current_time: datetime, grace_time: int):
```
chenwany: Is it true that restrict grace_time is always int?
In `insufficient_capacity_timeout`, we allows input as float:
        self.insufficient_capacity_timeout = config.getfloat(
            "clustermgtd", "insufficient_capacity_timeout", fallback=self.DEFAULTS.get("insufficient_capacity_timeout")
        )

src/slurm_plugin/clustermgtd.py
```diff
@@ -956,17 +965,30 @@ def _handle_protected_mode_process(self, active_nodes, partitions_name_map):
...
+            if len(nodes_name_failing_health_check) > 0:
+                log.warning(
+                    "Setting nodes failing health check type %s to DRAIN: %s",
+                    health_check_type,
+                    nodes_name_failing_health_check,
+                )
+                set_nodes_drain(nodes_name_failing_health_check, reason=f"Node failing {health_check_type}")
+            if len(nodes_name_recently_rebooted) > 0:
+                log.info(
+                    "These nodes were recently restarted and they were not set to DRAIN: %s",
```
chenwany: What about changing `restarted` to rebooted?

src/common/schedulers/slurm_commands.py
jdeamicis: Done.

src/common/schedulers/slurm_commands.py
jdeamicis: Unfortunately the line becomes too long and the autoformatter splits it into even more lines that it's using now. I'm going to keep it as it is.

src/common/utils.py
jdeamicis: Changing it to `float`, which in Python typing should cover also the cases when an integer is passed (see [The numeric tower](https://peps.python.org/pep-0484/#the-numeric-tower)).

src/slurm_plugin/clustermgtd.py
jdeamicis: Minor discussion here: strictly speaking, this condition may occur even if someone manually restarts `slurmd` on a compute node, without completely rebooting the compute node. I believe I had originally written something like `On these nodes the slurmd daemon was recently restarted...`, and that's why I kept the restarted word here.

Also, this condition may also occur the first time ever a compute node is booted, so _rebooted_ or _restarted_ may be the wrong words. 

Should we write something like: `The slurmd service was recently started on these nodes and therefore they were not set to DRAIN`? Do we like it or is it too convoluted?

src/slurm_plugin/slurm_resources.py
jdeamicis: Done.

tests/common/schedulers/test_slurm_commands.py
jdeamicis: What do you mean here exactly?

src/slurm_plugin/slurm_resources.py
jdeamicis: Done.

src/common/utils.py
jdeamicis: Done.

tests/common/schedulers/test_slurm_commands.py
chenwany: with `scontrol show nodes` result:
`Reason=test [root@2023-01-31T21:24:55]`
Previously, we are using regex `(Reason=.+) |(######)`, with that expression, the match will be `Reason=test `
With your change, `(Reason=.*)|(######)`, the match will be `Reason=test [root@2023-01-31T21:24:55]`
Can we include that in this test case?

jdeamicis: Done.

chenwany (dismissed):

src/common/schedulers/slurm_commands.py
```diff
@@ -362,15 +371,21 @@ def _parse_nodes_info(slurm_node_info):
...
         "Partitions": "partitions",
         "Reason": "reason",
+        "SlurmdStartTime": "slurmdstarttime",
     }
 
-    node_info = slurm_node_info.split("######")
+    node_info = slurm_node_info.split("######\n")
     slurm_nodes = []
     for node in node_info:
-        lines = node.strip().splitlines()
```
demartinofra, line 370: why did you remove the strip()? it could be useful in case for any reason there are trailing whitespaces in the Slurm output

src/slurm_plugin/clustermgtd.py
```diff
@@ -956,17 +965,34 @@ def _handle_protected_mode_process(self, active_nodes, partitions_name_map):
...
+            if len(nodes_name_failing_health_check) > 0:
+                log.warning(
+                    "Setting nodes failing health check type %s to DRAIN: %s",
+                    health_check_type,
+                    nodes_name_failing_health_check,
+                )
+                set_nodes_drain(nodes_name_failing_health_check, reason=f"Node failing {health_check_type}")
+            if len(nodes_name_recently_rebooted) > 0:
+                log.info(
+                    "These nodes were recently rebooted and they were not set to DRAIN: %s",
```
demartinofra: let's slightly change this message to make it more explicit: Ignoring health check failure due to reboot for nodes:

src/common/schedulers/slurm_commands.py
jdeamicis: I just checked. Actually this has more to do with the output of our `SCONTROL_OUTPUT_AWK_PARSER` than anything coming from `scontrol show node`: this `strip()` was removing the trailing `\n` at the end each node line, which was there because we were splitting by `######` and not `######\n`.

Now, since we split according to `######\n`, it's not strictly necessary. Actually, the `strip()` in that position would remove trailing spaces **only** in the last field we parse in our awk parser, which currently is `Reason`. If in the future we start parsing additional fields (`Comment` or `Extra`, for example), the `strip()` will start affecting a field different than the reason.

Given this information, I would be more inclined towards not having the `strip()` at all.

demartinofra (approved):

src/slurm_plugin/clustermgtd.py
jdeamicis: Done.

PR #492, opened 2023-02-01
[Fix] Multi NIC IP Alignment

Nbot07A:
### Description of changes
* set ip to NIC 0 Device 0 primary ip to ensure IP alignment with Multi NIC instances

### Tests
* Passed new unit test test_from_describe_instance_data 
* Added multi-nic structure to test instances

### Checklist


src/slurm_plugin/fleet_manager.py
```diff
@@ -47,12 +47,15 @@ def __hash__(self):
...
-            )
+            for network_interface in instance_info["NetworkInterfaces"]:
+                attachment = network_interface["Attachment"]
+                if attachment["DeviceIndex"] == 0 and attachment["NetworkCardIndex"] == 0:
+                    return EC2Instance(
+                        instance_info["InstanceId"],
+                        network_interface["PrivateIpAddress"],
+                        instance_info["PrivateDnsName"].split(".")[0],
+                        instance_info["LaunchTime"],
+                    )
```
francesco-giordano: **Suggestion:** Can you refactor the code in order to address CodeQL issue.
An approach could be assign `instance_info["PrivateIpAddress"]` as a starting value to a variable. In the loop try to assign the right `network_interface["PrivateIpAddress"]` to that variable. Then use the variable to set the second field of the `return EC2Instance`.
In this way  there is the fallback of `instance_info["PrivateIpAddress"]` and a single explicit return.

francesco-giordano (commented):
Have all the unit test passed?

Nbot07A:
> Have all the unit test passed?

Yes

francesco-giordano (approved):
LGTM!

chenwany (commented):
We need to make sure the private ip here is the one return from ec2 describe-instances in multi-nic instances

chenwany (changes requested):
We need to update describe_instances together in the PR, or we will introduce another problem: during the iteration of clustermgtd, the multi-nic slurm node will be terminated due to ip mismatch.
We need to update here accordingly:
https://github.com/aws/aws-parallelcluster-node/blob/3d69d7ad5c1b43308d598a9433c820a8d800d0eb/src/slurm_plugin/clustermgtd.py#L1028

gmarciani:
To be done in this PR: we must align the PrivateIpAddress also here

https://github.com/aws/aws-parallelcluster-node/blob/3d69d7ad5c1b43308d598a9433c820a8d800d0eb/src/slurm_plugin/instance_manager.py#L345

src/slurm_plugin/fleet_manager.py
```diff
@@ -47,9 +47,14 @@ def __hash__(self):
     @staticmethod
     def from_describe_instance_data(instance_info):
         try:
+            private_ip = instance_info["PrivateIpAddress"]
+            for network_interface in instance_info["NetworkInterfaces"]:
+                attachment = network_interface["Attachment"]
+                if attachment["DeviceIndex"] == 0 and attachment["NetworkCardIndex"] == 0:
+                    private_ip = network_interface["PrivateIpAddress"]
```
davprat, line 54: Any reason not to break out of this loop if we find what we are looking for?

gmarciani: No reason, we should break

gmarciani:
Closed in favour of https://github.com/aws/aws-parallelcluster-node/pull/512

PR #504, opened 2023-03-07
3.x Add structured logging for key cluster events

davprat:
### Description of changes
* This change introduces the common code for publishing JSON structured log events.
* This change adds structured log events for node launch failures from both clustermgtd and resume
* We now generate structured log events for node launch failure levels, bootstrap failures, nodes with invalid backing instances, unresponsive nodes, and health check failures.
* This change adds the method `is_down_not_responding` to the base Slurm node resource class to test if Slurm set the node to DOWN state due to no ping response.
* This change adds the abstract method `is_bootstrap_timeout`  to the base Slurm node resource class to test if the node timed out waiting for bootstrap to succeed.

### Tests
* Added unit tests for common structured log publisher
* Updated unit tests for clustermgtd
* Manually deployed cluster into dev account.
  * Induced some failure modes: "node down, not responding", "terminated static node", launched a dynamic node with a bad custom action, and removed ec2 permissions from the head node.
  * Observed expected events being generated to CloudWatch

### Checklist


src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -0,0 +1,371 @@
...
+    LAUNCH_FAILURE_GROUPING.update({failure: "vcpu-limit-failures"})
+
+for failure in ["VolumeLimitExceeded", "InsufficientVolumeCapacity"]:
+    LAUNCH_FAILURE_GROUPING.update({failure: "volume-limit-failures"})
+
+for failure in ["InvalidBlockDeviceMapping"]:
+    LAUNCH_FAILURE_GROUPING.update({failure: "custom-ami-errors"})
+
+for failure in ["UnauthorizedOperation", "AccessDeniedException"]:
+    LAUNCH_FAILURE_GROUPING.update({failure: "iam-policy-errors"})
```
chenwany: nit: this can be simplify as
```
LAUNCH_FAILURE_GROUPING = {
    **{failure: "ice-failures" for failure in [* SlurmNode.EC2_ICE_ERROR_CODES, "LimitedInstanceCapacity"]},
    **{failure: "vcpu-limit-failures" for failure in ["VcpuLimitExceeded"]},
    **{failure: "volume-limit-failures" for failure in ["VolumeLimitExceeded", "InsufficientVolumeCapacity"]},
    **{failure: "custom-ami-errors" for failure in ["InvalidBlockDeviceMapping"]},
    **{failure: "iam-policy-errors" for failure in ["UnauthorizedOperation", "AccessDeniedException"]}
}
```

src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -0,0 +1,371 @@
...
+
+
+class ClusterEventPublisher:
+    def __init__(self, event_publisher: Callable, max_list_size=100):
```
chenwany: Can we add a comment to explain what this Class is for?

src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -0,0 +1,371 @@
...
+        self.publish_event(
+            "DEBUG",
+            "After node maintenance, node failed replacement",
+            event_type="node-launch-failure",
+            timestamp=timestamp,
+            event_supplier=self._failed_node_supplier(unhealthy_static_nodes, failed_nodes),
+        )
+
+        self.publish_event(
+            "DEGUG",
```
chenwany: there's a typo here :)

src/slurm_plugin/common.py
```diff
@@ -135,3 +137,67 @@ def is_clustermgtd_heartbeat_valid(current_time, clustermgtd_timeout, clustermgt
...
+                    )
+
+                    event_logger.log(log_level, "%s", json.dumps(dict(event)))
+                except Exception as e:
+                    logger.error("Failed to publish event: %s\n%s", e, traceback.format_exception(*sys.exc_info()))
+
+    return emit_event
+
+
+def event_publisher_noop(*args, **kwargs):
```
chenwany: Looks like this function is not defined, what is this function used for here?

src/slurm_plugin/common.py
```diff
@@ -135,3 +137,67 @@ def is_clustermgtd_heartbeat_valid(current_time, clustermgtd_timeout, clustermgt
...
+                        kwargs,
+                        {
+                            "datetime": now,
+                            "version": 0,
+                            "scheduler": "slurm",
+                            "cluster-name": cluster_name,
+                            "node-role": node_role,
+                            "component": component,
+                            "level": event_level,
+                            "instance-id": instance_id,
```
chenwany: question: Is this instance id here always the headnode instance id? Or will it be compute node instance id if the event is publish from the compute node?

src/slurm_plugin/resume.py
```diff
@@ -221,6 +227,17 @@ def main():
                 default_log_file,
                 e,
             )
+        global _event_publisher
+        _event_publisher = ClusterEventPublisher(
```
chenwany: Instead of setting global _event_publisher and initialize _event_publisher in the main function and use it in _resume, what about initialize it in `def _resume`, similar to what we do for instance manager?

src/slurm_plugin/resume.py
```diff
@@ -195,6 +200,7 @@ def _resume(arg_nodes, resume_config):
         )
         for error_code, node_list in instance_manager.failed_nodes.items():
             _handle_failed_nodes(node_list, reason=f"(Code:{error_code})Failure when resuming nodes")
+    _event_publisher.publish_node_launch_events(instance_manager.failed_nodes)
```
chenwany: Why do we put it outside `if failed_nodes`? What do we publish if there's no failed nodes?

src/slurm_plugin/common.py
```diff
@@ -135,3 +137,67 @@ def is_clustermgtd_heartbeat_valid(current_time, clustermgtd_timeout, clustermgt
...
+    logging.CRITICAL: "CRITICAL",
+    logging.ERROR: "ERROR",
+    logging.WARNING: "WARNING",
+    logging.INFO: "INFO",
+    logging.DEBUG: "DEBUG",
+    logging.NOTSET: "NOTSET",
+}
+
+
+def event_publisher(event_logger, cluster_name, node_role, component, instance_id, **global_args):
```
chenwany: Why do we need to put event_publisher in `common.py`, can it belongs to a function in class ClusterEventPublisher?

src/slurm_plugin/clustermgtd.py
```diff
@@ -717,6 +729,7 @@ def _find_unhealthy_slurm_nodes(self, slurm_nodes):
                     ).append(node)
                 else:
                     unhealthy_dynamic_nodes.append(node)
+        self._event_publisher.publish_unhealthy_node_events(all_unhealthy_nodes)
```
chenwany, line 730: currently, we are not publishing events when there's no unhealthy nodes since we did not set  kwargs in both clustermgtd and slurm_resume. for emit_event(event_level, message, event_type, timestamp=None, event_supplier=None, **kwargs).
Will it be better to add a check for `if all_unhealthy_nodes` so we can skip going through `invalid_backing_instance_nodes` and `publish_event` to make it more efficient?

src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -0,0 +1,371 @@
...
+        failed_nodes: Dict[str, List[str]],
+    ):
+        timestamp = ClusterEventPublisher.timestamp()
+
+        nodes_in_replacement = list(nodes_in_replacement)
+
+        self.publish_event(
+            "WARNING" if failed_nodes else "DEBUG",
+            "Number of static nodes that failed replacement after node maintenance",
+            event_type="node-launch-failure-count",
```
chenwany: if there's no failed nodes and the event will be debug level, should this event be some other event type than "node-launch-failure-count"?

src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -0,0 +1,371 @@
...
+        self,
+        unhealthy_static_nodes: List[SlurmNode],
+        nodes_in_replacement: Iterable[str],
+        failed_nodes: Dict[str, List[str]],
+    ):
+        timestamp = ClusterEventPublisher.timestamp()
+
+        nodes_in_replacement = list(nodes_in_replacement)
+
+        self.publish_event(
```
chenwany: It seems that this event is specified for launch instances error from the run_instance/create_fleet API. What about we only publish this event only when there's failed_nodes?

src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -0,0 +1,371 @@
...
+            event_supplier=self._node_count_in_replacement_supplier(nodes_in_replacement),
+        )
+
+        self.publish_event(
+            "DEBUG",
+            "After node maintenance, node currently in replacement",
+            event_type="static-node-in-replacement",
+            timestamp=timestamp,
+            event_supplier=self._node_in_replacement_supplier(
+                (node for node in unhealthy_static_nodes if node.name in nodes_in_replacement)
```
chenwany: Is this event try to represent all static nodes that are currently in replacement after relaunching unhealthy static ndoes? Why these are nodes in unhealthy_static_nodes? Should it be just`node.name in nodes_in_replacement`?

src/slurm_plugin/clustermgtd.py
```diff
@@ -911,6 +932,7 @@ def _handle_bootstrap_failure_nodes(self, active_nodes):
             log.warning("Found the following bootstrap failure nodes: %s", print_with_count(bootstrap_failure_nodes))
             # Increase the partition failure count and add failed nodes to the set of bootstrap failure nodes.
             self._increase_partitions_protected_failure_count(bootstrap_failure_nodes)
+        self._event_publisher.publish_bootstrap_failure_events(bootstrap_failure_nodes)
```
chenwany, line 934: What about moving the function inside ` if bootstrap_failure_nodes:`?

src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -0,0 +1,371 @@
...
+            event_supplier=detail_supplier(),
+        )
+
+    @log_exception(logger, "publish_bootstrap_failure_events", catch_exception=Exception, raise_on_error=False)
+    def publish_bootstrap_failure_events(self, bootstrap_failure_nodes: List[SlurmNode]):
+        def dynamic_node_filter(node: SlurmNode) -> bool:
+            return isinstance(node, DynamicNode) and node.is_resume_failed() and node.is_nodeaddr_set()
+
+        def static_node_filter(node: SlurmNode) -> bool:
+            return isinstance(node, StaticNode) and node._is_replacement_timeout
```
chenwany: nit: since this function `_is_replacement_timeout` is used outside the class, can we rename it to `is_replacement_timeout`?
For function used only inside the class, what about we add `_` at the beginning?

chenwany: Since these are the logic in is_bootstrap_failure, what about we instead define a function inside class slurm node `is_bootstrap_timeout(self)` that have different implementation for static node and dynamic node, and the function can be reused in `is_bootstrap_failure` and here?

src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -0,0 +1,371 @@
...
+            event_supplier=protected_mode_error_count_supplier(),
+        )
+
+        self.publish_event(
+            "DEBUG",
+            "Number of nodes that failed to bootstrap",
+            "bootstrap-failure-count",
+            timestamp=timestamp,
+            event_supplier=bootstrap_failure_supplier(),
+        )
```
chenwany: Instead of publishing two events here,
what about we only publish one warning event, that have "static-replacement-timeout-errors", "dynamic-resume-timeout-errors" and "other-bootstrap-errors" in `detail`?

src/slurm_plugin/common.py
```diff
@@ -135,3 +137,67 @@ def is_clustermgtd_heartbeat_valid(current_time, clustermgtd_timeout, clustermgt
...
+    def emit_event(event_level, message, event_type, timestamp=None, event_supplier=None, **kwargs):
+        log_level = (
+            event_to_log_level_mapping.get(event_level, logging.NOTSET) if isinstance(event_level, str) else event_level
+        )
+
+        if event_logger.isEnabledFor(log_level):
+            event_level = log_to_event_level_mapping.get(log_level, "NOTSET")
```
chenwany: I saw some examples of using logging.getLevelName and logging.addLevelName
`logging.getLevelName(log.getEffectiveLevel())`: https://www.programcreek.com/python/example/2908/logging.getLevelName
Not sure if they can be used here so we don't need the mapping?

src/slurm_plugin/common.py
davprat: It should be the instance ID of the publishing node - so yes, in the case of the compute node, it should be the compute node instance ID.

src/slurm_plugin/common.py
davprat: This is here so that the daemons that don't have a containing class (e.g. resume/suspend/computemgtd), can declare and use a global publisher that noops up until they can configure a real event publisher. Otherwise, there will be a lot of 
```
if _event_publisher:
    _event_publisher(...)
```
scattered throughout the code. So I figured it was just easier to initialize it with a noop so that always has a definition.

Since we are not yet using it anywhere with the your suggested change to  resume, I will remove it for now.

src/slurm_plugin/cluster_event_publisher.py
davprat: Yeah, I can do that. I was just trying to minimize changes to working code.

davprat: Yeah, I can do that. I was just trying to minimize changes to working code.

src/slurm_plugin/common.py
davprat: That's great , I will use that method - thank you! 
Unfortunately, the mapping for the other direction doesn't exist until v3.11.

src/slurm_plugin/resume.py
davprat: I will make the change since we are only using it in _resume for now.

src/slurm_plugin/cluster_event_publisher.py
davprat: Unless the log level is less than or equal to debug level, no event should be published in the case where failed_nodes is empty, or am I missing something?

src/slurm_plugin/cluster_event_publisher.py
davprat: At debug level, it should log out that there were 0 node launch failures, so I think the event type is correct?

src/slurm_plugin/cluster_event_publisher.py
davprat: the _nodes_in_replacement_supplier should be logging out the full node description, so it needs the SlurmNode rather than just the node.name.

src/slurm_plugin/clustermgtd.py
davprat: I updated the code in publish_unhealthy_node_events so that it will basically noop if all_unheathy_nodes is empty if the log level is higher than debug.

src/slurm_plugin/clustermgtd.py
davprat: As we spoke about, I would like to provide at least some mechanism to differentiate no errors from no signal. With the way it is now, a user can configure the log level to debug and make this differentiation. This might not be a valid design though, so let me know and I can change it.

src/slurm_plugin/cluster_event_publisher.py
davprat: Done, thank you!

src/slurm_plugin/cluster_event_publisher.py
davprat: Added.

src/slurm_plugin/cluster_event_publisher.py
davprat: Fixed 🤦

src/slurm_plugin/common.py
davprat: Removed.

src/slurm_plugin/cluster_event_publisher.py
davprat: Added a method to the SlurmNode class as you suggested below.

davprat: Yes, this is much cleaner - thank you!

src/slurm_plugin/cluster_event_publisher.py
davprat: Changed. There is now an "other" for other errors that are not bootstrap timeouts.

src/slurm_plugin/common.py
davprat: Incorporated suggestion.

src/slurm_plugin/cluster_event_publisher.py
chenwany: After our discussion, we align on publishing value 0 at debug level. 
What about changing the message here to something like: Numbers of static nodes that failed to launch backing instance after node maintenance

chenwany (commented):
LGTM! Can we add more comments to the main functions?

chenwany:
Question:
Do we want to have a single place in clustermgtd that handle publishing all log events? 
```
def publish_json_log_events(all_unhealthy_nodes, protected_mode_nodes,...):
  self._event_publisher.publish_nodes_failing_health_check_events(
            health_check_type, nodes_name_failing_health_check
        )
  self._event_publisher.publish_unhealthy_static_node_events(
      unhealthy_static_nodes,
      self._static_nodes_in_replacement,
      launched_nodes,
      self._instance_manager.failed_nodes,
  )
...
```

chenwany (dismissed):
LGTM!

chenwany (dismissed):
LGTM!

src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -0,0 +1,408 @@
...
+
+    @staticmethod
+    def timestamp():
+        return datetime.now(timezone.utc).isoformat(timespec="milliseconds")
+
+    @staticmethod
+    def create_with_default_publisher(
+        event_logger, cluster_name, node_role, component, instance_id, max_list_size=100, **global_args
+    ):
+        """Create an instance of CreateEventPublisher with the standard event publisher."""
```
delongmeng: “an instance of CreateEventPublisher”
do you mean "an instance of ClusterEventPublisher"?

delongmeng:
"Manually deployed cluster into dev account."
Can you describe in more detail how the manual test was done? What was expected vs. observed?

src/slurm_plugin/cluster_event_publisher.py
davprat: Thanks, fixed it.

chenwany (dismissed):

chenwany (dismissed):

src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -0,0 +1,407 @@
...
+    @staticmethod
+    def timestamp():
+        return datetime.now(timezone.utc).isoformat(timespec="milliseconds")
+
+    @staticmethod
+    def create_with_default_publisher(
+        event_logger, cluster_name, node_role, component, instance_id, max_list_size=100, **global_args
+    ):
+        """Create an instance of ClusterEventPublisher with the standard event publisher."""
+        publiser = ClusterEventPublisher._get_event_publisher(
```
lukeseawalker: typo on `publiser`

src/slurm_plugin/slurm_resources.py
```diff
@@ -217,6 +220,15 @@ def is_resume_failed(self):
         """Check if node resume timeout expires."""
         return self.states == self.SLURM_SCONTROL_RESUME_FAILED_STATE
 
+    def is_down_not_responding(self):
```
lukeseawalker, line 223: where this method is used?

src/slurm_plugin/clustermgtd.py
```diff
@@ -707,8 +714,11 @@ def _find_unhealthy_slurm_nodes(self, slurm_nodes):
         unhealthy_static_nodes = []
         unhealthy_dynamic_nodes = []
         ice_compute_resources_and_nodes_map = {}
+        all_unhealthy_nodes = []
         for node in slurm_nodes:
             if not node.is_healthy(self._config.terminate_drain_nodes, self._config.terminate_down_nodes):
+                all_unhealthy_nodes.append(node)
```
lukeseawalker, line 720: `all_unhealthy_nodes` will contains unhealthy static/dynamic nodes plus ice nodes. Is expected that also the latter are published as unhealthy?

src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -0,0 +1,407 @@
...
+            event_type="node-launch-failure",
+            timestamp=timestamp,
+            event_supplier=self._flatten_failed_launch_nodes(failed_nodes),
+        )
+
+    def _generate_launch_failure_details(self, failed_nodes: Dict[str, List[str]]):
+        """
+        Build a dictionary based on failure category (e.g. ice-failure).
+
+        The elements contain the the number of nodes and the of node names in that category.
```
lukeseawalker: two typos in the sentence

src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -0,0 +1,407 @@
...
+        launched_nodes: Iterable[str],
+        failed_nodes: Dict[str, List[str]],
+    ):
+        """Publish events for unhealthy static nodes."""
+        timestamp = ClusterEventPublisher.timestamp()
+
+        for count, error_detail in self._generate_launch_failure_details(failed_nodes):
+            self.publish_event(
+                logging.WARNING if count else logging.DEBUG,
+                "Number of static nodes that failed to launch a backing instance after node maintenance",
```
lukeseawalker: would be cleaner to set "message" and "event_type" in a constant?

src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -0,0 +1,407 @@
...
+
+        self.publish_event(
+            logging.DEBUG,
+            "After node maintenance, nodes currently in replacement",
+            event_type="static-nodes-in-replacement-count",
+            timestamp=timestamp,
+            event_supplier=self._node_list_and_count_supplier(nodes_in_replacement),
+        )
+
+        self.publish_event(
```
lukeseawalker, line 212: how does an event looks like once written into the .events file? can you provide an example?

src/slurm_plugin/clustermgtd.py
davprat: I think the idea is that the event publisher needs to publish different events for different types of unhealthy nodes and these may not divide up in the same way that clustermgtd needs to group them. So we decided just to hand off the list of all unhealthy nodes and let the event publisher do its own sorting. Of course in some cases this may lead to duplicated logic, but I figured it was better that way than to put any event specific logic into clustermgtd. What do you think?

src/slurm_plugin/slurm_resources.py
davprat: It is used by the event publisher to find nodes that Slurm set to `DOWN` due to being unable to communicate with the node's slurmd.

src/slurm_plugin/cluster_event_publisher.py
davprat: Fixed - thank you!

src/slurm_plugin/cluster_event_publisher.py
davprat: Fixed - thank you!

src/slurm_plugin/cluster_event_publisher.py
davprat: Yes it would - moved all event types and messages to constants.

src/slurm_plugin/cluster_event_publisher.py
davprat: I've added an example event in comments for each event method. Does that work?

lukeseawalker: perfect! thanks

lukeseawalker (dismissed):
at high level, my concerns are about why the event class should know how to get the info of the object it needs to log (SlurmNode, StaticNode Vs Dynamic). I was expecting its interface to be scheduler agnostic. But than I see that all the "writing actors" are actually Slurm actors (clustermgtd, slurm resume) so probably wouldn't make sense to have this separation at the moment. For me we can move forward for now

chenwany (approved):

PR #526, opened 2023-05-29
Read SLURM_RESUME_FILE from resume program

lukeseawalker:
### Description of changes
Read the content of the SLURM_RESUME_FILE and make it available to the instance manager.

### Tests
unit tests added

### References
https://github.com/aws/aws-parallelcluster-cookbook/pull/2190

### Checklist

### DIFF
--- a/src/slurm_plugin/instance_manager.py
+++ b/src/slurm_plugin/instance_manager.py
@@ -66,6 +66,7 @@ class InstanceManager:
         fleet_config=None,
         run_instances_overrides=None,
         create_fleet_overrides=None,
+        slurm_resume=None,
     ):
         """Initialize InstanceLauncher with required attributes."""
         self._region = region
@@ -85,6 +86,7 @@ class InstanceManager:
         self._boto3_resource_factory = lambda resource_name: boto3.session.Session().resource(
             resource_name, region_name=region, config=boto3_config
         )
+        self.slurm_resume = slurm_resume
 
     def _clear_failed_nodes(self):
         """Clear and reset failed nodes list."""
--- a/src/slurm_plugin/resume.py
+++ b/src/slurm_plugin/resume.py
@@ -148,7 +148,7 @@ def _handle_failed_nodes(node_list, reason="Failure when resuming nodes"):
         log.error("Failed to place nodes %s into down with exception: %s", print_with_count(node_list), e)
 
 
-def _resume(arg_nodes, resume_config):
+def _resume(arg_nodes, resume_config, slurm_resume):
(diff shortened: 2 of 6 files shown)


src/slurm_plugin/resume.py
```diff
@@ -238,12 +239,20 @@ def main():
...
+
+        _resume(args.nodes, resume_config, _get_slurm_resume())
         log.info("ResumeProgram finished.")
     except Exception as e:
         log.exception("Encountered exception when requesting instances for %s: %s", args.nodes, e)
         _handle_failed_nodes(args.nodes)
 
 
+def _get_slurm_resume():
+    slurm_resume = read_json(os.environ.get("SLURM_RESUME_FILE"), default={})
```
EddyMM, line 251: What happens when the ENV VAR `SLURM_RESUME_FILE` is not set?
`os.environ.get` will probably return `None`. How does `read_json` handle `filepath=None`? It's likely to fail with a `TypeError` and return the default of `{}` as specified here but will we get any error logs to debug this case?

Given the importance of having `SLURM_RESUME_FILE` set, perhaps we can fetch it using `os.environ["SLURM_RESUME_FILE"]` with a try catch (KeyError) and log that `SLURM_RESUME_FILE` is not set.

src/slurm_plugin/instance_manager.py
```diff
@@ -66,6 +66,7 @@ def __init__(
         fleet_config=None,
         run_instances_overrides=None,
         create_fleet_overrides=None,
+        slurm_resume=None,
```
EddyMM, line 69: Given that the use of `slurm_resume` config will alter the behaviour of the InstanceManager (in a way that's different from how `clustermgtd` works), does it make sense to have a subclasss of InstanceManager dedicated to handling Resume events (instead of using an optional parameter in the `InstanceManager` class)?

I'm okay with this change if subclassing is something that might be considered in the subsequent PRs (since it's too early to know the exact changes will be done to the InstanceManager).

src/slurm_plugin/resume.py
lukeseawalker: Case when SLURM_RESUME_FILE isn't set is already covered, log will say "Unable to read file [...]". I will add a test case

src/slurm_plugin/instance_manager.py
lukeseawalker: Sure, this will be considered in other PRs. This is covering only the read of the SLURM_RESUME_FILE

EddyMM (approved):

PR #527, opened 2023-06-06
[develop] Job Level Scaling for exclusive nodes

lukeseawalker:
### Description of changes
Perform job level scaling for exclusive jobs, by reading at job information from `SLURM_RESUME_FILE`. 

Job objects are built from SLURM_RESUME_FILE content, and split into two list, based on the value of the Oversubscribe property.

Nodes for job with Oversubscribe=NO are launched on a per job basis, while nodes for job with Oversubscribe!=NO are launched all together in single call.

A new `job_level_scaling` resume configuration parameter is added, so to allow to switch back from job level scaling to node list scaling

### Tests
* unit tests added
* manual tests performed
* integration tests will be added in another PR (on CLI code)

### References
* https://github.com/aws/aws-parallelcluster-node/pull/526

### Checklist

### DIFF
--- a/src/slurm_plugin/instance_manager.py
+++ b/src/slurm_plugin/instance_manager.py
@@ -16,13 +16,13 @@ import logging
 # A nosec comment is appended to the following line in order to disable the B404 check.
 # In this file the input of the module subprocess is trusted.
 import subprocess  # nosec B404
-from typing import Iterable
+from typing import Dict, Iterable, List
 
 import boto3
 from botocore.config import Config
 from botocore.exceptions import ClientError
 from common.ec2_utils import get_private_ip_address
-from common.schedulers.slurm_commands import update_nodes
+from common.schedulers.slurm_commands import get_nodes_info, update_nodes
 from common.utils import grouper
 from slurm_plugin.common import ComputeInstanceDescriptor, log_exception, print_with_count
 from slurm_plugin.fleet_manager import EC2Instance, FleetManagerFactory
@@ -33,6 +33,8 @@ from slurm_plugin.slurm_resources import (
     EC2InstanceHealthState,
     InvalidNodenameError,
     SlurmNode,
+    SlurmResumeData,
+    SlurmResumeJob,
     parse_nodename,
 )
 
@@ -66,7 +68,6 @@ class InstanceManager:
         fleet_config=None,
         run_instances_overrides=None,
         create_fleet_overrides=None,
-        slurm_resume=None,
     ):
         """Initialize InstanceLauncher with required attributes."""
         self._region = region
@@ -86,18 +87,45 @@ class InstanceManager:
         self._boto3_resource_factory = lambda resource_name: boto3.session.Session().resource(
             resource_name, region_name=region, config=boto3_config
         )
-        self.slurm_resume = slurm_resume
 
     def _clear_failed_nodes(self):
         """Clear and reset failed nodes list."""
         self.failed_nodes = {}
 
-    def add_instances_for_nodes(
+    def add_instances(
+        self,
+        node_list,
+        launch_batch_size,
+        update_node_address=True,
+        all_or_nothing_batch=False,
+        job_level_scaling=False,
+        slurm_resume=None,
+    ):
+        # Reset failed_nodes
+        self._clear_failed_nodes()
+
+        if slurm_resume and job_level_scaling:
+            logger.debug("Job Level Scaling using Slurm Resume File")
+            self._add_instances_for_resume_file(
+                slurm_resume=slurm_resume,
+                node_list=node_list,
+                launch_batch_size=launch_batch_size,
+                update_node_address=update_node_address,
+                all_or_nothing_batch=all_or_nothing_batch,
+            )
+        else:
+            logger.debug("Node Scaling using Slurm Node Resume List")
+            self._add_instances_for_nodes(
+                node_list=node_list,
+                launch_batch_size=launch_batch_size,
+                update_node_address=update_node_address,
+                all_or_nothing_batch=all_or_nothing_batch,
+            )
+
+    def _add_instances_for_nodes(
         self, node_list, launch_batch_size, update_node_address=True, all_or_nothing_batch=False
     ):
         """Launch requested EC2 instances for nodes."""
-        # Reset failed_nodes
-        self._clear_failed_nodes()
         instances_to_launch = self._parse_requested_instances(node_list)
         for queue, compute_resources in instances_to_launch.items():
             for compute_resource, slurm_node_list in compute_resources.items():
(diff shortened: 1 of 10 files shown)


src/slurm_plugin/instance_manager.py
```diff
@@ -246,6 +297,68 @@ def _update_dns_hostnames(self, nodes):
                 )
         logger.info("DNS records update: COMPLETED")
 
+    def _get_slurm_jobs_and_nodes(
+        self, slurm_resume: Dict[str, any], node_list: List[str]
+    ) -> Tuple[Set[str], List[str], Set[str], List[str]]:
```
EddyMM: I think the return type fo this function should be: `tuple[list[Any], list[Any], list[Any], list[Any]]`

Non-Oversubscribed nodes is a list.

src/slurm_plugin/instance_manager.py
```diff
@@ -246,6 +297,68 @@ def _update_dns_hostnames(self, nodes):
...
+        """
+        Parse Slurm Resume File.
```
EddyMM: Can we include the format (or sample) of the Slurm Resume File in this DocString?

src/slurm_plugin/slurm_resources.py
```diff
@@ -81,6 +81,64 @@ def __eq__(self, other):
...
+    @classmethod
+    # Default to OK when Oversubscribe value cannot be parsed
+    def _missing_(cls, value):
+        return cls.OK
+
+    def __str__(self):
+        return str(self.value)
+
+
+class SlurmJob:
```
EddyMM: Thess attributes are specific to the format in the `SLURM_RESUME_FILE`. Can we call this class `SlurmResumeJob` instead? To avoid giving the impression that this is the standard format for all Slurm Jobs.

src/slurm_plugin/slurm_resources.py
```diff
@@ -81,6 +81,64 @@ def __eq__(self, other):
...
+        self.features = features
+        self.extra = extra
+
+    def is_exclusive(self):
+        return self.oversubscribe == JobOversubscribe.NO
+
+    def __eq__(self, other):
+        """Compare 2 SlurmJob objects."""
+        if isinstance(other, SlurmJob):
+            return self.__dict__ == other.__dict__
```
EddyMM, line 137: Is there any risk that `__dict__` might include other keys that end up making 2 jobs non-equal despite them being equal for our use-case?
If so, we could consider specifying the exact attributes that would classify that two jobs as equal?

src/slurm_plugin/instance_manager.py
```diff
@@ -246,6 +297,68 @@ def _update_dns_hostnames(self, nodes):
...
+
+    def _parse_slurm_resume(self, slurm_resume: Dict[str, any]) -> List[SlurmJob]:
+        slurm_resume_jobs = []
+        for job in slurm_resume.get("jobs", {}):
+            try:
+                slurm_resume_jobs.append(SlurmJob(**job))
+            except InvalidNodenameError:
+                nodes_resume = job.get("nodes_resume", "")
+                job_id = job.get("job_id", "")
+                logger.warning(
```
EddyMM, line 381: `InvalidNodenameError` could also be raised if the `nodes_alloc` is using an invalid name. Let's include it in this log message as well.

src/slurm_plugin/instance_manager.py
```diff
@@ -246,6 +297,68 @@ def _update_dns_hostnames(self, nodes):
...
+                    "Discarding NodeNames with invalid format for Job Id (%s): %s",
+                    job_id,
+                    nodes_resume,
+                )
+                self._update_failed_nodes(
+                    # if NodeNames in nodes_resume cannot be parsed, try to get info directly from Slurm
+                    set([node.name for node in get_nodes_info(nodes_resume)]),
```
EddyMM, line 389: `nodes_alloc` as well.

src/slurm_plugin/instance_manager.py
EddyMM: For readability, we could use a DataClass instead.

src/slurm_plugin/instance_manager.py
```diff
@@ -246,6 +297,68 @@ def _update_dns_hostnames(self, nodes):
...
+                jobs_no_oversubscribe.append(job)
+                nodes_no_oversubscribe.extend(job.nodes_resume)
+            else:
+                jobs_oversubscribe.append(job)
+                nodes_oversubscribe.extend(job.nodes_resume)
+
+        nodes_difference = list(set(node_list) - (set(nodes_oversubscribe) | (set(nodes_no_oversubscribe))))
+        if nodes_difference:
+            logger.warning(
+                "Discarding NodeNames because of mismatch in Slurm Resume File Vs NodeNames: %s",
```
EddyMM: NIT: `... mismatch in Slurm Resume File Vs Nodes passed to Resume Program`

tests/slurm_plugin/slurm_resources/test_slurm_resources.py
```diff
@@ -1194,3 +1202,360 @@ def test_fail_ec2_health_check(instance_health_state, current_time, expected_res
 )
 def test_fail_scheduled_events_health_check(instance_health_state, expected_result):
     assert_that(instance_health_state.fail_scheduled_events_check()).is_equal_to(expected_result)
+
+
+@pytest.mark.parametrize(
```
EddyMM, line 1207: Let's add a case combining static and dynamic nodes.

tests/slurm_plugin/slurm_resources/test_slurm_resources.py
```diff
@@ -1194,3 +1202,360 @@ def test_fail_ec2_health_check(instance_health_state, current_time, expected_res
...
+                    "reservation": None,
+                },
+                ["queue1-st-c5xlarge-1"],
+                ["queue1-st-c5xlarge-1"],
+                False,
+                False,
+            ),
+        ],
+    )
+    def test_slurm_job(
```
EddyMM, line 1391: Let's include a broken `nodes_alloc`.

src/slurm_plugin/instance_manager.py
lukeseawalker: sure

src/slurm_plugin/slurm_resources.py
lukeseawalker: sure, my idea was to use this to store also other information, so initially I thought about building an hierarchy, but was being overkill at this stage of the implementation

src/slurm_plugin/instance_manager.py
lukeseawalker: * return type fixed
* ~~for using dataclass, it doesn't allow me to override the type of the input attribute~~ dataclass added

src/slurm_plugin/slurm_resources.py
lukeseawalker: for now I'm happy to consider two jobs different if at least one attribute is different

src/slurm_plugin/instance_manager.py
lukeseawalker: done

src/slurm_plugin/instance_manager.py
lukeseawalker: no, because in node alloc there could be nodes running. Only node not able to be resumed need to be set into DOWN

src/slurm_plugin/instance_manager.py
lukeseawalker: done

tests/slurm_plugin/slurm_resources/test_slurm_resources.py
lukeseawalker: done

tests/slurm_plugin/slurm_resources/test_slurm_resources.py
lukeseawalker: done

EddyMM (approved):

PR #528, opened 2023-06-07
Fix FleetManager behavior when create-fleet returns no instances

NSsirena:
### Description of changes
Fix FleetManager behavior when create-fleet returns no instances

In case of ICE create-fleet doesn't raise an exception and the default behavior isn't triggered This change forces the failure reason to ICE when create-fleet doesn't return any instance

* Added new behavior when create-fleet returns no instances
* Changed unit test according to new behavior
* Updated Changelog

### Tests
* Unit tests executed locally.
* Integration test will be extended on the CLI.

### References
* TODO: Add CLI integ-test PR when ready.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,8 @@ This file is used to list changes made in each version of the aws-parallelcluste
 3.6.1
 ------
 
-**CHANGES**
-- There were no changes for this version.
+**BUG FIXES**
+- Fix fast insufficient capacity fail-over logic when using Multiple Instance Types and no instances are returned
 
 3.6.0
 ------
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,12 @@ This file is used to list changes made in each version of the aws-parallelcluste
 
 **BUG FIXES**
 
+3.6.1
+------
+
+**BUG FIXES**
+- Fix Fast Failover when using Multiple Instance Types and no instances are returned
```
lukeseawalker: mind rewording this to better identify the feature fixed?
In the changelog it's referred as "fast insufficient capacity fail-over"

lukeseawalker (commented):
I think we need to add test coverage in exception management of the EC2 fleet case, i.e.

https://app.codecov.io/gh/aws/aws-parallelcluster-node/pull/429/blob/src/slurm_plugin/fleet_manager.py#L303

for example, improve the test coverage for:
* when no instances are returned we do not expect an exception to be raised
* exception is raised

lukeseawalker (approved):

PR #530, opened 2023-06-09
Push instance-to-node mapping from clustermgtd to log file

Jennifer603:
### Description of changes
* Added a new logger nodemap
* Added a event publisher that uses the new logger
* Add a new publisher function to publish the instance_to_node information
* Added command in clustermgtd.py to call this new method

### Tests
* Created a unit test to test the publish_nodemapping_event
* The unit test contains situations: one node in nodemap does not have an instance, empty nodemap, and node map does not log warning.

### Checklist


src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -509,6 +513,47 @@ def publish_node_launch_events(self, failed_nodes: Dict[str, List[str]]):
...
+    #     "scheduler": "Slurm",
+    #     "message": <str>,
+    #     "detail": {
+    #         # instance to node mapping schema
+    #         "node_list":[
+    #             {
+    #                 "instance_id":"i-123",
+    #                 "node_name": "string",
+    #                 "instance_type": "string",
+    #                 "disable_multi_threading": "bool"
```
delongmeng: should be `threads_per_core`?

src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -597,6 +642,21 @@ def _failed_node_supplier(unhealthy_static_nodes: List[SlurmNode], failed_nodes:
                         }
                     }
 
+    def _nodemapping_supplier(self, node_map: List[SlurmNode]) -> Iterator:
+        for node in node_map:
```
delongmeng: `node_map` -> should just be `nodes`? or something like `nodes_to_map`?

delongmeng (commented):
LGTM
please merge into a feature branch instead of develop

src/slurm_plugin/cluster_event_publisher.py
Jennifer603: Done!

src/slurm_plugin/cluster_event_publisher.py
Jennifer603: Done!

src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -597,6 +642,21 @@ def _failed_node_supplier(unhealthy_static_nodes: List[SlurmNode], failed_nodes:
                         }
                     }
 
+    def _nodemapping_supplier(self, nodes: List[SlurmNode]) -> Iterator:
```
chenwany: Is this a static method? Could you add `@staticmethod`?

chenwany: could you separate nodemapping to node_mapping

src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -597,6 +642,21 @@ def _failed_node_supplier(unhealthy_static_nodes: List[SlurmNode], failed_nodes:
...
+        for node in nodes:
+            yield {
+                "detail": {
+                    "nodes": [
```
chenwany: With this we are creating one detail for each node and only have a single node in the `nodes` list , this is not what we expect to have.
Could you fix this by including all the nodes into the `nodes` list?

chenwany: After changing this, could you add an example test output of the json event in the PR description?

src/slurm_plugin/cluster_event_publisher.py
Jennifer603: Done

davprat (dismissed):
Looks good!

davprat (commented):
Actually, you should have some formatting errors as reported by tox, so run tox -e autoformat and update your PR, then I can approve it.

src/slurm_plugin/cluster_event_publisher.py
Jennifer603: done

PR #533, opened 2023-06-13
[release-3.6] Fix FleetManager behavior when create-fleet returns no instances

NSsirena:
In case of ICE create-fleet doesn't raise an exception and the default behavior isn't triggered This change forces the failure reason to ICE when create-fleet doens't return any instance

### Description of changes
* Describe *what* you're changing and *why* you're doing these changes.

### Tests
* Describe the automated and/or manual tests executed to validate the patch.
* Describe the added/modified tests.

### References
* [PR for develop branch](https://github.com/aws/aws-parallelcluster-node/pull/528)

### Checklist

### DIFF
--- a/src/slurm_plugin/instance_manager.py
+++ b/src/slurm_plugin/instance_manager.py
@@ -167,11 +167,20 @@ class InstanceManager:
                     print_with_count(zip(launched_nodes, launched_instances)),
                 )
             if fail_launch_nodes:
-                logger.warning(
-                    "Failed to launch instances due to limited EC2 capacity for following nodes: %s",
-                    print_with_count(fail_launch_nodes),
-                )
-                self._update_failed_nodes(set(fail_launch_nodes), "LimitedInstanceCapacity")
+                if launched_nodes:
+                    logger.warning(
+                        "Failed to launch instances due to limited EC2 capacity for following nodes: %s",
+                        print_with_count(fail_launch_nodes),
+                    )
+                    self._update_failed_nodes(set(fail_launch_nodes), "LimitedInstanceCapacity")
+                else:
+                    # EC2 Fleet doens't trigger any exception in case of ICEs and may return more than one error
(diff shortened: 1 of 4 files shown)


src/slurm_plugin/instance_manager.py
```diff
@@ -167,11 +167,20 @@ def _update_slurm_node_addrs(self, slurm_nodes, launched_instances):
...
+                    logger.warning(
+                        "Failed to launch instances due to limited EC2 capacity for following nodes: %s",
+                        print_with_count(fail_launch_nodes),
+                    )
+                    self._update_failed_nodes(set(fail_launch_nodes), "LimitedInstanceCapacity")
+                else:
+                    # EC2 Fleet doens't trigger any exception in case of ICEs and may return more than one error
+                    # for each request. So when no instances were launched we force the reason to ICE
+                    logger.error(
+                        "Failed to launch instances due to limited EC2 capacity for following nodes: %s",
```
gmarciani, line 180: Just for my better understanding: in this way will we surface an ICE failure even if it may be something else? Is this kind of a trick to make fats failover work?

NSsirena: Yes.
CreateFleet has a very complex and rich behavior and in many cases where Run Instances will return an ICE, it just returns an error. Problem is that it may return a list of errors (triggered while attempting to provision different instances) but without information of how to match an error to an Instance.
As of now we do not have a mechanism to associate more than one error code to each instance so we agreed to consider an empty list of provisioned instances as a generic capacity issue (ICE) even if there may be different causes for different instances.
There is room for improvement but will require further investigation.

gmarciani: Understood, thanks!

gmarciani (approved):

gmarciani (approved):

gmarciani (approved):

PR #536, opened 2023-06-15
Run jobinfomgtd to collect job information to a local file

chenwany:
### Description of changes
* Run jobinfomgtd to collect job info json log to a local file
* The daemon is running each 3 minutes

### Tests
* Manually test with enable and disable job information collection in cookbook.
* Example output of the json event
```
    # {
    #     "datetime": "2023-06-15T02:49:12.749+00:00",
    #     "version": 0,
    #     "scheduler": "slurm",
    #     "cluster-name": "gpu-cluster2",
    #     "node-role": "HeadNode",
    #     "component": "jobinfomgtd",
    #     "level": "INFO",
    #     "instance-id": "i-0332623ed854e713f",
    #     "event-type": "scontrol-show-job-information",
    #     "message": "Job information from scontrol",
    #     "detail": {
    #         "job_id": "29",
    #         "job_name": "wrap",
    #         "user_id": "root(0)",
    #         "account": "root",
    #         "job_state": "COMPLETED",
    #         "run_time": "00:00:30",
    #         "start_time": "2023-06-15T02:46:10.000+00:00",
    #         "end_time": "2023-06-15T02:46:40.000+00:00",
    #         "partition": "compute",
    #         "node_list": "compute-dy-defaultcompute-1",
    #         "nodes": ["compute-dy-defaultcompute-1"],
    #         "cpu_ids": ["0"], 
    #         "gres": []
    #     }
    # }
```

### References

### Checklist


src/common/schedulers/slurm_commands.py
```diff
@@ -63,11 +64,18 @@
     + "grep -oP '^(NodeName=\\S+)|(NodeAddr=\\S+)|(NodeHostName=\\S+)|(?<!Next)(State=\\S+)|"
     + "(Partitions=\\S+)|(SlurmdStartTime=\\S+)|(LastBusyTime=\\S+)|(Reason=.*)|(######)'"
 )
+SCONTROL_SHOW_JOB_OUTPUT_AWK_PARSER = (
+    'awk \'BEGIN{{RS="\\n\\n" ; ORS="######\\n";}} {{print}}\' | '
+    + "grep -oP '^(JobId=\\S+)|(JobName=\\S+)|(UserId=\\S+)|(Account=\\S+)|(JobState=\\S+)|(RunTime=\\S+)|"
+    + "(StartTime=\\S+)|(EndTime=\\S+)|(Partition=\\S+)|(\\bNodeList=\\S+)|(\\bNodes=\\S+)|"
+    + "(CPU_IDs=\\S+)|(\\bGres=\\S+)|(######)'"
+)
```
davprat, line 72: I know there are potential issues with having scontrol returning JSON, but have you tried it to see if it might work for this case? I think there is some effort in the scheduler team to get away from using text parsing on scontrol and switching to JSON - it might be worth talking to them if you haven't done so already.

src/common/schedulers/slurm_commands.py
```diff
@@ -403,3 +422,170 @@ def _parse_nodes_info(slurm_node_info: str) -> List[SlurmNode]:
...
+
+
+def _parse_job_info(job_info_string):
+    """Parse job info into job objects."""
+    # Example input
+    # [ec2-user@ip-192-168-8-106 ~]$ sudo /opt/slurm/bin/scontrol show jobs --details | awk 'BEGIN{{RS="\n\n" ;
+    # ORS="######\n";}} {{print}}' | grep -oP '^(JobId=\S+)|(JobName=\S+)|(UserId=\S+)|(Account=\S+)|(JobState=\S+)|
+    # (RunTime=\S+)|(StartTime=\S+)|(EndTime=\S+)|(Partition=\S+)|(\\bNodeList=\S+)|(\\bNodes=\S+)|(CPU_IDs=\S+)|
+    # (\\bGres=\S+)|(######)'
+    # JobId=39
```
davprat: I'd say reduce this to 2 or 3 examples unless you think they are all necessary?

src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -100,6 +100,14 @@
     "message": "Count of EC2 instances backing the cluster compute fleet",
     "event_type": "cluster-instance-count",
 }
+SCONTROL_SHOW_JOB_INFORMATION = {
+    "message": "Job information from scontrol",
+    "event_type": "scontrol-show-job-information",
```
davprat, line 109: I'm not sure we should include `scontrol` in the event name - maybe something more generic like just "job-information" or "scheduler-job-information"?

src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -100,6 +100,14 @@
...
+}
+SCONTROL_SHOW_JOB_COUNT = {
+    "message": "Count of the jobs from scontrol",
+    "event_type": "scontrol_show_job_count",
```
davprat: Same as above for this - maybe keep the name more generic rather than mechanism specific. Also replace `_` with `-`

src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -509,6 +517,52 @@ def publish_node_launch_events(self, failed_nodes: Dict[str, List[str]]):
...
+    def publish_job_events(self, jobs: List[SlurmJobInfo]):
+        """Publish events for job information."""
+        timestamp = ClusterEventPublisher.timestamp()
+        self.publish_event(
+            logging.DEBUG,
+            **SCONTROL_SHOW_JOB_COUNT,
+            timestamp=timestamp,
+            event_supplier=self._job_count_supplier(jobs),
+        )
+        for job in jobs:
```
davprat: I would push this loop into your `_job_info_supplier` method so that it will only loop if the log level is not filtered.

src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -509,6 +517,52 @@ def publish_node_launch_events(self, failed_nodes: Dict[str, List[str]]):
...
+            self.publish_event(
+                logging.INFO,
+                **SCONTROL_SHOW_JOB_INFORMATION,
+                timestamp=timestamp,
+                event_supplier=self._job_info_supplier(job),
+            )
+
+    # Example event generated by this function:
```
davprat, line 565: Could you move the example to above the method definition?

src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -509,6 +517,52 @@ def publish_node_launch_events(self, failed_nodes: Dict[str, List[str]]):
             event_supplier=self._flatten_failed_launch_nodes(failed_nodes),
         )
 
+    @log_exception(logger, "publish_job_info_events", catch_exception=Exception, raise_on_error=False)
```
davprat: Should the name in log_exception match the method name?

src/slurm_plugin/cluster_event_publisher.py
```diff
@@ -775,3 +829,32 @@ def _describe_instance(instance):
             if instance
             else None
         )
+
+    @staticmethod
+    def _job_info_supplier(job: SlurmJobInfo) -> Iterator:
+        yield {
```
davprat: As commented above, I think you should iterate over the job list in here, yielding the detail object on every iteration.

src/slurm_plugin/jobinfomgtd.py
```diff
@@ -0,0 +1,184 @@
...
+        # Configure root logger
+        try:
+            fileConfig(config.logging_config, disable_existing_loggers=False)
+        except Exception as e:
+            log.warning(
+                "Unable to configure logging from %s, using default logging settings.\nException: %s",
+                config.logging_config,
+                e,
+            )
+        job_info_manager.process_job_info()
```
davprat: Do we want to terminate if `process_job_info` raises an excepiton?

src/slurm_plugin/jobinfomgtd.py
```diff
@@ -0,0 +1,184 @@
...
+
+def _minutes(min):
+    """Convert minutes to seconds."""
+    return min * 60
+
+
+def _write_json_to_file(filename, json_data):
+    """Write json to file."""
+    with open(filename, "w", encoding="utf-8") as file:
+        file.write(json.dumps(json_data))
```
davprat: Are these being used? I don't see them called from anywhere?

src/slurm_plugin/logging/parallelcluster_clustermgtd_logging.conf
```diff
@@ -45,7 +45,7 @@ qualname=slurm_plugin.clustermgtd.events
 format=%(message)s
 
 [handler_eventsHandler]
-class=FileHandler
+class=FileHandler a
 level=INFO
```
davprat: Is the `a` intentional?

davprat (commented):
Looks good - I would just talk to the scheduler team about using the same text parsing that we do for node info and see if we can't use JSON output instead and also move the job iteration loop into your event supplier.

src/common/schedulers/slurm_commands.py
chenwany: The problem with this case is that the scontrol json doesn't have the cpu_id information unfortunately, we are not able to use that

src/common/schedulers/slurm_commands.py
chenwany: yeah, updated the code

src/slurm_plugin/cluster_event_publisher.py
chenwany: The reason is that now there's another cluster SlurmJob that represent the jobs information from slurm resume for job level based scaling. In case there will be json log passing from there, I rename the jobs info from scontrol shows jobs output to avoid confusion

src/slurm_plugin/cluster_event_publisher.py
chenwany: Good idea, changed

src/slurm_plugin/cluster_event_publisher.py
chenwany: Good catch, fixed

src/slurm_plugin/cluster_event_publisher.py
chenwany: Good idea, changed

src/slurm_plugin/cluster_event_publisher.py
chenwany: sure, changed

src/slurm_plugin/jobinfomgtd.py
chenwany: No, it was used for testing before I get the event publisher, now I can remove it

src/slurm_plugin/logging/parallelcluster_clustermgtd_logging.conf
chenwany: fixed the typo

src/slurm_plugin/jobinfomgtd.py
chenwany: no, we catch the exception in the main function, should that already handled this case?

davprat: I thought `main` just re-raises the error?

davprat (approved):
It would be good to get the test coverage up if we had time.

src/slurm_plugin/jobinfomgtd.py
chenwany: In the `main` function, the exception will be caught by the `except` block. The error message will be log out, and the exception will be re-raised. With the `@retry` decorator, after `LOOP_TIME` seconds, the `main` function will be invoked again, it doesn't terminate the entire process. But I agree it is good to add exception handling for process_job_info

davprat (commented):
👍

PR #537, opened 2023-06-16
[3.6] Avoid duplication of nodes in `ClusterManager._find_active_nodes()`

jdeamicis:
### Description of changes
* Avoid duplication of nodes in `ClusterManager._find_active_nodes()` when nodes belong to multiple Slurm partitions.
* This is relevant only in cases when the Slurm partition configuration is edited and PC-managed nodes are added to multiple custom partitions.
* Small extention to the `SlurmNode` class constructor to allow specifying the `slurm_nodes` attribute at object creation.

### Tests
* Added unit test to verify that overlapping partitions do not lead to duplication of nodes in the group returned by
`ClusterManager._find_active_nodes()`.
* Adapted existing tests to new logic.

### References
* Related to https://github.com/aws/aws-parallelcluster/issues/5404
* See also https://github.com/aws/aws-parallelcluster-node/pull/538

### Checklist
- [x] Make sure you are pointing to **the right branch** and add a label in the PR title (i.e. **2.x** vs **3.x**)
- [x] Check all commits' messages are clear, describing what and why vs how.
- [x] Make sure **to have added unit tests or integration tests** to cover the new/modified code.
- [ ] ~Check if documentation is impacted by this change.~

Please review the [guidelines for contributing](../CONTRIBUTING.md) and [Pull Request Instructions](https://github.com/aws/aws-parallelcluster/wiki/Git-Pull-Request-Instructions).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #538, opened 2023-06-16
[develop] Avoid duplication of nodes in `ClusterManager._find_active_nodes()`

jdeamicis:
### Description of changes
* Avoid duplication of nodes in `ClusterManager._find_active_nodes()` when nodes belong to multiple Slurm partitions.
* This is relevant only in cases when the Slurm partition configuration is edited and PC-managed nodes are added to multiple custom partitions.
* Small extention to the `SlurmNode` class constructor to allow specifying the `slurm_nodes` attribute at object creation.

### Tests
* Added unit test to verify that overlapping partitions do not lead to duplication of nodes in the group returned by
`ClusterManager._find_active_nodes()`.
* Adapted existing tests to new logic.

### References
* Related to https://github.com/aws/aws-parallelcluster/issues/5404
* See also https://github.com/aws/aws-parallelcluster-node/pull/537

### Checklist
- [ ] ~Check if documentation is impacted by this change.~

### DIFF
--- a/src/slurm_plugin/clustermgtd.py
+++ b/src/slurm_plugin/clustermgtd.py
@@ -1123,12 +1123,11 @@ class ClusterManager:
 
     @staticmethod
     def _find_active_nodes(partitions_name_map):
-        active_nodes = set()
+        active_nodes = []
         for partition in partitions_name_map.values():
             if partition.state != "INACTIVE":
-                active_nodes |= set(partition.slurm_nodes)
-        # TODO: Returning the set breaks some unit tests (I believe the unit test should be revised though).
-        return sorted(list(active_nodes), key=str)
+                active_nodes += partition.slurm_nodes
+        return list(dict.fromkeys(active_nodes))
 
     def _is_node_in_replacement_valid(self, node, check_node_is_valid):
         """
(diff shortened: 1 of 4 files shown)


davprat (dismissed):
lgtm.

src/slurm_plugin/clustermgtd.py
```diff
@@ -1123,11 +1123,12 @@ def _find_bootstrap_failure_nodes(slurm_nodes):
...
     def _find_active_nodes(partitions_name_map):
-        active_nodes = []
+        active_nodes = set()
         for partition in partitions_name_map.values():
             if partition.state != "INACTIVE":
-                active_nodes += partition.slurm_nodes
-        return active_nodes
+                active_nodes |= set(partition.slurm_nodes)
+        # TODO: Returning the set breaks some unit tests (I believe the unit test should be revised though).
+        return sorted(list(active_nodes), key=str)
```
lukeseawalker: what about leaving `active_nodes = []` and at the end returning
```
return list(dict.fromkeys(active_nodes))
```
this should preserve the original order, and there is no need to change the test

jdeamicis: Implemented. Just one doubt: does `list(dict)` always return the list of keys? Is it guaranteed in Python? Should we use `list(dict.fromkeys(active_nodes).keys())` to make it explicit?

lukeseawalker: it's in official doc https://docs.python.org/3/tutorial/datastructures.html#dictionaries
> Performing list(d) on a dictionary returns a list of all the keys used in the dictionary, in insertion order

lukeseawalker (approved):

PR #546, opened 2023-07-17
[develop] Keep fast capacity failover consistent with current behavior

lukeseawalker:
### Description of changes
* Keep fast capacity failover consistent with current behavior, which is activating fast failover at the first batch launch failure, and not wait to consider failure on the whole aggregated capacity

* Switch preliminary scale-all nodes attempt to be best-effort 
  
  Optimize job level scaling with preliminary scale-all nodes attempt.
  Switch this preliminary scaling from all-or-nothing to best-effort, so to maximize instance launches

### Tests
* unit tests added
* manually tested on running cluster

### References
n/a

### Checklist

### DIFF
--- a/src/slurm_plugin/instance_manager.py
+++ b/src/slurm_plugin/instance_manager.py
@@ -776,7 +776,7 @@ class JobLevelScalingInstanceManager(InstanceManager):
         self.unused_launched_instances |= self._launch_instances(
             nodes_to_launch=self._parse_requested_nodes(node_list),
             launch_batch_size=launch_batch_size,
-            all_or_nothing_batch=True,
+            all_or_nothing_batch=False,
         )
 
         self._scaling_for_jobs(
--- a/src/slurm_plugin/instance_manager.py
+++ b/src/slurm_plugin/instance_manager.py
@@ -466,9 +466,15 @@ class InstanceManager(ABC):
             logger.error("Failed when terminating compute fleet with error %s", e)
             return False
 
-    def _update_failed_nodes(self, nodeset, error_code="Exception"):
+    def _update_failed_nodes(self, nodeset, error_code="Exception", override=True):
         """Update failed nodes dict with error code as key and nodeset value."""
-        self.failed_nodes[error_code] = self.failed_nodes.get(error_code, set()).union(nodeset)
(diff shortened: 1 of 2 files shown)


src/slurm_plugin/instance_manager.py
```diff
@@ -466,9 +466,15 @@ def terminate_all_compute_nodes(self, terminate_batch_size):
             logger.error("Failed when terminating compute fleet with error %s", e)
             return False
 
-    def _update_failed_nodes(self, nodeset, error_code="Exception"):
+    def _update_failed_nodes(self, nodeset, error_code="Exception", override=True):
         """Update failed nodes dict with error code as key and nodeset value."""
-        self.failed_nodes[error_code] = self.failed_nodes.get(error_code, set()).union(nodeset)
+        if not override:
+            # Remove nodes already present in any failed_nodes key so to not override the error_code if already set
```
EddyMM, line 472: Why do we need to remove the nodes from all `error_codes`?

lukeseawalker: in order to not override the error_code if it is already set for those nodes. It's not removing nodes from error_codes, but removing nodes with already set error_code from the nodeset.

src/slurm_plugin/instance_manager.py
```diff
@@ -1045,6 +1055,11 @@ def _launch_instances(
                                 print_with_count(batch_nodes),
                                 e,
                             )
+                            update_failed_nodes_parameters = {"nodeset": set(batch_nodes)}
+                            if isinstance(e, ClientError):
+                                update_failed_nodes_parameters["error_code"] = e.response.get("Error", {}).get("Code")
```
EddyMM, line 1029: Will there be any downstream impact of allowing any ClientError to be used as an `error_code`?
E.g. Event publishing etc.

lukeseawalker: This is what is done in current behavior, for which we want the new scaling logic to keep consistency, see https://github.com/aws/aws-parallelcluster-node/blob/v3.6.1/src/slurm_plugin/instance_manager.py#L134

EddyMM (approved):

PR #563, opened 2023-09-15
[develop] Remove dependency between JobLevelScaling and NodeListScaling

lukeseawalker:
### Description of changes
* Remove dependency between JobLevelScalingInstanceManager and NodeListScalingInstanceManager, by implementing the _add_instances_for_nodes methods also for the JobLevelScalingInstanceManager class. This allows to:
  * have a node-list scaling fallback when slurm resume file content is not available
  * remove the NodeListScalingInstanceManager implementation eventually in the future

* Build a map of launched instances associated to nodes, so that on the processing of the node_resume list, the nodes for which there is already an associated running instance are discarded

### Tests
* unit tests added
* manual tests on running cluster

### References
n/a

### Checklist

### DIFF
--- a/src/slurm_plugin/instance_manager.py
+++ b/src/slurm_plugin/instance_manager.py
@@ -173,113 +173,6 @@ class InstanceManager(ABC):
     ):
         """Add EC2 instances to Slurm nodes."""
 
-    def _add_instances_for_nodes(
-        self,
-        node_list: List[str],
-        launch_batch_size: int,
-        update_node_address: bool = True,
-        all_or_nothing_batch: bool = False,
-    ):
-        """Launch requested EC2 instances for nodes."""
-        nodes_to_launch = self._parse_nodes_resume_list(node_list)
-        for queue, compute_resources in nodes_to_launch.items():
-            for compute_resource, slurm_node_list in compute_resources.items():
-                logger.info("Launching instances for Slurm nodes %s", print_with_count(slurm_node_list))
-
-                # each compute resource can be configured to use create_fleet or run_instances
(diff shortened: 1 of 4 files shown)


src/slurm_plugin/instance_manager.py
```diff
@@ -357,10 +251,14 @@ def _parse_nodes_resume_list(self, node_list: List[str]) -> defaultdict[str, def
         Sample NodeName: queue1-st-computeres1-2
         """
         nodes_to_launch = defaultdict(lambda: defaultdict(list))
+        logger.debug("Nodes already assigned to running instances: %s", self.nodes_assigned_to_instances)
         for node in node_list:
             try:
                 queue_name, node_type, compute_resource_name = parse_nodename(node)
-                nodes_to_launch[queue_name][compute_resource_name].append(node)
+                if node not in self.nodes_assigned_to_instances.get(queue_name, {}).get(compute_resource_name, []):
```
EddyMM: NIT: This logic is because the job might have been submitted with non-exclusive node allocation? If so, maybe we can add a comment to mention this.

---

Optional:
- Can we perform this filtering in a different function in the node launching workflow? This function is meant to parse the resume list as it is.
- Alternatively we can rename this function to something like `get_nodes_to_launch_from_node_list`

lukeseawalker: Thanks, I have added a comment for now.
There reason why I put the logic in the same method was in order to not redo the for loop once again. I will reconsider the implementation in next changes

EddyMM (approved):

PR #564, opened 2023-09-18
[develop] Job Level Scaling for Node Sharing

lukeseawalker:
### Description of changes
* Reset the failure for nodes that were launched successful, for which it was possible to assign an instance.
This to cover the node-sharing (oversubscribe) case when nodes that failed in a job call,
are actually launched (and assigned to instances) in a next iteration of the job loop.

* Add job-level scaling for the node sharing case.
  Before entering the job loop, perform the same optimizations done for the exclusive job case:
  * scale best-effort for all single node jobs
  * scale all for all multi node jobs

* Avoid to set nodes into DOWN, hence avoid calling Slurm scontrol update, if node list is empty.
  Avoided log line is
  ```
  2023-09-19 10:56:39,439 - [slurm_plugin.resume:_handle_failed_nodes] - INFO - Setting following failed nodes into DOWN state (x0) [] with reason: (Code:LimitedInstanceCapacity)Failure when resuming nodes
  ```

### Tests
* unit tests added
* manual tests performed on running cluster given the following submission command:

```
sbatch --wrap "sleep 10000" -N 4 --constraint="[(c5.4xlarge)*3&(p4d.24xlarge)*1]" -p q4; sbatch --wrap "sleep 10000" -N 4 --constraint="[(c5.4xlarge)*3&(p4d.24xlarge)*1]" -p q4; sbatch --wrap "sleep 10000" -N 3 --constraint="[(c5.4xlarge)*3]" -p q4
```

where there is capacity for c5.4xlarge but not for p4d.24xlarge the two scaling strategies were tested:

all_or_nothing_batch = true
expected nodes running at the end of the resume call: (x3) q4-dy-c4-1-*

resume log:
```
2023-09-19 10:56:32,549 - [slurm_plugin.resume:main] - INFO - ResumeProgram startup.
2023-09-19 10:56:32,550 - [slurm_plugin.resume:_get_config] - INFO - Reading /etc/parallelcluster/slurm_plugin/parallelcluster_slurm_resume.conf
[...]

PR #566, opened 2023-10-04
Support specification of multiple scaling strategies

EddyMM:
### Description of changes
* There is need to support more than 2 scaling strategies (i.e. `all-or-nothing` and `best-effort`).
* These changes refactor the code base to accommodate more than 2 scaling strategies.
* This has been done by using an Enum that defines/enumerates the possible strategies
* The mapping between the Enum values and existing `all_or_nothing_batch` configuration parameter has been done as follows:
    * `all_or_nothing_batch=False` maps to `ScalingStrategy.BEST_EFFORT`
    * `all_or_nothing_batch=True` maps to `ScalingStrategy.ALL_OR_NOTHING`
    * `all_or_nothing_batch` missing/undefined maps to `ScalingStrategy.ALL_OR_NOTHING`
* The existing unit tests have been updated to use this `Enum`.

### Tests
* Updated existing unit tests and added a unit test to check the behaviour of the ScalingStrategy Enum based on various values used to initialise it
* Submitted jobs matching the ones done here [#564](https://github.com/aws/aws-parallelcluster-node/pull/564#issue-1900927691) and confirmed that we get the same output

### References
* N/A

### Checklist

### DIFF
--- a/src/slurm_plugin/instance_manager.py
+++ b/src/slurm_plugin/instance_manager.py
@@ -25,7 +25,7 @@ from botocore.exceptions import ClientError
 from common.ec2_utils import get_private_ip_address_and_dns_name
 from common.schedulers.slurm_commands import get_nodes_info, update_nodes
 from common.utils import grouper, setup_logging_filter
-from slurm_plugin.common import ComputeInstanceDescriptor, log_exception, print_with_count
+from slurm_plugin.common import ComputeInstanceDescriptor, ScalingStrategy, log_exception, print_with_count
 from slurm_plugin.fleet_manager import EC2Instance, FleetManagerFactory
 from slurm_plugin.slurm_resources import (
     EC2_HEALTH_STATUS_UNHEALTHY_STATES,
@@ -165,7 +165,7 @@ class InstanceManager(ABC):
         node_list: List[str],
         launch_batch_size: int,
         update_node_address: bool = True,
-        all_or_nothing_batch: bool = False,
+        scaling_strategy: ScalingStrategy = ScalingStrategy.BEST_EFFORT,
         slurm_resume: Dict[str, any] = None,
         assign_node_batch_size: int = None,
         terminate_batch_size: int = None,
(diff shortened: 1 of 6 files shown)


src/slurm_plugin/instance_manager.py
```diff
@@ -1008,9 +1008,13 @@ def _launch_instances(
...
                     logger.info(
                         "Launching %s instances for nodes %s",
-                        "all-or-nothing" if all_or_nothing_batch else "best-effort",
+                        scaling_strategy,
                         print_with_count(slurm_node_list),
                     )
+                    # At instance launch level, the various scaling strategies can be grouped based on the actual
+                    # launch behaviour i.e. all-or-nothing or best-effort
+                    all_or_nothing_batch = scaling_strategy in [ScalingStrategy.ALL_OR_NOTHING]
+
```
NSsirena, line 1017: I'm not sure I understand this comment.

`all_or_nothing_batch` is actually a boolean that is passed downstream to `FleetManagerFactory` as `all_or_nothing`  ..so maybe it would be better to rename this parameter accordingly since is not clear if/what the `_batch` adds

While, regards doing `scaling_strategy in [ScalingStrategy.ALL_OR_NOTHING]`
instead of `scaling_strategy == ScalingStrategy.ALL_OR_NOTHING:`

as done a few lines above, I assume that this is because we are planning to add another strategy that falls in the "all or nothing" category.

Is this the meaning of the comment above ?

src/slurm_plugin/instance_manager.py
```diff
@@ -1217,17 +1222,21 @@ def add_instances(
...
         launch_batch_size: int,
         update_node_address: bool = True,
-        all_or_nothing_batch: bool = False,
+        scaling_strategy: ScalingStrategy = ScalingStrategy.BEST_EFFORT,
     ):
         """Launch requested EC2 instances for nodes."""
+        # At fleet management level, the scaling strategies can be grouped based on the actual
+        # launch behaviour i.e. all-or-nothing or best-effort
+        all_or_nothing_batch = scaling_strategy in [ScalingStrategy.ALL_OR_NOTHING]
+
```
NSsirena, line 1239: See comment above.

src/slurm_plugin/instance_manager.py
EddyMM: In `all_or_nothing_batch` the `_batch` alludes to the way we chunk the launch attempts in "batches" of a certain size/count of instances. So `all_or_nothing_batch` implies that each batch will be done in an all-or-nothing fashion.

> as done a few lines above, I assume that this is because we are planning to add another strategy that falls in the "all or nothing" category.

Exactly, it was done with this in mind. That expression will be reused in the upcoming PR.

NSsirena (approved):

PR #568, opened 2023-10-06
Skip job-level launch when using BEST_EFFORT scaling strategy.

EddyMM:
### Description of changes
* When performing scaling for jobs multi node, we do a scale "all-in" as an optimisation (i.e. an "all-in" scale up is a launch attempt that includes all the capacity needed by the jobs in the resume call).
* We then iterate over the jobs and assign instances to each node based on the launched capacity. We do an additional job-level launch in case the instances needed for a job could not be obtained during the all-in scale up e.g. due to ICE.
* These code changes prevent a job-level launch attempt when using `BEST_EFFORT` scaling strategy. This means for `BEST_EFFORT` scaling, the all-in launch is the only launch attempt to get the capacity needed for jobs.
* This has been done by introducing a `skip_launch` flag which allows callers of the `_launch_instances` function to specify if an actual launch attempt should be made or if it should proceed to reuse already launched instances (i.e. from the `all-in` scale up)

### Tests
* Updated existing unit tests to confirm the skip_launch is applied as expected when using `BEST_EFFORT` scaling
* Ran manual tests involving submission of  jobs that have an ICE for some instances and confirmed that we don't perform a job-level launch attempt when using `BEST_EFFORT`

### References
* N/A

### Checklist

### DIFF
--- a/src/slurm_plugin/instance_manager.py
+++ b/src/slurm_plugin/instance_manager.py
@@ -575,6 +575,7 @@ class JobLevelScalingInstanceManager(InstanceManager):
         assign_node_batch_size: int,
         update_node_address: bool,
         scaling_strategy: ScalingStrategy,
+        skip_launch: bool = False,
     ) -> None:
         """Scaling for job list."""
         # Setup custom logging filter
@@ -592,6 +593,7 @@ class JobLevelScalingInstanceManager(InstanceManager):
                     assign_node_batch_size=assign_node_batch_size,
                     update_node_address=update_node_address,
                     scaling_strategy=scaling_strategy,
+                    skip_launch=skip_launch,
                 )
 
     def _terminate_unassigned_launched_instances(self, terminate_batch_size):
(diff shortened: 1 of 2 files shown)


src/slurm_plugin/instance_manager.py
```diff
@@ -994,8 +1005,10 @@ def _launch_instances(
         launch_batch_size: int,
         scaling_strategy: ScalingStrategy,
         job: SlurmResumeJob = None,
+        skip_launch: bool = False,
```
NSsirena, line 1008: It seems that if `skip_launch == True` we do nothing in this method.

If so we could avoid calling it from `_add_instances_for_nodes`

src/slurm_plugin/instance_manager.py
```diff
@@ -695,12 +697,18 @@ def _scaling_for_jobs_multi_node(
             ),
         )
 
+        # Avoid a job level launch if scaling strategy is BEST_EFFORT
+        # The scale all-in launch has been performed already hence from this point we want to skip the extra
+        # job level launch of instances for jobs that are unable to get the needed capacity from the all-in scaling
+        skip_launch = scaling_strategy == ScalingStrategy.BEST_EFFORT
+
```
NSsirena, line 704: Unless there are other side effects that I'm missing, the following call seems in charge of making an extra attempt to provision new instances for a job if we weren't able to retrieve them with the first `all-in` launch and IF the strategy IS NOT `best-effort`.

If this is the case, can we simply avoid this call instead of creating a new boolean flag and passing it through the whole call-chain ?

something like 

```
if scaling_strategy != ScalingStrategy.BEST_EFFORT:
  self. _scaling_for_jobs(
  ...
 # else skip call because ...
```

src/slurm_plugin/instance_manager.py
EddyMM: This method at the moment is tightly coupled with the logic used to reuse already launched instances (`_resize_slurm_node_list`).

Admittedly this should be factored out of the function but such a change is relatively invasive.

src/slurm_plugin/instance_manager.py
EddyMM: There is a side-effect: https://github.com/aws/aws-parallelcluster-node/pull/568#discussion_r1348593364

NSsirena (approved):

PR #576, opened 2023-10-13
Permit test_slurm_commands to pass on MacOS

enrico-usai:
### Description of changes
Test was failing on MacOS because by default it uses BSD grep, that does not accept `-P` option.

With this change, after installing GNU grep the test will pass. GNU grep can be easily installed from Homebrew: `brew install grep`.

```
$ grep --version
grep (BSD grep, GNU compatible) 2.6.0-FreeBSD

$ brew install grep
$ ggrep --version
ggrep (GNU grep) 3.11
...
```

### Tests
* Executed `test_slurm_commands` on MacOS
* Executed all unit tests for the node package and there are no failures.


gmarciani (approved):

jdeamicis (approved):
I will regret this :D

PR #602, opened 2023-11-13
[develop] Add mechanism to permit to mock generic methods/functions of the python package

enrico-usai:
### Description of changes

The new `ApiMocker.mockable` is a decorator that can be applied to mock any method.

The decorator will automatically search in the `<function-name-path>.overrides.py` file (if it exists) a function with the same name of the function.
If the function exists, the decorator will execute it in place of the original one.

The same `slurm_plugin.overrides.py` file is already used by `run_instances` and `create_fleet` methods,
but without the generic decorator mechanism. We can use this new mechanism for them in the future.

I'm using it for `describe_capacity_reservations` to permit to mock it when running e2e tests.
An example of bash script to create overrides file is:
```
node_virtualenv_path=$(sudo find / -iname "site-packages" | grep "node_virtualenv")
# the overrides.py file must be in the same folder of the module of the function to be mocked
cat << EOF | sudo tee -a "${node_virtualenv_path}/aws/overrides.py"
from aws.ec2 import CapacityReservationInfo

def describe_capacity_reservations(_, capacity_reservations_ids):
    return [
        CapacityReservationInfo({
            "CapacityReservationId": "cr-123456",
            "OwnerId": "123456789",
            "CapacityReservationArn": "arn:aws:ec2:us-east-2:123456789:capacity-reservation/cr-123456",
            ...
        })
    ]
EOF
```

### Tests

Tested with:
* not existing overrides.py (`ImportError`)
* empty overrides.py (`AttributeError`)
* overrides.py with other functions defined on it (`AttributeError`)
* overrides.py with the mocked `describe_capacity_reservations` (mocked output)

### DIFF
--- a/src/common/utils.py
+++ b/src/common/utils.py
@@ -375,3 +375,39 @@ def setup_logging_filter(logger: logging.Logger, custom_field: str):
     finally:
         # Remove the custom log filter
         logger.removeFilter(custom_filter)
+
+
+class ApiMocker:
+    """API mocker."""
+
+    @staticmethod
+    def mockable(func):
+        """
+        Try to mock passed function by searching for an overrides.py file in the same path of the given func.
+
+        This function can be used a decorator and applied any method.
+
+        The function will check if a function called with the name of the given function exists
+        in the <function-dir>/overrides.py, and if it does, the function will execute it.
+
+        E.g. if the method with ApiMocker.mockable decorator is defined in Ec2Client class
+        of the ${node_virtualenv_path}/aws/ec2.py module, the mocked function should be defined
+        in the ${node_virtualenv_path}/aws/overrides.py file.
+        """
+
+        def wrapper(*args, **kwargs):
+            try:
+                function_name = func.__name__
+                # retrieve parent module of the given function that has the ApiMocker.mockable decorator
+                func_module = func.__module__
+                func_parent_module = func_module[: func_module.rindex(".")]
+                # try to import overrides.py module in the same folder of the module to mock
+                overrides_module = __import__(f"{func_parent_module}.overrides", fromlist=function_name)
+                overrided_func = getattr(overrides_module, function_name)
+                log.info("Calling %s override with args: %s and kwargs: %s", function_name, args, kwargs)
+                result = overrided_func(*args, **kwargs)
+            except (ImportError, AttributeError):
+                result = func(*args, **kwargs)
+            return result
+
+        return wrapper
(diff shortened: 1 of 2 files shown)


src/common/test_utils.py
```diff
@@ -0,0 +1,44 @@
+# Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
lukeseawalker: What about putting this in the `utils.py` already present? Reason is that it can be used even outside a "test" context

src/common/test_utils.py
```diff
@@ -0,0 +1,44 @@
...
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+import logging
+
+# Path of the module used to permit to override python modules/functions from the node package.
+# node_virtual_env path can be retrieved with: `sudo find / -iname slurm_plugin -print0 | grep -FzZ node_virtualenv`
+OVERRIDES_PYTHON_MODULE = "slurm_plugin.overrides"
```
lukeseawalker: can be this generalized in a way that it searches for the override.py file in the same directory where the mocked function is defined?

e.g. the method to be mocked is in a/b/class.py file, and to override it, I can create a file at a/b/override.py

src/common/test_utils.py
enrico-usai: done.

lukeseawalker (dismissed):

src/common/test_utils.py
enrico-usai: done, nice idea!

src/common/utils.py
```diff
@@ -375,3 +375,39 @@ def filter(self, record: logging.LogRecord) -> bool:
...
+    def mockable(func):
+        """
+        Try to mock passed function by searching for an overrides.py file in the same path of the given func.
+
+        This function can be used a decorator and applied any method.
+
+        The function will check if a function called with the name of the given function exists
+        in the <function-dir>/overrides.py, and if it does, the function will execute it.
+
+        E.g. if the method with ApiMOcker.mockable decorator is defined in Ec2Client class
```
lukeseawalker: ApiMOcker -> ApiMocker

lukeseawalker (approved):

src/common/utils.py
enrico-usai: fixed

PR #613, opened 2024-02-13
Add a backing instance max count config option for clustermgtd to be robust to eventual EC2 consistency

dreambeyondorange:
### Description of changes
* Add a backing_instance_max_count config option for clustermgtd to be robust to eventual EC2 consistency

* Adding a config option to clustermgtd, ec2_backing_instance_max_count, to allow more time for describe-instances to reach eventual consistency with run-instances data
* Passes the max count and map to is_healthy() for static and dynamic nodes to evaluate the count for individual instances.

### Tests
* Existing unit tests are passing
* Added unit tests for additional logic

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ This file is used to list changes made in each version of the aws-parallelcluste
 ------
 
 **ENHANCEMENTS**
+- Add a clustermgtd config option `ec2_instance_missing_max_count` to allow a configurable amount of retries for eventual EC2 
+  describe instances consistency with run instances
 
 **CHANGES**
 
--- a/src/slurm_plugin/clustermgtd.py
+++ b/src/slurm_plugin/clustermgtd.py
@@ -150,6 +150,7 @@ class ClustermgtdConfig:
         "terminate_drain_nodes": True,
         "terminate_down_nodes": True,
         "orphaned_instance_timeout": 300,
+        "ec2_instance_missing_max_count": 2,
         # Health check configs
         "disable_ec2_health_check": False,
         "disable_scheduled_event_health_check": False,
@@ -304,6 +305,11 @@ class ClustermgtdConfig:
         self.insufficient_capacity_timeout = config.getfloat(
             "clustermgtd", "insufficient_capacity_timeout", fallback=self.DEFAULTS.get("insufficient_capacity_timeout")
         )
+        self.ec2_instance_missing_max_count = config.getint(
+            "clustermgtd",
+            "ec2_instance_missing_max_count",
+            fallback=self.DEFAULTS.get("ec2_instance_missing_max_count"),
+        )
         self.disable_nodes_on_insufficient_capacity = self.insufficient_capacity_timeout > 0
 
     def _get_dns_config(self, config):
@@ -384,6 +390,7 @@ class ClusterManager:
         self._insufficient_capacity_compute_resources = {}
         self._static_nodes_in_replacement = set()
         self._partitions_protected_failure_count_map = {}
+        self._nodes_without_backing_instance_count_map = {}
         self._compute_fleet_status = ComputeFleetStatus.RUNNING
         self._current_time = None
         self._config = None
@@ -492,7 +499,10 @@ class ClusterManager:
         partitions_protected_failure_count_map = self._partitions_protected_failure_count_map.copy()
         for partition, failures_per_compute_resource in partitions_protected_failure_count_map.items():
             partition_online_compute_resources = partitions_name_map[partition].get_online_node_by_type(
-                self._config.terminate_drain_nodes, self._config.terminate_down_nodes
+                self._config.terminate_drain_nodes,
+                self._config.terminate_down_nodes,
+                self._config.ec2_instance_missing_max_count,
+                self._nodes_without_backing_instance_count_map,
             )
             for compute_resource in failures_per_compute_resource.keys():
                 if compute_resource in partition_online_compute_resources:
@@ -762,6 +772,8 @@ class ClusterManager:
             if not node.is_healthy(
                 consider_drain_as_unhealthy=self._config.terminate_drain_nodes,
                 consider_down_as_unhealthy=self._config.terminate_down_nodes,
+                ec2_instance_missing_max_count=self._config.ec2_instance_missing_max_count,
+                nodes_without_backing_instance_count_map=self._nodes_without_backing_instance_count_map,
                 log_warn_if_unhealthy=node.name not in reserved_nodenames,
             ):
                 if not self._config.disable_capacity_blocks_management and node.name in reserved_nodenames:
@@ -778,7 +790,11 @@ class ClusterManager:
                     ).append(node)
                 else:
                     unhealthy_dynamic_nodes.append(node)
-        self._event_publisher.publish_unhealthy_node_events(all_unhealthy_nodes)
+        self._event_publisher.publish_unhealthy_node_events(
+            all_unhealthy_nodes,
+            self._config.ec2_instance_missing_max_count,
+            self._nodes_without_backing_instance_count_map,
+        )
         return (
             unhealthy_dynamic_nodes,
             unhealthy_static_nodes,
@@ -1167,11 +1183,12 @@ class ClusterManager:
         """Check if a static node is in replacement but replacement time is expired."""
         return self._is_node_in_replacement_valid(node, check_node_is_valid=False)
 
-    @staticmethod
-    def _find_bootstrap_failure_nodes(slurm_nodes):
+    def _find_bootstrap_failure_nodes(self, slurm_nodes):
         bootstrap_failure_nodes = []
         for node in slurm_nodes:
-            if node.is_bootstrap_failure():
+            if node.is_bootstrap_failure(
+                self._config.ec2_instance_missing_max_count, self._nodes_without_backing_instance_count_map
+            ):
                 bootstrap_failure_nodes.append(node)
         return bootstrap_failure_nodes
 
--- a/src/slurm_plugin/slurm_resources.py
+++ b/src/slurm_plugin/slurm_resources.py
@@ -63,12 +63,24 @@ class SlurmPartition:
     def has_running_job(self):
         return any(node.is_running_job() for node in self.slurm_nodes)
 
-    def get_online_node_by_type(self, terminate_drain_nodes, terminate_down_nodes):
+    def get_online_node_by_type(
+        self,
+        terminate_drain_nodes,
+        terminate_down_nodes,
+        ec2_instance_missing_max_count,
+        nodes_without_backing_instance_count_map,
+    ):
         online_compute_resources = set()
         if not self.state == "INACTIVE":
             for node in self.slurm_nodes:
                 if (
-                    node.is_healthy(terminate_drain_nodes, terminate_down_nodes, log_warn_if_unhealthy=False)
+                    node.is_healthy(
+                        terminate_drain_nodes,
+                        terminate_down_nodes,
+                        ec2_instance_missing_max_count,
+                        nodes_without_backing_instance_count_map,
+                        log_warn_if_unhealthy=False,
+                    )
                     and node.is_online()
                 ):
                     logger.debug("Currently online node: %s, node state: %s", node.name, node.state_string)
@@ -233,6 +245,7 @@ class SlurmNode(metaclass=ABCMeta):
         self.is_failing_health_check = False
         self.error_code = self._parse_error_code()
         self.queue_name, self._node_type, self.compute_resource_name = parse_nodename(name)
+        self.ec2_backing_instance_valid = None
 
     def is_nodeaddr_set(self):
         """Check if nodeaddr(private ip) for the node is set."""
@@ -378,7 +391,7 @@ class SlurmNode(metaclass=ABCMeta):
         pass
 
     @abstractmethod
-    def is_bootstrap_failure(self):
+    def is_bootstrap_failure(self, ec2_instance_missing_max_count, nodes_without_backing_instance_count_map: dict):
         """
         Check if a slurm node has boostrap failure.
 
@@ -394,7 +407,14 @@ class SlurmNode(metaclass=ABCMeta):
         pass
 
     @abstractmethod
-    def is_healthy(self, consider_drain_as_unhealthy, consider_down_as_unhealthy, log_warn_if_unhealthy=True):
+    def is_healthy(
+        self,
+        consider_drain_as_unhealthy,
+        consider_down_as_unhealthy,
+        ec2_instance_missing_max_count,
+        nodes_without_backing_instance_count_map: dict,
+        log_warn_if_unhealthy=True,
+    ):
         """Check if a slurm node is considered healthy."""
         pass
 
@@ -404,8 +424,18 @@ class SlurmNode(metaclass=ABCMeta):
         # for example because of a short SuspendTimeout
         return self.is_nodeaddr_set() and (self.is_power() or self.is_powering_down())
 
-    def is_backing_instance_valid(self, log_warn_if_unhealthy=True):
+    def is_backing_instance_valid(
+        self,
+        ec2_instance_missing_max_count,
+        nodes_without_backing_instance_count_map: dict,
+        log_warn_if_unhealthy=True,
+    ):
         """Check if a slurm node's addr is set, it points to a valid instance in EC2."""
+        # Perform this logic only once and return the result thereafter
+        if self.ec2_backing_instance_valid is not None:
+            return self.ec2_backing_instance_valid
+        # Set ec2_backing_instance_valid to True since it will be the result most often
+        self.ec2_backing_instance_valid = True
         if self.is_nodeaddr_set():
             if not self.instance:
                 if log_warn_if_unhealthy:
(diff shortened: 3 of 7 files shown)


src/slurm_plugin/clustermgtd.py
```diff
@@ -763,6 +767,7 @@ def _find_unhealthy_slurm_nodes(self, slurm_nodes):
                 consider_drain_as_unhealthy=self._config.terminate_drain_nodes,
                 consider_down_as_unhealthy=self._config.terminate_down_nodes,
                 log_warn_if_unhealthy=node.name not in reserved_nodenames,
+                ec2_backing_instance_timeout=self._config.ec2_backing_instance_timeout,
```
lukeseawalker: there is no need to pass this all the way through the methods call up to where it's used.
In the method you need this value, you can access it through `self._config.ec2_backing_instance_timeout`

src/slurm_plugin/slurm_resources.py
```diff
@@ -394,7 +394,13 @@ def is_bootstrap_timeout(self):
         pass
 
     @abstractmethod
-    def is_healthy(self, consider_drain_as_unhealthy, consider_down_as_unhealthy, log_warn_if_unhealthy=True):
+    def is_healthy(
+        self,
+        consider_drain_as_unhealthy,
+        consider_down_as_unhealthy,
+        log_warn_if_unhealthy=True,
+        ec2_backing_instance_timeout=300,
```
lukeseawalker: no need, see https://github.com/aws/aws-parallelcluster-node/pull/613/files#r1490652983

src/slurm_plugin/slurm_resources.py
```diff
@@ -404,7 +410,7 @@ def is_powering_down_with_nodeaddr(self):
         # for example because of a short SuspendTimeout
         return self.is_nodeaddr_set() and (self.is_power() or self.is_powering_down())
 
-    def is_backing_instance_valid(self, log_warn_if_unhealthy=True):
+    def is_backing_instance_valid(self, log_warn_if_unhealthy=True, ec2_backing_instance_timeout=None):
```
lukeseawalker: no need, see https://github.com/aws/aws-parallelcluster-node/pull/613/files#r1490652983

src/slurm_plugin/slurm_resources.py
```diff
@@ -414,7 +420,12 @@ def is_backing_instance_valid(self, log_warn_if_unhealthy=True):
                         self,
                         self.state_string,
                     )
-                return False
+                # Allow a few minutes for the eventual consistency of EC2 data
+                # This can only be done if the start time and timeout exist
+                if self.slurmdstarttime is None or ec2_backing_instance_timeout is None:
+                    return False
+                if (datetime.now() - self.slurmdstarttime).seconds > ec2_backing_instance_timeout:
+                    return False
```
lukeseawalker: * I think datetime.now() is not enough, because you should consider the timezone, i.e. datetime.now(tz=timezone.utc)
* instead of doing this double if, you should leverage the `time_is_up` util function we have in the common package
* you should add logging messaging to understand when what's happening here, e.g. instance is not present but  it's within the timeout Vs instance is not present and timeout expired
* unit tests are missing for this new logic
* are you sure that slurmdstarttime is already set when the node is configured but the slurmd hasn't actually started yet (because instance didn't bootstrapped yet)? I fear this is going to always return False

src/slurm_plugin/slurm_resources.py
```diff
@@ -478,11 +489,19 @@ def __init__(
             reservation_name=reservation_name,
         )
 
-    def is_healthy(self, consider_drain_as_unhealthy, consider_down_as_unhealthy, log_warn_if_unhealthy=True):
+    def is_healthy(
+        self,
+        consider_drain_as_unhealthy,
+        consider_down_as_unhealthy,
+        log_warn_if_unhealthy=True,
+        ec2_backing_instance_timeout=300,
```
lukeseawalker: no need, see https://github.com/aws/aws-parallelcluster-node/pull/613/files#r1490652983

src/slurm_plugin/slurm_resources.py
```diff
@@ -478,11 +489,19 @@ def __init__(
...
+    ):
         """Check if a slurm node is considered healthy."""
         return (
             self._is_static_node_ip_configuration_valid(log_warn_if_unhealthy=log_warn_if_unhealthy)
-            and self.is_backing_instance_valid(log_warn_if_unhealthy=log_warn_if_unhealthy)
+            and self.is_backing_instance_valid(
+                log_warn_if_unhealthy=log_warn_if_unhealthy, ec2_backing_instance_timeout=ec2_backing_instance_timeout
```
lukeseawalker: no need, see https://github.com/aws/aws-parallelcluster-node/pull/613/files#r1490652983

src/slurm_plugin/slurm_resources.py
```diff
@@ -618,9 +637,17 @@ def is_state_healthy(self, consider_drain_as_unhealthy, consider_down_as_unhealt
             return False
         return True
 
-    def is_healthy(self, consider_drain_as_unhealthy, consider_down_as_unhealthy, log_warn_if_unhealthy=True):
+    def is_healthy(
+        self,
+        consider_drain_as_unhealthy,
+        consider_down_as_unhealthy,
+        log_warn_if_unhealthy=True,
+        ec2_backing_instance_timeout=300,
```
lukeseawalker: no need, see https://github.com/aws/aws-parallelcluster-node/pull/613/files#r1490652983

src/slurm_plugin/slurm_resources.py
```diff
@@ -618,9 +637,17 @@ def is_state_healthy(self, consider_drain_as_unhealthy, consider_down_as_unhealt
...
+    ):
         """Check if a slurm node is considered healthy."""
-        return self.is_backing_instance_valid(log_warn_if_unhealthy=log_warn_if_unhealthy) and self.is_state_healthy(
+        return self.is_backing_instance_valid(
+            log_warn_if_unhealthy=log_warn_if_unhealthy, ec2_backing_instance_timeout=ec2_backing_instance_timeout
```
lukeseawalker: no need, see https://github.com/aws/aws-parallelcluster-node/pull/613/files#r1490652983

src/slurm_plugin/clustermgtd.py
```diff
@@ -150,6 +150,7 @@ class ClustermgtdConfig:
         "terminate_drain_nodes": True,
         "terminate_down_nodes": True,
         "orphaned_instance_timeout": 300,
+        "ec2_backing_instance_timeout": 300,
```
lukeseawalker: I think the 5 minutes default is too much. Let's sync on this

src/slurm_plugin/clustermgtd.py
dreambeyondorange: This node object doesn't have the value that is stored in cluster manager.  Is there a good way to add this variable to the SlurmNode object without resetting every time this loop runs?  I am sort of treating the slurm nodes as stateless because the state is reset every time through the loop

src/slurm_plugin/clustermgtd.py
dreambeyondorange: I'm ok with shortening to say 2 mins to allow 2-3 loops

src/slurm_plugin/slurm_resources.py
dreambeyondorange: Added unit tests and switched to using the count interval since the start time is unknown

dreambeyondorange: Added additional logging as well

src/slurm_plugin/clustermgtd.py
```diff
@@ -304,6 +305,11 @@ def _get_terminate_config(self, config):
         self.insufficient_capacity_timeout = config.getfloat(
             "clustermgtd", "insufficient_capacity_timeout", fallback=self.DEFAULTS.get("insufficient_capacity_timeout")
         )
+        self.ec2_backing_instance_max_count = config.getfloat(
```
lukeseawalker: does it need to be a float?

src/slurm_plugin/slurm_resources.py
```diff
@@ -414,7 +427,31 @@ def is_backing_instance_valid(self, log_warn_if_unhealthy=True):
...
+                logger.info(
+                    f"ec2_backing_instance_max_count {ec2_backing_instance_max_count} "
+                    f"nodes_without_backing_instance_count_map {nodes_without_backing_instance_count_map}"
+                )
+                if any(
+                    args in [None]
+                    for args in [ec2_backing_instance_max_count, nodes_without_backing_instance_count_map]
+                ):
+                    logger.info("No max count or map provided, ignoring backing instance timeout")
+                    return False
```
lukeseawalker: I think this is useless since the ec2_backing_instance_max_count has a default value. Even if it's not present in the config, its value is always set

src/slurm_plugin/slurm_resources.py
```diff
@@ -414,7 +427,31 @@ def is_backing_instance_valid(self, log_warn_if_unhealthy=True):
                         self,
                         self.state_string,
                     )
-                return False
+                # Allow a few iterations for the eventual consistency of EC2 data
+                logger.info(
+                    f"ec2_backing_instance_max_count {ec2_backing_instance_max_count} "
+                    f"nodes_without_backing_instance_count_map {nodes_without_backing_instance_count_map}"
```
lukeseawalker: do we want to log this at info level? 
* the ec2_backing_instance_max_count is already logged as clustermgtd config value
* the nodes_without_backing_instance_count_map probably make sense at debug value, but a wording about what it's being logged should be added

src/slurm_plugin/slurm_resources.py
```diff
@@ -414,7 +427,31 @@ def is_backing_instance_valid(self, log_warn_if_unhealthy=True):
...
+                    args in [None]
+                    for args in [ec2_backing_instance_max_count, nodes_without_backing_instance_count_map]
+                ):
+                    logger.info("No max count or map provided, ignoring backing instance timeout")
+                    return False
+                elif nodes_without_backing_instance_count_map.get(self.name, 0) >= ec2_backing_instance_max_count:
+                    logger.warning(f"Instance {self.name} availability has timed out.")
+                    return False
+                else:
+                    if not self.missing_count_incremented:
```
lukeseawalker: instead of introducing another variable, cannot we cache the value of the returned value of the function? I think it would be easier to understand the flow

src/slurm_plugin/slurm_resources.py
```diff
@@ -414,7 +427,31 @@ def is_backing_instance_valid(self, log_warn_if_unhealthy=True):
...
+                        nodes_without_backing_instance_count_map[self.name] = (
+                            nodes_without_backing_instance_count_map.get(self.name, 0) + 1
+                        )
+                        logger.warning(
+                            f"Instance {self.name} is not yet available in EC2, "
```
lukeseawalker: This is similar to "Node state check: no corresponding instance in EC2 for node %s, node state: %s" at line 426.
Is it possible to keep just one? I would be inclined to keep what's already present, at most by adding other details, rather than changing it completely.

src/slurm_plugin/clustermgtd.py
```diff
@@ -150,6 +150,7 @@ class ClustermgtdConfig:
         "terminate_drain_nodes": True,
         "terminate_down_nodes": True,
         "orphaned_instance_timeout": 300,
+        "ec2_backing_instance_max_count": 2,
```
lukeseawalker: I feel the naming of this var doesn't communicate it's purpose.
If I don't know the behind the scene on how this value it's used, the variable name should tell me that this is the limit we are allowing an instance to be missing from EC2.

src/slurm_plugin/slurm_resources.py
```diff
@@ -414,7 +427,31 @@ def is_backing_instance_valid(self, log_warn_if_unhealthy=True):
...
+                ):
+                    logger.info("No max count or map provided, ignoring backing instance timeout")
+                    return False
+                elif nodes_without_backing_instance_count_map.get(self.name, 0) >= ec2_backing_instance_max_count:
+                    logger.warning(f"Instance {self.name} availability has timed out.")
+                    return False
+                else:
+                    if not self.missing_count_incremented:
+                        nodes_without_backing_instance_count_map[self.name] = (
+                            nodes_without_backing_instance_count_map.get(self.name, 0) + 1
```
lukeseawalker: [BUG] only incrementing the missing count lead to a bug triggered starting from the second time the clustermgtd checks the healthiness of another instance with the same node name. In this case, the count for that node will not start from zero, and the retry logic will not work anymore. This condition must be covered by the unit test

src/slurm_plugin/clustermgtd.py
dreambeyondorange: I don't think it does, but I was following the precedent above it.  I'll change it!

src/slurm_plugin/slurm_resources.py
dreambeyondorange: The default value is None though.  It may have a default value in the Cluster manager, but not in the SlurmNode, so it is not guaranteed to be None if the usage of the function changes

src/slurm_plugin/slurm_resources.py
dreambeyondorange: I think it makes more sense to track the state if this with the Object rather than the function.  That would also mutate the interface for the is_healthy/is_backing_instance_valid functions to return an object rather than a bool, which I'm not a fan of doing.  That would break the design intent of the is_ function to return a bool, which is what you would expect, not an actual object

src/slurm_plugin/clustermgtd.py
dreambeyondorange: I agree and I was struggling a bit with this name.  I thought it was getting too long lol

something like `ec2_instance_missing_max_count`?

src/slurm_plugin/slurm_resources.py
dreambeyondorange: Oh, I hadn't considered this edge case.  Is this possible in a single loop of clustermgtd?  This is a property of the `SlurmNode` object, not `ClusterManager` and should be reset on each loop.  I only see one call to get_ec2_instances so it doesn't look like this object will ever get updated with a new instance

src/slurm_plugin/slurm_resources.py
dreambeyondorange: I agree that is does end up duplicating the info, but we don't know if the missing count will be incremented before the if block.  I could remove the first part

src/slurm_plugin/slurm_resources.py
```diff
@@ -414,8 +440,33 @@ def is_backing_instance_valid(self, log_warn_if_unhealthy=True):
...
-        return True
+                # Allow a few iterations for the eventual consistency of EC2 data
+                logger.debug(
+                    f"Map of slurm nodes without backing instances {nodes_without_backing_instance_count_map}"
+                )
+                missing_instance_loop_count = nodes_without_backing_instance_count_map.get(self.name, 0)
+                # If the loop count has been reached, the instance is unhealthy and will be terminated
+                if missing_instance_loop_count >= ec2_instance_missing_max_count:
+                    if log_warn_if_unhealthy:
+                        logger.warning(f"Instance {self.name} availability has timed out.")
```
lukeseawalker: what about
"EC2 instance availability for node {self.name} has timed out."
?

src/slurm_plugin/slurm_resources.py
```diff
@@ -414,8 +440,33 @@ def is_backing_instance_valid(self, log_warn_if_unhealthy=True):
...
+                    # Remove the slurm node from the map since a new instance will be launched
+                    nodes_without_backing_instance_count_map.pop(self.name, None)
+                    self.ec2_backing_instance_valid = False
+                else:
+                    nodes_without_backing_instance_count_map[self.name] = (
+                        nodes_without_backing_instance_count_map.get(self.name, 0) + 1
+                    )
+                    if log_warn_if_unhealthy:
+                        logger.warning(
+                            f"incrementing missing count to {nodes_without_backing_instance_count_map[self.name]}."
```
lukeseawalker: what about
"Incrementing missing EC2 instance count for node {self.name} to {nodes_without_backing_instance_count_map[self.name]}."
?

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ This file is used to list changes made in each version of the aws-parallelcluste
 ------
 
 **ENHANCEMENTS**
+- Add a clustermgtd config option `ec2_instance_missing_max_count` to allow a configurable amount of time for eventual EC2
```
lukeseawalker: MINOR - instead of "to allow a configurable amount of time" what about "to allow a configurable amount of retry"?

dreambeyondorange: that's fair, more correct

src/slurm_plugin/slurm_resources.py
```diff
@@ -414,8 +440,34 @@ def is_backing_instance_valid(self, log_warn_if_unhealthy=True):
...
+                # If the loop count has been reached, the instance is unhealthy and will be terminated
+                if missing_instance_loop_count >= ec2_instance_missing_max_count:
+                    if log_warn_if_unhealthy:
+                        logger.warning(f"EC2 instance availability for node {self.name} has timed out.")
+                    # Remove the slurm node from the map since a new instance will be launched
+                    nodes_without_backing_instance_count_map.pop(self.name, None)
+                    self.ec2_backing_instance_valid = False
+                else:
+                    nodes_without_backing_instance_count_map[self.name] = (
+                        nodes_without_backing_instance_count_map.get(self.name, 0) + 1
```
lukeseawalker: you have already stored the value of "nodes_without_backing_instance_count_map.get(self.name)" in missing_instance_loop_count

dreambeyondorange: oh thx i meant to use that here

lukeseawalker (dismissed):

lukeseawalker (approved):

PR #639, opened 2024-07-10
[Bugfix] Fix a bug that was causing `clustermgtd` to fail when a field returned by the command `scontrol show nodes` has a value that contains the character `=`.

gmarciani:
### Description of changes
Fix a bug that was causing `clustermgtd` to fail when a field returned by the command `scontrol show nodes` has a value that contains the character `=`.

In particular, before this change when the field `Reason` is set to a string containing `=`, then clustermgtd would fail with the below error casuing the nodes management loop to be interrupted.

```
Unable to get partition/node info from slurm, no other action can be performed. Sleeping... Exception: too many values to unpack (expected 2)
```

### Tests
* Unit Test

### References

### Checklist

### DIFF
--- a/tests/common/schedulers/test_slurm_commands.py
+++ b/tests/common/schedulers/test_slurm_commands.py
@@ -194,23 +194,23 @@ def test_is_static_node(nodename, expected_is_static):
             False,
         ),
         (
-            "NodeName=multiple-dy-c5xlarge-4\n"
-            "NodeAddr=multiple-dy-c5xlarge-4\n"
-            "NodeHostName=multiple-dy-c5xlarge-4\n"
+            "NodeName=multiple-dy-c5xlarge-3\n"
+            "NodeAddr=multiple-dy-c5xlarge-3\n"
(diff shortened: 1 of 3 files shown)


tests/common/schedulers/test_slurm_commands.py
```diff
@@ -194,23 +194,23 @@ def test_is_static_node(nodename, expected_is_static):
             False,
         ),
         (
-            "NodeName=multiple-dy-c5xlarge-4\n"
-            "NodeAddr=multiple-dy-c5xlarge-4\n"
-            "NodeHostName=multiple-dy-c5xlarge-4\n"
+            "NodeName=multiple-dy-c5xlarge-3\n"
```
hgreebe, line 197: Why did you change the name from `multiple-dy-c5xlarge-4` to `multiple-dy-c5xlarge-3`?

hgreebe (approved):

tests/common/schedulers/test_slurm_commands.py
gmarciani: no reason, just fake unit test data

PR #685, opened 2026-01-26
[Observability] Emit metric `ClustermgtdHeartbeat` to signal clustermgtd heartbeat.

gmarciani:
### Description of changes
Emit metric `ClustermgtdHeartbeat` to signal clustermgtd heartbeat.
The metric is emitted with dimensions: ClusterName and InstanceId.

The metric is intentionally emitted at the end of the clustermgtd loop to represent the real health of the daemon.
If it was emitted at the beginning of the iteration, it would be open to false negatives.

This PR depends on the permissions added in https://github.com/aws/aws-parallelcluster/pull/7209

### Tests
* Unit tests (extended to cover the current changes)
* Manually verified by creating a cluster and checking that (i) the metric is pushed (ii) when the metric push fails (manually removing the permissions to push it), the overall clustermgtd iteration is not compromised.

### References
1. The max length for a metric name is 255 chars. See https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_MetricDatum.html

### DIFF
--- a/src/slurm_plugin/clustermgtd.py
+++ b/src/slurm_plugin/clustermgtd.py
@@ -41,6 +41,7 @@ from common.time_utils import seconds
 from common.utils import check_command_output, read_json, sleep_remaining_loop_time, time_is_up, wait_remaining_time
 from retrying import retry
 from slurm_plugin.capacity_block_manager import CapacityBlockManager
+from slurm_plugin.cloudwatch_utils import CloudWatchMetricsPublisher
 from slurm_plugin.cluster_event_publisher import ClusterEventPublisher
 from slurm_plugin.common import TIMESTAMP_FORMAT, ScalingStrategy, log_exception, print_with_count
 from slurm_plugin.console_logger import ConsoleLogger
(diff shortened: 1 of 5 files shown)


himani2411 (approved):
LGTM!

src/slurm_plugin/clustermgtd.py
```diff
@@ -574,6 +588,9 @@ def manage_cluster(self):
         # Write clustermgtd heartbeat to file
         self._write_timestamp_to_file()
 
+        # Publish heartbeat metric to CloudWatch
+        self._metrics_publisher.put_metric(metric_name=CW_METRICS_HEARTBEAT, value=1)
+
```
hanwen-cluster, line 593: How about surrounding this with a try catch and adding a warning log line? This is not a critical cluster management logic. We should avoid it throwing an Exception out of the function.

gmarciani: I agree this is done inside the metrics publisher itself.

PR #694, opened 2026-03-23
[Node] Replace IP-based matching with instance-ID-based matching to handle EC2 eventual consistency

hehe7318:
### Description of changes
When launching large numbers of instances, DescribeInstances may return instances with missing PrivateIpAddress due to EC2 API eventual consistency. Previously, clustermgtd matched EC2 instances to Slurm nodes by IP address, causing these healthy nodes were treated as missing, replaced, and their instances terminated as orphans, driving the queue into protected mode.

This commit replaces IP-based matching with instance-ID-based matching by leveraging Slurm's native scontrol update InstanceId support (since 23.11).

### Changes
**1. Match by InstanceId instead of private IP (clustermgtd)**
- Store the EC2 instance ID on each Slurm node during launch by setting `InstanceId` in the same batched `scontrol update` as `NodeAddr`, and parse it back from `scontrol show nodes`.
- clustermgtd now associates instances to nodes by `InstanceId` (always present in `DescribeInstances`) rather than `PrivateIpAddress`.
- `get_cluster_instances` no longer discards instances with missing IP info; they are kept with empty IP fields so they can still be matched by instance ID. (The only clustermgtd consumer of `private_ip` is an event log line, which tolerates empty.)
- An unset `InstanceId` reported by Slurm as `(null)` is normalized to `None`.

Batched per-node `InstanceId` assignment requires Slurm >= 25.11.6, which is already in develop.

**2. Configurable CreateFleet DescribeInstances retry timeout**
- For CreateFleet (flexible instance types), `_get_instances_info` retries `DescribeInstances` after launch to tolerate eventual consistency. Replaced the fixed ~11s budget with a configurable total timeout (`instance_info_retrieval_timeout`, default 90s) and a per-attempt backoff cap, per EC2 guidance.

### Tests
* Unit tests passed
* Manually tests ongoing

### Related PRs
- https://github.com/aws/aws-parallelcluster/pull/7425
- https://github.com/aws/aws-parallelcluster-cookbook/pull/3197

### Checklist


src/slurm_plugin/fleet_manager.py
```diff
@@ -24,6 +24,12 @@
 
 logger = logging.getLogger(__name__)
 
+# Total time budget (seconds) and per-attempt backoff cap for retrying DescribeInstances after a CreateFleet
+# launch, to tolerate EC2 API eventual consistency.
+# See https://docs.aws.amazon.com/ec2/latest/devguide/eventual-consistency.html
+INSTANCE_INFO_RETRIEVAL_TIMEOUT_DEFAULT = 90
+INSTANCE_INFO_RETRIEVAL_MAX_BACKOFF = 30
```
hanwen-cluster, line 31: How about putting all defaults in [‎cli/src/pcluster/constants.py](https://github.com/aws/aws-parallelcluster/pull/7425/changes#diff-37df1ebf3d5f2407ceaa922df965c53d0c447ebb7bc4f1145c9c247619f6e3ef) instead of scattering the defaults in CLI, cookbook, node?

src/slurm_plugin/fleet_manager.py
```diff
@@ -436,22 +446,39 @@ def _get_instances_info(self, instance_ids: list):
...
+                    print_with_count(partial_instance_ids),
+                    self._instance_info_retrieval_timeout,
+                    attempt_count + 1,
+                )
+                break
+            elapsed_backoff += base_backoff
             attempt_count += 1
-            if attempt_count < retries:
-                time.sleep(0.3 * 2**attempt_count + (secrets.randbelow(500) / 1000))
+            time.sleep(base_backoff + (secrets.randbelow(500) / 1000))
```
hanwen-cluster, line 481: minor: why is `randbelow` needed?

hehe7318: It was there previously and it's a standard jitter in exponential backoff algorithm. See https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/.

src/slurm_plugin/fleet_manager.py
hehe7318: I don't fully understand, then how to import it in node package?

hanwen-cluster: We can pass the value from CLI, e.g. cli -> chef attribute -> parallelcluster_clustermgtd.conf.erb/parallelcluster_slurm_resume.conf.erb -> node

However, this comment does not block this PR. After offline discussion, we noticed other instances of using defaults directly in node package.

hanwen-cluster (approved):

PR #707, opened 2026-06-10
[Node] Reset inactive partition nodes on fleet start to ensure recovering from protected mode

hehe7318:
### Description of changes
`_start_partitions` now resets the nodes of INACTIVE partitions before bringing partitions back UP, via a new `reset_nodes_in_inactive_partitions` helper in `slurm_commands`. Only INACTIVE partitions are reset, so nodes of active partitions (which may be running jobs) are not touched. The stop path is unchanged.

Why: when protected mode disables a partition, it sets it INACTIVE but does not reset its nodes in the same iteration (cleanup is normally done on a later clustermgtd iteration). Dynamic nodes left powering up without a backing instance can therefore keep a stale nodeaddr. Starting the fleet brought partitions back UP without resetting these nodes, so right after start clustermgtd saw them again as bootstrap failures and sent the cluster straight back into protected mode. As documented, starting the fleet alone should be enough to recover from protected mode without requiring a fleet stop; resetting these nodes on start makes that reliable.

### Tests
* Unit test passed as expected.
* Integ-test `test_slurm_custom_partitions` passed.

### Checklist

### DIFF
--- a/src/common/schedulers/slurm_commands.py
+++ b/src/common/schedulers/slurm_commands.py
@@ -240,6 +240,7 @@ def reset_nodes_in_inactive_partitions():
             if PartitionStatus(part.state) == PartitionStatus.INACTIVE and part.nodenames
         ]
         if not inactive_partition_nodes:
+            log.info("No nodes in INACTIVE partitions, nothing to reset.")
             return
         # Reset only the nodes that actually need it (e.g. nodeaddr still set), mirroring clustermgtd's inactive
         # cleanup.
@@ -252,6 +253,8 @@ def reset_nodes_in_inactive_partitions():
                 sorted(nodes_to_reset),
             )
             reset_nodes(nodes_to_reset, state="down", reason="inactive partition", raise_on_error=False)
+        else:
+            log.info("No nodes of INACTIVE partitions need to be reset.")
     except Exception as e:
         log.error("Failed when resetting nodes of INACTIVE partitions with error %s", e)
 
--- a/src/common/schedulers/slurm_commands.py
+++ b/src/common/schedulers/slurm_commands.py
@@ -221,6 +221,27 @@ def update_all_partitions(state, reset_node_addrs_hostname):
         return False
 
 
+def reset_nodes_in_inactive_partitions():
+    """
+    Reset nodeaddr/nodehostname of the nodes belonging to INACTIVE partitions and set them back to idle.
+
+    This is meant to be called when starting the compute fleet, before bringing partitions back UP, to clean up nodes
+    that an INACTIVE partition may have left behind (e.g. dynamic nodes that were powering up without a backing
+    instance when protected mode disabled the partition). If not reset, those nodes would be detected as bootstrap
+    failures right after start and send the cluster back into protected mode. Only INACTIVE partitions are considered,
+    so nodes of active partitions (which may be running jobs) are not affected.
+    """
+    try:
+        inactive_nodes = [
+            part.nodenames for part in get_partitions_info() if PartitionStatus(part.state) == PartitionStatus.INACTIVE
+        ]
+        if inactive_nodes:
+            log.info("Resetting nodes of INACTIVE partitions: %s", inactive_nodes)
+            set_nodes_idle(",".join(inactive_nodes), reset_node_addrs_hostname=True)
+    except Exception as e:
+        log.error("Failed when resetting nodes of INACTIVE partitions with error %s", e)
+
+
 def _batch_attribute(attribute, batch_size, expected_length=None):
     """Parse an attribute into batches."""
     if type(attribute) is str:
--- a/src/common/schedulers/slurm_commands.py
+++ b/src/common/schedulers/slurm_commands.py
@@ -223,7 +223,7 @@ def update_all_partitions(state, reset_node_addrs_hostname):
 
 def reset_nodes_in_inactive_partitions():
     """
-    Reset nodeaddr/nodehostname of the nodes belonging to INACTIVE partitions and set them back to idle.
+    Reset nodeaddr/nodehostname and set to down the nodes belonging to INACTIVE partitions.
 
     This is meant to be called when starting the compute fleet, before bringing partitions back UP, to clean up nodes
     that an INACTIVE partition may have left behind (e.g. dynamic nodes that were powering up without a backing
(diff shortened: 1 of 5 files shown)


src/slurm_plugin/fleet_status_manager.py
```diff
@@ -99,7 +99,10 @@ def _start_partitions():
 
 def _stop_partitions(config):
     log.info("Setting slurm partitions to INACTIVE and terminating all compute nodes...")
-    update_all_partitions(PartitionStatus.INACTIVE, reset_node_addrs_hostname=True)
+    # Reset nodes for all partitions, even those already INACTIVE: protected mode may have left dirty nodes.
+    update_all_partitions(
```
gmarciani: In our public doc we declare it is enough to start the fleet to recover from the protected mode.

See https://docs.aws.amazon.com/parallelcluster/latest/ug/slurm-protected-mode-v3.html#slurm-protected-mode-exit-v3

This is the correct expectation because we do not want to force the user to stop the fleet just to recover one queue from protected mode. 

However, it seems from the code change that the fleet stop is required to make the recovery reliable.
Is it possible to make the start fleet able to recover without the need for the stop?

hehe7318: That's a good question, we may want to move the logic from cluster.stop() to cluster.start().
For example during the cluster.start(), we reset all the nodes in **INACTIVE** partitions.

hehe7318: Done in new version!

tests/common/schedulers/test_slurm_commands.py
```diff
@@ -976,6 +977,40 @@ def test_update_all_partitions(
...
+        (
+            [
+                SlurmPartition("part-1", "node-1,node-2", "UP"),
+                SlurmPartition("part-2", "node-3,node-4", "UP"),
+            ],
+            None,
+        ),
+    ],
+)
+def test_reset_nodes_in_inactive_partitions(mock_partitions, expected_idle_call, mocker):
```
gmarciani: [Testing] It is important that the function `reset_nodes_in_inactive_partitions` is executed before `update_all_partitions(PartitionStatus.UP)`. Can we capture the ordering with unit test?

src/slurm_plugin/fleet_status_manager.py
```diff
@@ -91,6 +95,7 @@ def _manage_fleet_status_transition(config, computefleet_status_data_path):
 
 def _start_partitions():
     log.info("Setting slurm partitions to UP and resuming nodes...")
+    reset_nodes_in_inactive_partitions()
     update_all_partitions(PartitionStatus.UP, reset_node_addrs_hostname=False)
```
gmarciani, line 99: We explicitly set `reset_node_addrs_hostname=False`. which seems the opposite of what we are doing by adding `reset_nodes_in_inactive_partitions`. Why do we think this is not a contradiction and it is fine?

src/common/schedulers/slurm_commands.py
```diff
@@ -221,6 +221,27 @@ def update_all_partitions(state, reset_node_addrs_hostname):
...
+    failures right after start and send the cluster back into protected mode. Only INACTIVE partitions are considered,
+    so nodes of active partitions (which may be running jobs) are not affected.
+    """
+    try:
+        inactive_nodes = [
+            part.nodenames for part in get_partitions_info() if PartitionStatus(part.state) == PartitionStatus.INACTIVE
+        ]
+        if inactive_nodes:
+            log.info("Resetting nodes of INACTIVE partitions: %s", inactive_nodes)
+            set_nodes_idle(",".join(inactive_nodes), reset_node_addrs_hostname=True)
```
gmarciani: The function `set_nodes_idle` sets the state of nodes in INACTIVE partitions to `resume`. If I am not wrong this is the first time we set resume on inactive partitions. What is the state of those nodes before we resume those? what is the effect of setting them idle when the partition is inactive?

gmarciani: Would it be possible for inactive_nodes to contain empty elements?
If possible then we must put a safeguard because empty elements could produce a malformed comma separated list.

Example:

```
inactive_nodes = ["node1,node2", "", "node4,node5"]
",".join(inactive_nodes) -> "node1,node2,,node3,node4"  (two commas with no elements inside)

```

src/slurm_plugin/fleet_status_manager.py
hehe7318: `update_all_partitions(UP, reset_node_addrs_hostname=False)` only flips the partition state and intentionally leaves node addrs alone. And the node reset is a separate, explicit step (`reset_nodes_in_inactive_partitions`) scoped only to INACTIVE partitions and executed before the partitions are brought back UP.

So not a contradiction.

tests/common/schedulers/test_slurm_commands.py
hehe7318: Done!

src/common/schedulers/slurm_commands.py
hehe7318: Good catch, now I switched to `reset_nodes(state="down")`, same as clustermgtd's `_reset_nodes_in_inactive_partitions` does for inactive partitions. Also added `needs_reset_when_inactive()` filtering so that only the nodes that actually need it (e.g. nodeaddr still set) are reset, again to keep the behavior aligned with clustermgtd's inactive cleanup.

Because I saw https://github.com/aws/aws-parallelcluster-node/blob/d31f163cf480f6add683f4b99cff1d81ba9666e2/src/slurm_plugin/clustermgtd.py#L637

The comment there says power_down doesn't seem to be applied while the partition is inactive, so my previous change had a chance of not working as well.

hehe7318: Thanks! Fixed, partitions with empty `nodenames` are skipped, and the list now goes through `get_nodes_info` + per-node filtering. Added a unit test for the empty-partition case.

src/common/schedulers/slurm_commands.py
```diff
@@ -221,6 +221,41 @@ def update_all_partitions(state, reset_node_addrs_hostname):
...
+    """
+    try:
+        # Collect the node names of INACTIVE partitions, skipping partitions with no nodes to avoid building a
+        # malformed (e.g. empty or with consecutive commas) node list.
+        inactive_partition_nodes = [
+            part.nodenames
+            for part in get_partitions_info()
+            if PartitionStatus(part.state) == PartitionStatus.INACTIVE and part.nodenames
+        ]
+        if not inactive_partition_nodes:
```
gmarciani, line 242: Could you please add a log line saying that there is no nodes in inactive partitions to reset?
I know it may seem a minor, but considering that we are adding this logic to fix a race condition, it may be useuful for the troubleshooting to know exactly what the daemons sees when it checks

hehe7318: Done!

gmarciani (approved):

PR #727, opened 2026-07-24
[Node] Log that an unhealthy static node is not protected from replacement

hehe7318:
### Description of changes
A static node that clustermgtd is replacing is shielded from being replaced again while it bootstraps (the replacement protection window). When that protection did not apply, the logs never said so: they showed the node being terminated and relaunched with no indication that it had lost, or never had, the grace period. This is what made the maintenance reservation issue hard to diagnose, and the same gap affects every other reason a node can end up unprotected.

- When a DOWN or DRAINED static node is found unhealthy, the existing warning now also reports that the node is not within the replacement protection window and that it will be replaced. This is the point where the decision is made, and it holds for every reason the window may not apply.
- The log listing the unhealthy static nodes now includes the node states, so the state that led to the replacement is visible without correlating against `scontrol` output.

The node state carries the reason, so the logging does not assume a specific cause and stays accurate if the conditions change.

This also fixes the format string of the debug log covering the opposite case (node still protected), which was missing a placeholder and raised `TypeError: not all arguments converted during string formatting` whenever it was emitted.

#### Example

A node being replaced fails its EC2 health check while bootstrapping, so it is removed from the replacement set and replaced again immediately:

```
Detected failed health check for static node in replacement. No longer considering nodes ['queue1-st-cr1-1'] as in replacement process. Will attempt to replace node again immediately.
Node state check: node queue1-st-cr1-1(192.168.104.165) in DOWN and not within the replacement protection window, it is considered unhealthy and will be replaced, node state: DOWN+CLOUD   <-- reason and consequence added
Found the following unhealthy static nodes: (x1) ['queue1-st-cr1-1(192.168.104.165) in state DOWN+CLOUD']   <-- node states added
Setting unhealthy static nodes to DOWN
Terminating instances backing unhealthy static nodes
Terminating instances (x1) ['i-021cb4b7fa1050c13']
Launching new instances for unhealthy static nodes
```

A DRAINED node reads the same way:

```
Node state check: node queue1-st-cr1-1(192.168.104.165) in DRAINED and not within the replacement protection window, it is considered unhealthy and will be replaced, node state: IDLE+CLOUD+DRAIN
```

While a node is inside the protection window it is reported as healthy and only a debug line is emitted, so the warning above only shows up for nodes that are not shielded.

### Tests
* Unit tests passed, including new cases asserting the unhealthy-state warning reports the missing protection window for DOWN and DRAINED nodes, and that no warning is emitted while the node is still protected.
* Verified on a real cluster with this branch installed via `DevSettings/NodePackage`.

### References
* Follow-up to the COE action item for clearer termination logging.

### Checklist

### DIFF
--- a/src/slurm_plugin/clustermgtd.py
+++ b/src/slurm_plugin/clustermgtd.py
@@ -961,7 +961,11 @@ class ClusterManager:
             log.info("Found the following unhealthy dynamic nodes: %s", print_with_count(unhealthy_dynamic_nodes))
             self._handle_unhealthy_dynamic_nodes(unhealthy_dynamic_nodes)
         if unhealthy_static_nodes:
-            log.info("Found the following unhealthy static nodes: %s", print_with_count(unhealthy_static_nodes))
+            # Include the node states, which carry the reason the nodes are considered unhealthy.
+            log.info(
+                "Found the following unhealthy static nodes: %s",
+                print_with_count([f"{node} in state {node.state_string}" for node in unhealthy_static_nodes]),
+            )
             self._handle_unhealthy_static_nodes(unhealthy_static_nodes)
 
         # evaluate partitions to put in protected mode and ICEs nodes to terminate
--- a/src/slurm_plugin/slurm_resources.py
+++ b/src/slurm_plugin/slurm_resources.py
@@ -589,20 +589,30 @@ class StaticNode(SlurmNode):
                 return True
             else:
                 if log_warn_if_unhealthy:
-                    logger.warning("Node state check: node %s in DRAINED, node state: %s", self, self.state_string)
+                    logger.warning(
+                        "Node state check: node %s in DRAINED and not within the replacement protection window, "
+                        "it is considered unhealthy and will be replaced, node state: %s",
+                        self,
+                        self.state_string,
+                    )
                 return False
         # Check to see if node is in DOWN, ignoring any node currently being replaced
         elif self.is_down() and consider_down_as_unhealthy:
             if self.is_being_replaced:
                 logger.debug(
-                    "Node state check: node %s in DOWN but is currently being replaced, ignoring. Node state: ",
+                    "Node state check: node %s in DOWN but is currently being replaced, ignoring, node state: %s",
                     self,
                     self.state_string,
                 )
                 return True
             else:
                 if log_warn_if_unhealthy:
-                    logger.warning("Node state check: node %s in DOWN, node state: %s", self, self.state_string)
+                    logger.warning(
+                        "Node state check: node %s in DOWN and not within the replacement protection window, "
+                        "it is considered unhealthy and will be replaced, node state: %s",
+                        self,
+                        self.state_string,
+                    )
                 return False
         return True
 
--- a/tests/slurm_plugin/slurm_resources/test_slurm_resources.py
+++ b/tests/slurm_plugin/slurm_resources/test_slurm_resources.py
@@ -719,6 +719,40 @@ def test_slurm_node_is_state_healthy(
 
 
 @pytest.mark.parametrize(
+    "node, is_being_replaced, expected_warning",
+    [
+        pytest.param(
+            StaticNode("queue-st-c5xlarge-1", "some_ip", "hostname", "DOWN+CLOUD+MAINTENANCE+RESERVED", "queue"),
+            False,
+            "node queue-st-c5xlarge-1(some_ip) in DOWN and not within the replacement protection window",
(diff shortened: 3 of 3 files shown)


src/slurm_plugin/slurm_resources.py
```diff
@@ -589,20 +589,30 @@ def is_state_healthy(self, consider_drain_as_unhealthy, consider_down_as_unhealt
                 return True
             else:
                 if log_warn_if_unhealthy:
-                    logger.warning("Node state check: node %s in DRAINED, node state: %s", self, self.state_string)
+                    logger.warning(
+                        "Node state check: node %s in DRAINED and not within the replacement protection window, "
```
gmarciani, line 593: 1. how can we say here that the node is not in replacement protection window?
2. the log becomes useful if it states what is the consequence of not being in the protection window

hehe7318: >how can we say here that the node is not in replacement protection window?

The `else` is reached only when `is_being_replaced` is false, and that attribute is recomputed for every node immediately before this check runs, in the same `_maintain_nodes()` iteration. See

https://github.com/aws/aws-parallelcluster-node/blob/f0b55b9865e421eb90d5080724c1305a1e4fcc0a/src/slurm_plugin/clustermgtd.py#L954-L965

and

https://github.com/aws/aws-parallelcluster-node/blob/f0b55b9865e421eb90d5080724c1305a1e4fcc0a/src/slurm_plugin/clustermgtd.py#L756-L759

`_is_node_being_replaced()` is exactly the protection-window check:

https://github.com/aws/aws-parallelcluster-node/blob/f0b55b9865e421eb90d5080724c1305a1e4fcc0a/src/slurm_plugin/clustermgtd.py#L1217-L1243

One thing worth notice: `is_being_replaced` being false covers two different situations. Either the node was in the protection window and left it (its health check failed or the timeout reached), or it was never in the window at all -> only nodes that clustermgtd itself launched as replacements are added to `_static_nodes_in_replacement`, so a node that bootstrapped normally and was later set DOWN never had a window to begin with. And it will have same logging.

>the log becomes useful if it states what is the consequence of not being in the protection window

What about:

```
Node state check: node queue1-st-cr1-1(192.168.104.165) in DOWN and not within the replacement protection window, it is considered unhealthy and will be replaced, node state: ...
```

Is it okay?

hehe7318: Applied, thanks. Both warnings now state the consequence:

```
Node state check: node queue1-st-cr1-1(192.168.104.165) in DOWN and not within the replacement protection window, it is considered unhealthy and will be replaced, node state: DOWN+CLOUD
Node state check: node queue1-st-cr1-1(192.168.104.165) in DRAINED and not within the replacement protection window, it is considered unhealthy and will be replaced, node state: IDLE+CLOUD+DRAIN
```

src/slurm_plugin/clustermgtd.py
```diff
@@ -903,7 +903,13 @@ def _handle_unhealthy_static_nodes(self, unhealthy_static_nodes):
         instances_to_terminate = [node.instance.id for node in unhealthy_static_nodes if node.instance]
 
         if instances_to_terminate:
-            log.info("Terminating instances backing unhealthy static nodes")
+            # Include the node states, which carry the reason the nodes are being replaced.
+            log.info(
+                "Terminating instances backing unhealthy static nodes: %s",
+                print_with_count(
```
gmarciani: This array can potentially contain thousands of entries. Are we sure the logger would be able to handle?

hehe7318: I checked whether the logger actually has a problem with the full list.

It does not: clustermgtd logs through a plain `StreamHandler` to stdout, and supervisord writes that to `/var/log/parallelcluster/clustermgtd` with `stdout_logfile_maxbytes = 0`, i.e. no size cap and no rotation on that stream. There is no line-length limit anywhere in that path.

More to the point, the surrounding logs already print full lists of the same size — `Found the following unhealthy static nodes` listed every node before this PR, and `delete_instances` logs every instance id. So this does not introduce a new order of magnitude.

I also moved the states onto the existing `Found the following unhealthy static nodes` line instead of adding a second duplicated line about the same set of nodes, so the node names and their states are reported together and `Terminating instances backing unhealthy static nodes` is left untouched:

```
Found the following unhealthy static nodes: (x1) ['queue1-st-cr1-1(192.168.104.165) in state DOWN+CLOUD']
```

gmarciani (approved):

PR #591, opened 2023-11-03
[develop] Add support for Capacity Blocks for ML

enrico-usai:
### Capacity Block for ML

Capacity Block reservation for Machine Learning (CB) allow AWS users to reserve blocks of GPU capacity for a future start date and for a time duration of their choice. They are capacity reservations of type `capacity-block`, with additional attributes like: `StartDate`, `EndDate` and additional states.

Capacity Block instances are managed by ParallelCluster as a static nodes but in a peculiar way. ParallelCluster permits to create a cluster even if the CB reservation is not yet active and automatically launch the instances when it will become active.

The Slurm nodes corresponding to compute resources associated to CB reservations are kept in maintenance until the CB start time is reached. When CB is not active, the slurm nodes will stay in a Slurm reservation/maintenance state, associated to the slurm admin user, it means that the nodes can accept jobs but they will stay in pending, until the reservation will be removed.

Clustermgtd will automatically create/delete slurm reservations, putting the related CB nodes in maintenance or not, accordingly to the CB state. The Slurm reservation will be removed when CB is active, nodes will start and become available for the pending jobs or for new jobs submissions.

When the CB end time is reached nodes will be re-inserted in a reservation/maintenance state. It’s up to the user to resubmit/requeue the jobs to a new queue/compute-resource when CB time is ended.

### CapacityBlockManager

`CapacityBlockManager` is a new class that will perform the following actions:
* when CB is not active or expired, a new Slurm reservation will be created/updated
* when CB is active, Slurm reservation will be removed and nodes will become standard static instances

Slurm reservation name will be `pcluster-{capacity_block_reservation_id}`

CapacityBlockManager will be initialized by clustermgtd with region, boto3_config and fleet_config info.
Fleet config is the existing `fleet-config.json` file, that has been extended to include capacity reservation ids and capacity type (ondemand vs spot vs capacity-block).
The manager will reload fleet-config info and capacity reservation info every time the daemon is restarted
or when the config is modified.
Current logic updates CB reservation info every 10 minutes.

The manager will remove the nodes associated with CB reservations that are not active
from the list of unhealthy static nodenames.

### CapacityBlock

`CapacityBlock` is a new class to store internal info about the capacity block,
merging them from ec2 (e.g. capacity block state) and from the config (i.e. list of nodes associated to it).

### Managing slurm reservations
Created a new set of commands (using scontrol) to manage the Slurm reservations:
* create a reservation
[...]

PR #728, opened 2026-07-27
Stop removing nodes in maintenance from in replacement set

hgreebe:
### Description of changes
* Stop removing nodes in maintenance from in replacement set
* Resolve bug where static nodes instance replacement fails when node is in a user created maintenance reservation
* Does not remove nodes in maintenance from `nodes_still_in_replacement`
* Nodes in maintenance will still wait the replacement timeout window before they are deemed unhealthy

### Tests
* Ran:
```
test-suites:
  schedulers:
    test_slurm.py::test_slurm_maintenance_reservation:
      dimensions:
        - regions: ["us-east-1"]
          instances: {{ common.INSTANCES_DEFAULT_X86 }}
          oss: ["alinux2023"]
          schedulers: ["slurm"]
```

### References
*Bug originally introduced in this PR: https://github.com/aws/aws-parallelcluster-node/pull/591

### DIFF
--- a/src/slurm_plugin/clustermgtd.py
+++ b/src/slurm_plugin/clustermgtd.py
@@ -741,13 +741,13 @@ class ClusterManager:
         return is_instance_status_unhealthy
 
     def _update_static_nodes_in_replacement(self, slurm_nodes: List[SlurmNode]):
-        """Remove from self.static_nodes_in_replacement nodes that are up or that are in maintenance."""
+        """Remove from self.static_nodes_in_replacement nodes that are up."""
         nodename_to_slurm_nodes_map = {node.name: node for node in slurm_nodes}
         nodes_still_in_replacement = set()
         for nodename in self._static_nodes_in_replacement:
             node = nodename_to_slurm_nodes_map.get(nodename)
-            # Consider nodename still in replacement if node is not up and not in maintenance
-            if node and not node.is_up() and not node.is_in_maintenance():
+            # Consider nodename still in replacement if node is not up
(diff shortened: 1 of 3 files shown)


src/slurm_plugin/clustermgtd.py
```diff
@@ -741,13 +741,13 @@ def _is_instance_unhealthy(self, instance_status: EC2InstanceHealthState, health
...
-        """Remove from self.static_nodes_in_replacement nodes that are up or that are in maintenance."""
+        """Remove from self.static_nodes_in_replacement nodes that are up."""
         nodename_to_slurm_nodes_map = {node.name: node for node in slurm_nodes}
         nodes_still_in_replacement = set()
         for nodename in self._static_nodes_in_replacement:
             node = nodename_to_slurm_nodes_map.get(nodename)
-            # Consider nodename still in replacement if node is not up and not in maintenance
-            if node and not node.is_up() and not node.is_in_maintenance():
+            # Consider nodename still in replacement if node is not up
+            if node and not node.is_up():
```
gmarciani, line 750: With this change a static node with a pcluster-managed reservation is now added to the list of nodes in replacement. What are the effects on a static node with CB reservation that is not active?

hgreebe: There are no effects. Static nodes with the CB reservation are never added to the replacement list so this change does not affect them. They are never replaced because they are skipped in`_find_unhealthy_slurm_nodes`

gmarciani (approved):

PR #734, opened 2026-08-20
Fix EC2 throttling during computenode launch is not retried and is reported as ICE when using multi InstanceTypes or SubnetIds

hehe7318:
#### Customer impact

Since [#578](https://github.com/aws/aws-parallelcluster-node/pull/578), EC2 throttling during a compute node scale-up is never retried, and every launch failure is instead reported as insufficient capacity.

**Preconditions**: a compute resource with more than one instance type **or** more than one subnet, and a scale-up large enough to exhaust the `RunInstances` resource token bucket, which holds 1000 instances and refills at 2/s. See the number in ref link.

**On `best-effort` (the default)**: a single throttled call disables the **whole compute resource**, not only the batch that failed. Measured with `MaxCount 3000` and `sbatch -N 1500`: 500 nodes went DOWN and the remaining 1500 got `Temporarily disabling node due to insufficient capacity` — 2000 of 3000 nodes unusable for 600s (`insufficient_capacity_timeout`), with the job stuck `PENDING`.

**On `all-or-nothing`**: the compute resource cannot scale up reliably. Every round releases the instances already launched and starts over, so a large scale-up would fail.

#### Root cause

`CreateFleet` returns one entry in `Errors` per launch template override, so a compute resource with more than one instance type or more than one subnet gets an entry for **every** override that was not fulfilled, not only for the one that actually failed. Those extra entries all carry the same pair:

```
("UnfulfillableCapacity", "Failed to fulfill capacity. Please review errors in the response.")
```

which points at the real error rather than being one.

With 36 overrides a single failure therefore arrives as 36 entries, so the pre-existing `len(err_list) == 1` condition in `Ec2CreateFleetManager._launch_instances` never holds. No `LaunchInstancesError` is raised, and `InstanceManager._launch_ec2_instances` falls back to a hardcoded `InsufficientInstanceCapacity`. Three things follow:

* **Throttling is never retried.** `launch_ec2_instances` already implements the exponential backoff [the EC2 documentation recommends](https://docs.aws.amazon.com/ec2/latest/devguide/ec2-api-throttling.html#api-backoff), sized to wait for the bucket to refill a full `launch_max_batch_size` batch, but no exception ever reaches it. The batch is abandoned after a single attempt even though retrying is all it needed.
* **The compute resource is failed over for nothing.** `InsufficientInstanceCapacity` is in `SlurmNode.EC2_ICE_ERROR_CODES`, so `clustermgtd` treats the failure as a capacity shortage and disables the whole compute resource, not just the batch that failed.
* **The real cause is lost.** A `VcpuLimitExceeded`, an `Unsupported` instance type or an `UnauthorizedOperation` is all flattened into insufficient capacity.

#### The fix

1. **Drop the entries that only point at the real error** before deciding what to report, so the existing `len(err_list) == 1` condition works again for multi-override compute resources. The response is left untouched when those entries are all it carries, so the worst case is today's behaviour.
2. **Prefer throttling over any other remaining cause**, as a safety net for a response carrying more than one real error. Throttling is the only cause that resolves on its own, and reporting it as insufficient capacity costs a fail-over of the whole compute resource.

### Tests

* End-to-end test on real clusters of both best-effort and all-or-nothing chain, both worked as expected.

### References

* https://docs.aws.amazon.com/ec2/latest/devguide/ec2-api-throttling.html
* https://docs.aws.amazon.com/ec2/latest/devguide/ec2-api-throttling.html#throttling-limits-cost-based

### Checklist

### DIFF
--- a/src/slurm_plugin/fleet_manager.py
+++ b/src/slurm_plugin/fleet_manager.py
@@ -35,8 +35,7 @@ INSTANCE_INFO_RETRIEVAL_MAX_BACKOFF = 30
 LAUNCH_THROTTLING_ERROR_CODE = "RequestLimitExceeded"
 
 # An override that CreateFleet did not fulfill because another override failed is reported with this exact
-# code and message pair, which points at the real error rather than being one. Other UnfulfillableCapacity
-# messages, such as the one about MinTargetCapacity constraints, do describe a real cause and are kept.
+# code and message pair, which points at the real error rather than being one.
 UNFULFILLED_OVERRIDE_ERROR = (
     "UnfulfillableCapacity",
     "Failed to fulfill capacity. Please review errors in the response.",
@@ -443,7 +442,7 @@ class Ec2CreateFleetManager(FleetManager):
                 logger.error("Unable to retrieve instance info for instances: %s", partial_instance_ids)
 
             if not instances:
-                # Every error is logged above, but only the ones describing a cause are worth reporting.
+                # Drop the entries that only point at the real error, unless the response carries nothing else.
                 real_errors = [
                     err
                     for err in err_list
@@ -451,8 +450,9 @@ class Ec2CreateFleetManager(FleetManager):
                 ]
                 if real_errors:
                     err_list = real_errors
-                # Throttling is reported even alongside other causes, so that the launch is retried with backoff
-                # instead of the nodes being recorded as insufficient capacity, which fails the compute resource over.
+                # A single cause is normally left. Should there be several, prefer throttling as a safety net: it is
+                # the only cause that resolves on its own, and reporting it as insufficient capacity would instead
+                # fail the compute resource over.
                 throttling = next(
                     (err for err in err_list if err.get("ErrorCode") == LAUNCH_THROTTLING_ERROR_CODE), None
                 )
--- a/src/slurm_plugin/fleet_manager.py
+++ b/src/slurm_plugin/fleet_manager.py
@@ -30,6 +30,18 @@ logger = logging.getLogger(__name__)
 INSTANCE_INFO_RETRIEVAL_TIMEOUT_DEFAULT = 90
 INSTANCE_INFO_RETRIEVAL_MAX_BACKOFF = 30
 
+# The only error code EC2 returns when a launch request is throttled, see
+# https://docs.aws.amazon.com/ec2/latest/devguide/ec2-api-throttling.html
+LAUNCH_THROTTLING_ERROR_CODE = "RequestLimitExceeded"
+
+# An override that CreateFleet did not fulfill because another override failed is reported with this exact
+# code and message pair, which points at the real error rather than being one. Other UnfulfillableCapacity
+# messages, such as the one about MinTargetCapacity constraints, do describe a real cause and are kept.
+UNFULFILLED_OVERRIDE_ERROR = (
+    "UnfulfillableCapacity",
+    "Failed to fulfill capacity. Please review errors in the response.",
+)
+
 
 class EC2Instance:
     def __init__(self, id, private_ip, hostname, all_private_ips, launch_time):
@@ -430,6 +442,23 @@ class Ec2CreateFleetManager(FleetManager):
             if partial_instance_ids:
                 logger.error("Unable to retrieve instance info for instances: %s", partial_instance_ids)
 
+            if not instances:
+                # Every error is logged above, but only the ones describing a cause are worth reporting.
+                real_errors = [
+                    err
+                    for err in err_list
+                    if (err.get("ErrorCode"), err.get("ErrorMessage")) != UNFULFILLED_OVERRIDE_ERROR
+                ]
+                if real_errors:
+                    err_list = real_errors
+                # Throttling is reported even alongside other causes, so that the launch is retried with backoff
+                # instead of the nodes being recorded as insufficient capacity, which fails the compute resource over.
+                throttling = next(
+                    (err for err in err_list if err.get("ErrorCode") == LAUNCH_THROTTLING_ERROR_CODE), None
+                )
+                if throttling:
+                    raise LaunchInstancesError(throttling.get("ErrorCode"), throttling.get("ErrorMessage"))
+            # Any other cause is reported only when the response is unambiguous, unchanged from before.
             if not instances and len(err_list) == 1:
                 raise LaunchInstancesError(err_list[0].get("ErrorCode"), err_list[0].get("ErrorMessage"))
             return {"Instances": instances}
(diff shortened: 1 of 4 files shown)


src/slurm_plugin/fleet_manager.py
```diff
@@ -30,6 +30,17 @@
 INSTANCE_INFO_RETRIEVAL_TIMEOUT_DEFAULT = 90
 INSTANCE_INFO_RETRIEVAL_MAX_BACKOFF = 30
 
+# The only error code EC2 returns when a launch request is throttled, see
+# https://docs.aws.amazon.com/ec2/latest/devguide/ec2-api-throttling.html
+LAUNCH_THROTTLING_ERROR_CODE = "RequestLimitExceeded"
+
+# An override that CreateFleet did not fulfill because another override failed is reported with this exact
+# code and message pair, which points at the real error rather than being one.
+UNFULFILLED_OVERRIDE_ERROR = (
```
hanwen-cluster, line 39: What does "override" mean here

src/slurm_plugin/fleet_manager.py
```diff
@@ -430,6 +441,24 @@ def _launch_instances(self, launch_params):
             if partial_instance_ids:
                 logger.error("Unable to retrieve instance info for instances: %s", partial_instance_ids)
 
+            if not instances:
+                # Drop the entries that only point at the real error, unless the response carries nothing else.
+                real_errors = [
+                    err
+                    for err in err_list
+                    if (err.get("ErrorCode"), err.get("ErrorMessage")) != UNFULFILLED_OVERRIDE_ERROR
```
hanwen-cluster, line 449: UnfulfillableCapacity is real error to right?
According to https://docs.aws.amazon.com/ec2/latest/devguide/errors-overview.html:
> At this time there isn't enough spare capacity to fulfill your request for Spot Instances. You can wait a few minutes to see whether capacity becomes available for your request. Alternatively, create a more flexible request. For example, include additional instance types, include additional Availability Zones, or use the capacity-optimized allocation strategy.

Is it safe to drop it?

src/slurm_plugin/fleet_manager.py
hehe7318: It means the launch template "override".

src/slurm_plugin/fleet_manager.py
hehe7318: Yes it's safe. The filter matches the code and the error message as a pair, not the code
alone.

The code+message pair we dropped is:
```
("UnfulfillableCapacity", "Failed to fulfill capacity. Please review errors in the response.")
```
Every other `UnfulfillableCapacity` message passes through untouched, including
"Unable to fulfill request due to MinTargetCapacity constraints. Please adjust your request and try again.",
which is what all-or-nothing produces and is a genuine cause.

And the message itself is a evidence. "Please review errors in the response" means something if the response carries another error to review. As the sole entry it would be self-referential.

hanwen-cluster (approved):

src/slurm_plugin/fleet_manager.py
```diff
@@ -430,7 +441,27 @@ def _launch_instances(self, launch_params):
...
+                ]
+                if real_errors:
+                    err_list = real_errors
+                # A single cause is normally left. Should there be several, prefer throttling as a safety net: it is
+                # the only cause that resolves on its own, and reporting it as insufficient capacity would instead
+                # fail the compute resource over.
+                throttling = next(
```
gmarciani, line 456: I agree with the safety net approach, but not on limiting it to throttling.
With your change, we surface throttling if we have a mix of root causes because we want to favor retryable errors and I agree. However, throttling is not the only retriable error, .e.g internal errors are retriable as well.

src/slurm_plugin/fleet_manager.py
```diff
@@ -430,7 +441,27 @@ def _launch_instances(self, launch_params):
...
+                    (err for err in err_list if err.get("ErrorCode") == LAUNCH_THROTTLING_ERROR_CODE), None
+                )
+                if throttling:
+                    raise LaunchInstancesError(throttling.get("ErrorCode"), throttling.get("ErrorMessage"))
+            # Normally a single cause is left. Reporting the first one of several is a second safety net: the
```
gmarciani, line 461: I imagine that you are keeping this approach of bubbling up only the first error to keep the fix minimal and I agree on the overall minimization approach. However:
1. Why do we still want to bubble up only the first error code?
1. The alternative is to let `LaunchInstancesError` contain all the types of errors so that the upstream logic can take a decision on a complete observation. Why not doing that?

Example: if the request fails due to ICE + another unretriable error E2, but E2 is listed first, we will bubble up only E2. What would be the consequence?

src/slurm_plugin/fleet_manager.py
hehe7318: That's a real behavior change and we should not do it without clear decision doc. Other error code will be handled in other code path. It's safe to only handle and retry the throttling error here - the exponential backoff is designed only for it.

Also I raise a doubt here:
For example `VCpuLimitExceed` error, I don't think we should retry, but we should fail, requeue the job, print the error in log.

src/slurm_plugin/fleet_manager.py
hehe7318: Just to clarify - if everything goes correct and well, after the fix, after removing the duplicated UnfulfillableCapacity, the `err_list` should only contain 1 error. I added a safety net here in case there's edge case.

>Why do we still want to bubble up only the first error code?

To introduce the minimal changes, because node package is sensitive and I want to avoid any regression possibility.

>The alternative is to let LaunchInstancesError contain all the types of errors so that the upstream logic can take a decision on a complete observation. Why not doing that

Good question, I tried once, but other code chain also use this returned error. To do it we need to restructure and it introduce too much risky code changes. To solve an edge case, I don't think it deserve.

>if the request fails due to ICE + another unretriable error E2, but E2 is listed first, we will bubble up only E2. What would be the consequence?

Our code logic will treat it as an ICE issue and trigger fast failover mechanism. Which aligns with today's(3.16.0) behavior. If it's the opposite, another unretriable error E1 + ICE, we fail and requeue job without retry, which fix today's wrong behavior to the correct behavior.

src/slurm_plugin/fleet_manager.py
hgreebe: I agree that we should not limit it to throttling. If there are other errors due to real issues, it does not make sense to retry when it will fail anyways.

Also, it there is more than one error code, then the current behavior is to also not retry. I think we should keep this existing behavior.

src/slurm_plugin/fleet_manager.py
hgreebe: If we are using a fallback to always retry if there is throttling, why bubble up into one error code. It seems like the behavior would be the same whether we do that or not.

PR #740, opened 2026-09-11
Retry a throttled CreateFleet only when its other errors are pool-scoped

hehe7318:
### Description of changes
When a CreateFleet response that launched nothing carries RequestLimitExceeded next to other real errors, retrying the throttling only helps if the pools EC2 did not get to try can still serve the batch. That holds when every other error is confined to one instance type and subnet pool: a capacity shortage, an unsupported type in one AZ, a full or missing subnet, an AZ that lacks the requested volume type or is temporarily unavailable. Any other error, such as an account quota or a permission problem, would fail the retry on every pool alike, so it is now reported immediately instead of waiting for the rate limit to refill while the instances already launched for the job sit idle.

The pool-scoped set reuses EC2_ICE_ERROR_CODES and adds six subnet and AZ level codes. Codes outside the set, including unknown ones, are not retried: the batch fails with that code, as it did before the throttling fix minus the invented insufficient capacity. Responses without RequestLimitExceeded are unchanged.

### Reference
This is a follow-up PR of https://github.com/aws/aws-parallelcluster-node/pull/734: Fix EC2 throttling during computenode launch is not retried and is reported as ICE when using multi InstanceTypes or SubnetIds

### Tests
* E2E test run on a 3.16.1 cluster with 2 instance types × 2 subnets, feeding synthetic CreateFleet responses through the `slurm_plugin.overrides` hook: throttling alone was retried and launched as before, while throttling next to an injected `VcpuLimitExceeded` was not retried — the nodes went DOWN with `(Code:VcpuLimitExceeded)`, the job was requeued without an insufficient-capacity fail-over and completed on its next resume.
* The `VcpuLimitExceeded` entry was injected rather than produced by EC2; that a single response can carry several real errors, with the rate limit being the last pool tried, was confirmed with the EC2 Fleet team.

### DIFF
--- a/src/slurm_plugin/fleet_manager.py
+++ b/src/slurm_plugin/fleet_manager.py
@@ -21,6 +21,7 @@ from common.ec2_utils import get_private_ip_address_and_dns_name
 from common.utils import setup_logging_filter
 from retrying import retry
 from slurm_plugin.common import print_with_count
+from slurm_plugin.slurm_resources import SlurmNode
 
 logger = logging.getLogger(__name__)
 
@@ -39,6 +40,21 @@ LAUNCH_THROTTLING_ERROR_CODE = "RequestLimitExceeded"
 UNFULFILLED_OVERRIDE_ERROR = (
     "UnfulfillableCapacity",
     "Failed to fulfill capacity. Please review errors in the response.",
+)
+
+# Errors confined to one instance type and subnet pool. Reported alongside throttling, they leave the throttled
+# pools able to serve the batch once the rate limit refills, so the throttling is still retried. Any other error
+# alongside throttling would fail the retry on every pool alike and is reported instead.
+POOL_LEVEL_ERROR_CODES = frozenset(
+    SlurmNode.EC2_ICE_ERROR_CODES
+    | {
+        "InsufficientFreeAddressesInSubnet",
+        "InvalidSubnetID.NotFound",
+        "InvalidSubnet",
+        "InsufficientVolumeCapacity",
+        "VolumeTypeNotAvailableInZone",
+        "ServiceUnavailable",
+    }
 )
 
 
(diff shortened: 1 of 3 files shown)


src/slurm_plugin/fleet_manager.py
```diff
@@ -41,6 +42,21 @@
...
 
+# Errors confined to one instance type and subnet pool. Reported alongside throttling, they leave the throttled
+# pools able to serve the batch once the rate limit refills, so the throttling is still retried. Any other error
+# alongside throttling would fail the retry on every pool alike and is reported instead.
+POOL_LEVEL_ERROR_CODES = frozenset(
+    SlurmNode.EC2_ICE_ERROR_CODES
+    | {
+        "InsufficientFreeAddressesInSubnet",
+        "InvalidSubnetID.NotFound",
+        "InvalidSubnet",
```
gmarciani, line 53: I've never seen this error code. Always seen something like `InvalidSubnetSOMETHING`.
Where did we get this error code?

gmarciani:
At first I thought this fix deserves an entry in the changelog. However it seems to me it is included in the line below:
as it is a fix of https://github.com/aws/aws-parallelcluster-node/pull/734
```
Fix an issue where EC2 throttling during compute node launch is not retried and is reported as insufficient capacity when using Multiple Instance Types or multiple subnets.
```

Just double check this is intentional

src/slurm_plugin/fleet_manager.py
```diff
@@ -450,14 +466,24 @@ def _launch_instances(self, launch_params):
                 ]
                 if real_errors:
                     err_list = real_errors
-                # A single cause is normally left. Should there be several, prefer throttling as a safety net: it is
-                # the only cause that resolves on its own, and reporting it as insufficient capacity would instead
-                # fail the compute resource over.
+                # A single cause is normally left. Should there be several, throttling is retried only when every
+                # other cause is confined to a pool, since the throttled pools can then still serve the batch once the
+                # rate limit refills. Any other cause would fail the retry on every pool alike and is reported instead.
```
gmarciani, line 471: Can you specify an example in the comment of an error that when mixed with the throttling will not trigger a retry?

gmarciani:
The current change is basically fixing a missed part in https://github.com/aws/aws-parallelcluster-node/pull/734.
If this is the case, please link that PR in the description so that we can have the full picture in future

hehe7318:
> At first I thought this fix deserves an entry in the changelog. However it seems to me it is included in the line below: as it is a fix of #734
> 
> ```
> Fix an issue where EC2 throttling during compute node launch is not retried and is reported as insufficient capacity when using Multiple Instance Types or multiple subnets.
> ```
> 
> Just double check this is intentional

Yes, it's intentional, the new changes don't change the changelog wording.

src/slurm_plugin/fleet_manager.py
hehe7318: I got it from EC2 public doc: 
https://docs.aws.amazon.com/ec2/latest/devguide/errors-overview.html
Search `InvalidSubnet` in it. And you get:
>InvalidSubnet:
The specified subnet ID is not valid or does not exist.

hehe7318:
Linked: the Reference section of the description now points to #734.

src/slurm_plugin/fleet_manager.py
hehe7318: Added: 

```
# A single cause is normally left. Should there be several, throttling is retried only when every
# other cause is confined to a pool, since the throttled pools can then still serve the batch once the
# rate limit refills. Any other cause would fail the retry on every pool alike and is reported instead.
# For example:
# - [RequestLimitExceeded, InsufficientInstanceCapacity] -> retry, other pools can still serve the batch
# - [RequestLimitExceeded, VcpuLimitExceeded] -> report VcpuLimitExceeded, a retry hits the same limit
```

hanwen-cluster (approved):

